PR #12, opened 2021-11-11
Foreign Key Usage Design Spec

ghislaineguerin:
This PR contains the specs needed to implement the design for [Usage of Foreign Key Contraints](https://github.com/centerofci/mathesar/issues/243)

### DIFF
--- a/design/specs/foreign-key-constraints.md
+++ b/design/specs/foreign-key-constraints.md
@@ -168,6 +168,14 @@ The table selector allows users to select tables from a schema. This spec doesn'
 
 The record preview allows users to identify records linked in other tables. It shows the first five field names and values. A specific limit for the value length needs to be defined so that space usage is optimal.
 
+## Other Considerations
+
+### Search Columns
+
+- Some primary key columns would be useful, some very not useful (e.g., the one we create). Text helps with this, but there may also be other considerations.
+- Multi-column unique constraints are potentially useful, but only if all columns are included.
+- Non-unique columns are never useful unless a unique column is included (that gives info for the user to choose between rows with duplicate entries in non-unique columns).
+
 ## Related Discussions
 
 - [Editing values for columns with a foreign key constraint applied](https://github.com/centerofci/mathesar/discussions/796)
--- a/design/specs/foreign-key-constraints.md
+++ b/design/specs/foreign-key-constraints.md
@@ -163,8 +163,7 @@ Linked records can be changed or removed according to the user's preference. Sin
   - When clicked, a list of records is displayed inside the [record selector](#record-selector) component. The list shows the first 50 records.
   - An text input field is available within the dropdown to filter the records based on partial or complete value matches. As the user types in a value, the list narrows down the records. The system can search only the first five fields. Search across all types is supported.
   - If there's a single match for the entered value, it should become highlighted, allowing the user to confirm the selection.
-  - If the table contains no records, the user sees an option to add a record.
-  - If there aren't any matches, the user sees an option to add a record for the newly entered value or try changing the 'Search Columns' table preference.
+  - If there aren't any matches, a suggestion to change the 'Search Columns' table preference is displayed.
 
 ## Scenario 6: User changes the table link preferences
 
@@ -196,11 +195,11 @@ The record preview allows users to identify records linked in other tables. It s
 
 ### Linked Records for Multi-Column Foreign Key Constraints
 
-- The current UI does not consider multi-column foreign key constraints and how those would be retrieved and selected.
+- The current UI does not consider multi-column foreign key constraints and how those would be retrieved and selected. Record search and selection through the dropdown component will not be supported, instead the cells will be limited to input and edit of text values. The text values will have to match the primary keys of the referenced records and be unique.
 
 ### Usage of Color
 
-- A suggestion to use color as a means to differentiate table references has been included in this spec. However, there are implementation details that need to be discussed before the team can make a decision.
+- A suggestion to use color as a means to differentiate table references has been included in this spec. However, there are implementation details that need to be discussed before deciding if color will be used to create visual connection between elements.
 
 ## Related Discussions
 
--- a/design/specs/foreign-key-constraints.md
+++ b/design/specs/foreign-key-constraints.md
@@ -9,7 +9,21 @@ dateCreated: 2021-10-19T09:20:55.088Z
 ---
 ## Context
 
+This spec describes the design solution for [Usage of Foreign Key constraints](https://github.com/centerofci/mathesar/issues/243).
+
 ## Prototype Link and Video Walkthrough
+
+### Prototype
+
+[Prototype Link](https://mathesar-prototype.netlify.app/)
+
+### Videos
+
+- Scenario 1: <https://www.loom.com/share/def72e9ff7754b9e833c18dbdf9958c7>
+- Scenario 2: <https://www.loom.com/share/6aeeb12f776f4307beca97057ff60c48>
+- Scenario 3: <https://www.loom.com/share/d2cf81ac75ce47e8941669720ecc28e9>
+- Scenario 4: <https://www.loom.com/share/be12a128b41d443281babab0ccbcd112>
+- Scenario 5: <https://www.loom.com/share/9b5ab511981e4850acd7453fc907c81b>
 
 ## Scenarios
 
@@ -17,7 +31,7 @@ dateCreated: 2021-10-19T09:20:55.088Z
 
 A user wants to link records from another table into their current table. Both tables belong to the same schema. For this purpose, foreign key constraints can be applied to change the relationship between one table and another based on the values of columns of the linked tables.
 
-### Scenario 1a: Automatically from the 'Link Table' dialog
+### Scenario 1a: Automatically adding foreign key constraints from the 'Link Table' dialog
 
 #### Steps for 1a
 
--- a/design/specs/foreign-key-constraints.md
+++ b/design/specs/foreign-key-constraints.md
@@ -63,9 +63,11 @@ A user wants to link records from another table into their current table. Both t
 - The user can open a specific foreign key constraint form by clicking on the foreign key link in the column header.
 - If the modified FK constraint causes existing column values to become invalid, the user will receive a warning and be aware of the potential implications of this change, like loss of data or broken views.
 
-## User Identifies a Column With a Foreign Key Constraint Applied
+## Scenario 3 : User Identifies a Column With a Foreign Key Constraint Applied
 
-### The foreign key constraint is set to a single column
+### Scenario 3a: The foreign key constraint is set to a single column
+
+#### Steps for Scenario 3a
 
 - The user opens a table that contains columns with a foreign key constraint applied.
 - The user identifies the columns in two ways:
@@ -74,21 +76,31 @@ A user wants to link records from another table into their current table. Both t
     - The indicator can be hovered over to display a tooltip with information about the foreign key constraints.
   - The user looks at the column fields and sees that values are added with different styling, a tag-like element with a colored background.
 
-### The foreign key constraint is set to multiple columns
+### Scenario 3b: The foreign key constraint is set to multiple columns
+
+#### Steps for Scenario 3b
 
 - The user opens a table that contains columns with a multi-column foreign key constraint applied.
   - The indicator, in this case, looks similar to the single-column one. However, the user will understand that the foreign key constraints are referencing the same table.
     - The user can tell that a referenced table is the same because they share the same color. The system could use other UI elements instead of color. Still, the idea is to create a visual distinction that can be easily scanned.
   - The indicator can be hovered over to see additional information. From here, the user will understand that the constraint references more than one column.
 
-## User Sees a Preview of the Linked Record in a Column With a Foreign Key Constraint Applied
+## Scenario 4: User Sees a Preview of the Linked Record in a Column With a Foreign Key Constraint Applied
+
+### Scenario 4a: The option for record preview is enabled
+
+#### Steps for Scenario 4a
 
 - The user opens a table that contains columns with a foreign key constraint applied.
   - The column contains the referenced values for each cell. The first five fields are displayed.
     - The displayed fields are presented as concatenated values. Visual contrast is created between the field name and its corresponding value to be easier to read.
     - When clicked, the column displays a dropdown containing a pre-filtered list. The visible list items correspond to the linked record.
 
-## User Edits the Values of a Column With a Foreign Key Constraint Applied
+### Scenario 4b: The option for record preview is disabled
+
+#### Steps for Scenario 4b
+
+## Scenario 5: User Edits the Values of a Column With a Foreign Key Constraint Applied
 
 - The user opens a table that contains columns with a foreign key constraint applied.
   - The column contains the referenced values for each cell.
--- a/design/specs/foreign-key-constraints.md
+++ b/design/specs/foreign-key-constraints.md
@@ -88,7 +88,7 @@ A user wants to link records from another table into their current table. Both t
 
 ## Scenario 4: User Sees a Preview of the Linked Record in a Column With a Foreign Key Constraint Applied
 
-In most cases, the values displayed within a foreign key column won't serve to identify the associated record. A preview of the linked record in a column with a foreign key constraint applied can help the user make this identification.
+In most cases, the values displayed within a foreign key column won't serve to identify the associated record. A preview of the linked record in a column with a foreign key constraint applied can help the user identify it.
 
 ### Scenario 4a: The option for record preview is enabled
 
@@ -130,10 +130,27 @@ Linked records can be changed or removed according to the user's preference. Sin
   - An text input field is available within the dropdown to filter the records based on partial or complete value matches. As the user types in a value, the list narrows down the records. The system can search only the first five fields. Search across all types is supported.
   - If there's a single match for the entered value, it should become highlighted, allowing the user to confirm the selection.
 
+### Scenario 5 Questions
+
+- What if the PK field is not one of the first five fields?
+
 ## Components
 
 ### Record Selector
 
+The record selector component is used to retrieve records from other tables and add them as values to cells according to the referenced column.
+
 ### Table Selector
 
+The table selector allows users to select tables from a schema. This spec doesn't go into details on how this selector would work. It also doesn't consider the scenario for selecting both tables and views.
+
 ### Record Preview
+
+The record preview allows users to identify records linked in other tables. It shows the first five field names and values. A specific limit for the value length needs to be defined so that space usage is optimal.
+
+## Related Discussions
+
+- [Editing values for columns with a foreign key constraint applied](https://github.com/centerofci/mathesar/discussions/796)
+- [Handling duplicate foreign key columns from the 'Link Table' dialog](https://github.com/centerofci/mathesar/discussions/791)
+- [Help users link tables via question-based forms](https://github.com/centerofci/mathesar/discussions/790)
+- [Handling Composite Primary Keys and Junction Tables](https://github.com/centerofci/mathesar/discussions/804)
--- a/design/specs/foreign-key-constraints.md
+++ b/design/specs/foreign-key-constraints.md
@@ -7,37 +7,34 @@ tags:
 editor: markdown
 dateCreated: 2021-10-19T09:20:55.088Z
 ---
+## Context
 
-> This spec is a work in progress and shouldn't be used for implementation.
-{.is-warning}
+## Prototype Link and Video Walkthrough
 
-# Context
+## Scenarios
 
-# Prototype Link and Video Walkthrough
+### Scenario 1 : User Adds a Foreign Key Constraint
 
+A user wants to link records from another table into their current table. Both tables belong to the same schema.
 
-# Scenarios
+### Scenario 1a: Automatically from the 'Link Table' dialog
 
-## User Adds a Foreign Key Constraint
+#### Steps for Scenario 1a
 
-### Automatically from the 'Link Table' dialog
-
-#### Steps
-- A user wants to link values from a table to another table belonging to the same schema.
 - The user starts the 'Link Table' process by clicking on the 'Link Table' button in the table toolbar area.
-- The user reads the instructions in the 'Link Table' dialog and understands that the tables will be linked by setting up a foreign key constraint. They also understand that manual configuration is available. 
-- The user selects the table they wish to link to.
-- The user answers 'yes' or 'no' to the questions presented. The answers will determine the location of the foreign key or whether a new table needs to be created
-	
-  - Answering 'yes' to both questions will set up a mapping table with foreign key columns for both tables, creating a many-to-many relationship
+- The user reads the instructions in the 'Link Table' dialog and understands that the tables will be linked by setting up a foreign key constraint. They also understand that manual configuration is available.
+- The user selects the table they wish to link to from the [table selector](#table-selector).
+- A list of questions is displayed, the user can answer 'yes' or 'no' depending on the type of relationship they want to create.
+- The user answers all the questions listed. The answers will determine the location of the foreign key or whether a new table needs to be created.
+  - Answering 'yes' to both questions will set up a mapping table with foreign key columns for both tables, creating a many-to-many relationship.
   - Answering 'yes' to any of the questions and 'no' to the other will set up a foreign key column in the appropriate table, creating a one-to-many relationship. The column is added to the table on the 'many' side of the relationship.
-
 - Once questions are answered,  a summary of the system's changes is displayed in a section titled 'Under the Hood.'
 - Before creating the link, the user will have the chance to rename the new columns or tables. Invalid column or table names should display an inline warning and prevent saving.
 
-### Manually from the 'Table Constraints' settings
+### Scenario 1b: Manually from the 'Table Constraints' settings
 
-#### Steps
+#### Steps for Scenario 1b
+
 - A user wants to set up a foreign key constraint to single or multiple columns to a currently active table.
 - The user opens the table options by clicking on the table name label in the toolbar area. From the menu, the user selects the option 'Table Constraints'
 - The user sees a layout with two panels. On one side, all the existing constraints are listed. On the other side, a form with the constraints configuration is displayed. The form corresponds to the currently active list item.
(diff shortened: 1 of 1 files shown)


design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
-- Click on 'Add new Column,' and a new column is added with text type by default
-#### The user opens the new column menu and selects the 'Link to Another Table' option
-- Click on the column header to show the menu and select the 'Link to Another Table' option
+- The user starts the 'Link Table' process by clicking on the 'Link Table' button in the table toolbar area.
+- The user reads the instructions presented within the [link table](#link-table) component and understands that the tables will be linked by setting up a foreign key constraint. They also understand that manual configuration is available.
+- The user selects the table they wish to link to from the [table selector](#table-selector).
+- A list of questions is displayed. The user can answer 'yes' or 'no' depending on the relationship they want to create.
+- The user answers all the questions listed. The answers will determine the location of the foreign key or whether a new table needs to be created.
+  - Answering 'yes' to both questions will set up a mapping table with foreign key columns for both tables, creating a many-to-many relationship.
```
kgodey, line 43: We also need to handle the case where the answer is no for both questions. In this case, we can create the FK on the table the user is currently viewing and also set a unique constraint on that column so that each record can only be linked to another, unique, record.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+## Scenario 4: User Sees a Preview of the Linked Record in a Column With a Foreign Key Constraint Applied
+
+In most cases, the values displayed within a foreign key column won't serve to identify the associated record. A preview of the linked record in a column with a foreign key constraint applied can help the user identify it.
+
+### Scenario 4a: The option for record preview is enabled
+
+#### Steps for 4a
+
+- The user opens a table that contains columns with a foreign key constraint applied.
+  - The column contains the referenced values for each cell. The first five fields are displayed.
```
kgodey, line 112: I've been thinking about this more and I think we should have a better algorithm for this than the first five fields. I do think we should show five fields, but I don't think we should rely on the order.
- I think we should prioritize columns that have unique constraints applied.
- We should also prioritize text columns.

We should also allow the user to change which columns we use. They should be able to set (up to) 5 columns to use for search and display of record preview. This should be set for the whole column. We would need a UI for this.

kgodey (commented):
I think this looks good overall. Added some comments with further thoughts.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+- The user reads the instructions presented within the [link table](#link-table) component and understands that the tables will be linked by setting up a foreign key constraint. They also understand that manual configuration is available.
+- The user selects the table they wish to link to from the [table selector](#table-selector).
+- A list of questions is displayed. The user can answer 'yes' or 'no' depending on the relationship they want to create.
+- The user answers all the questions listed. The answers will determine the location of the foreign key or whether a new table needs to be created.
+  - Answering 'yes' to both questions will set up a mapping table with foreign key columns for both tables, creating a many-to-many relationship.
+  - Answering 'yes' to any of the questions and 'no' to the other will set up a foreign key column in the appropriate table, creating a one-to-many relationship. The column is added to the table on the 'many' side of the relationship.
+- Once questions are answered,  a summary of the system's changes is displayed in a section titled 'Under the Hood.'
```
seancolsen, line 45: Regarding the diagram, I'd like to make sure we only display it once both questions are answered.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
----
+## Scenario 2: User Edits an Existing Foreign Key Constraint
+
+### Scenario 2a: From the column header menu
+
+#### Steps for 2a
+
+- The user hovers over the foreign key reference link below the column name.
+- Information about the constraints applied to the column is displayed in a tooltip.
+- The user clicks on the foreign key reference link below the column name.
```
seancolsen, line 70: In #786 @pavish changed the column header to trigger the column header menu when clicking anywhere inside the header (instead of clicking only on the arrow), which I like. With this spec, we'd move from having one interaction inside the column header to having three: the column menu, the constraint hover, and the constraint edit. I think that's too many. Plus, I don't like the idea of having separate click/hover actions on the same screen area because once we begin to support touch devices we'll lose the hover functionality. 

I'd prefer to move the functionality from the tooltip to be inside the column header menu. Clicking anywhere in the column header would bring up the column header menu. When the column has an FK constraint, there would be a section in that menu to explain it. The user could click somewhere within that section to bring up the constraints modal and edit the constraint.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+### Scenario 3a: The foreign key constraint is set to a single column
+
+#### Steps for 3a
+
+- The user opens a table that contains columns with a foreign key constraint applied.
+- The user identifies the columns in two ways:
+  - The user looks at the column header and sees the foreign key indicator. This indicator shows a key icon and the referenced table and column next to it.
+    - The user can click the indicator to display the foreign key settings for the specific column.
+    - The indicator can be hovered over to display a tooltip with information about the foreign key constraints.
+  - The user looks at the column fields and sees that values are added with different styling, a tag-like element with a colored background.
```
seancolsen, line 92: In the loom you mentioned the idea of using different colors for different tables. That's a neat idea, but I'd prefer we skip that at first for the sake of simplicity.

Using dynamic colors introduces some additional UX complexities such as

- How do we theme those colors for light-mode/dark-mode?
- How do we ensure the colors are sufficiently distinct?
- If tables A and B both have FKs to table C, then do they they both use the same color for table C?
- Can the user configure the colors?

All of these questions are solvable of course, but it's low-priority compared to the other tasks we have, so I say let's do it later.

design/specs/foreign-key-constraints.md
seancolsen: Agreed. Especially important if the user can't reorder the columns in a table.

dmos62:
Nice work! I've a few quick notes. These may have been addressed already.

#### Additional options and defaults

Postgres provides some options for FKs, like what to do when a referenced column is deleted or updated, or whether to even allow that (`ON UPDATE`, `ON DELETE`). Also, there's an option for whether or not a multi-column FK can have referencing cells that are undefined (`MATCH FULL`, `MATCH SIMPLE`).

For the sake of development speed we might just want to use a sensible default at the moment. If so, it might be nice to tell the user what that default is. I think defaults were discussed at some point and we arrived at `ON UPDATE CASCADE`, `ON DELETE RESTRICT`, and `MATCH SIMPLE`.

#### Terminology

SQL community's terms around FK are pretty confusing and not well established (I've seen referencing, referred, source, target, columns, etc.). When you're looking at a different SQL library or a different documentation, chances are that it will be calling these properties something new.

Concern here is that when the user will be looking at the create/edit FK constraint dialog it might be taxing for him to understand what do the terms referenced table, reference column, column (no prefix) mean. I'd say there isn't an easy solution for this: we either make it verbose or short but confusing. I'm leaning towards verbosity: referred table/column and referring table/column. Referred/referring being popular abbreviations for referenced/referencing.

Also, in the context of FKs, saying column or table without specifying whether we're talking about the referred or referring column/table is somewhat ambiguous. There's a colloquial convention that when you're ambiguous you mean the referring object, but it might take a user some extra time to realise that. Something to think about.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+### Scenario 2a: From the column header menu
+
+#### Steps for 2a
+
+- The user hovers over the foreign key reference link below the column name.
+- Information about the constraints applied to the column is displayed in a tooltip.
+- The user clicks on the foreign key reference link below the column name.
+- Constraint settings open for the column.
+
+### Scenario 2b: From table constraint settings
```
pavish, line 73: The table constraints modal is updated in this spec, but we are working on issues which reference the previous figma designs which are very different.

We need to update the spec for the new table constraints modal. We also need to define how to manipulate FKs from within this modal. It is not very clear from the loom videos.

design/specs/foreign-key-constraints.md
pavish: @ghislaineguerin I like showing the referenced table on the header. But I agree with @seancolsen that we should not open the constraints modal by clicking on it.
* Editing constraints is a rarely performed operation. But opening the column dropdown is essential for anything users would want to do with the column. So, it makes sense to only use the entire column header space to open the dropdown.
* As Sean described, we can show it within the dropdown, have a section for it and open the constraints modal when user clicks on it.

design/specs/foreign-key-constraints.md
pavish: I think these can be on the spec, but we can discuss and de-prioritize them when working on the issues. We could probably link this comment to the issue, it's a conversation better placed there.

design/specs/foreign-key-constraints.md
pavish: +1 on this

pavish (commented):
@ghislaineguerin This looks really good and detailed. I had a couple comments.

design/specs/foreign-key-constraints.md
kgodey: I agree that we can keep them in the spec. I'll make sure to create a separate issue for implementing the colors that we can defer if it makes sense.

design/specs/foreign-key-constraints.md
mathemancer: Agree with folks on this.  Tricky things to consider:
- Some primary key columns would be useful, some very not useful (e.g., the one we create).  Text helps with this, but there may also be other considerations.
- Multi-column unique constraints are potentially useful, but only if all columns are included.
- non-unique columns are _never_ useful unless a unique column is included (that gives info for the user to choose between rows with duplicate entries in non-unique columns).

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+- This opens a dialog from which the user can set all supported constraint types.
+- The user sees a layout with two panels. On one side, all the existing constraints are listed. On the other side, a form with the constraints configuration is displayed. The form corresponds to the currently active list item.
+- The list panel contains actions at the bottom of the panel. From there, the user can choose to add or delete an item from the list. The user clicks on 'Add' to create a new constraint.
+- The user is presented with a form. From there, they select the type of constraint to be added. The user selects the 'Foreign Key' constraint option from the list.
+- The user selects single or multiple columns they wish to apply the constraint to.
+- The user selects a table to be referenced by these columns.
+- The user selects a column in the reference table from which the column will match values. By default, this is set to the 'Primary Key' column in the referenced table. When changing this field, the user can select only columns with a UNIQUE or PRIMARY KEY constraint.
 
----
+## Scenario 2: User Edits an Existing Foreign Key Constraint
```
mathemancer, line 62: At the DB level, altering foreign keys isn't possible (at least for the things we're actually defining for foreign key constraints through the UI).  Is it really needed?  The DB process to do this would be to create a new foreign key constraint, and drop the old.  This would need to involve some transactions that would make doing it step-by-step ill-advised, since it would involve repetitively creating and dropping foreign key constraints.  These operations would each lock multiple tables for reading and writing.

Thus, I think we should consider focusing on helping users go through the "drop constraint", then "create new constraint" process smoothly instead.  This will avoid confusion, in particular around potential errors arising from a different process occurring under the hood than is portrayed by the UI.

design/specs/foreign-key-constraints.md
mathemancer, line 92: Are the colors per-foreign-table, or just to denote "foreign key"?  I think it would be cool if I could tell at a glance which table was linked by color (could get too busy though).

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+    - The indicator can be hovered over to display a tooltip with information about the foreign key constraints.
+  - The user looks at the column fields and sees that values are added with different styling, a tag-like element with a colored background.
+
+### Scenario 3b: The foreign key constraint is set to multiple columns
+
+#### Steps for 3b
+
+- The user opens a table that contains columns with a multi-column foreign key constraint applied.
+  - The indicator, in this case, looks similar to the single-column one. However, the user will understand that the foreign key constraints are referencing the same table.
+    - The user can tell that a referenced table is the same because they share the same color. The system could use other UI elements instead of color. Still, the idea is to create a visual distinction that can be easily scanned.
```
mathemancer, line 100: This seems to hint that the colors will be per-foreign-table, which is cool.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,169 @@
...
+  - The indicator can be hovered over to see additional information. From here, the user will understand that the constraint references more than one column.
+
+## Scenario 4: User Sees a Preview of the Linked Record in a Column With a Foreign Key Constraint Applied
```
mathemancer, line 103: We'll have a similar UX challenge (identifying the linked records) for multi-column foreign key constraints, but I don't understand whether this UI will assist with that.

mathemancer (commented):
Overall, I think this looks pretty good.  I did leave some specific concerns on different parts of the write-up.

My biggest concern is that there isn't a way to edit foreign keys (at least w.r.t. the included columns, referenced table, or the constraint name) at the DB level.  Thus, portraying to the user that they're modifying an existing foreign key constraint may lead to error states that would be confusing, and which may cause data inconsistencies.  I strongly suggest we simplify and streamline the process of dropping a current foreign key constraint and creating a new one (or creating the new then dropping the old) instead.

dmos62 (commented):
No explicit request or approval, just [above comment](https://github.com/centerofci/mathesar-wiki/pull/12#issuecomment-970307540).

design/specs/foreign-key-constraints.md
ghislaineguerin: @kgodey I have updated both the prototype and spec to include the one-to-one scenario

design/specs/foreign-key-constraints.md
silentninja: Based on the loom video, the colours are per-foreign-key-table. I like how the colours are different based on the foreign key table rather than the foreign key column.

silentninja: Based on the loom video, the colours are per-foreign-key-table. I like how the colours are different based on the foreign key table rather than the foreign key column.

design/specs/foreign-key-constraints.md
silentninja: Agree with @mathemancer. The foreign key constraint should not be updatable and should rather point to deleting the constraint and creating a different one. Instead of apply button something like delete and copy template(in case if all they want is to include a constraint on an additional column from the same table)

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,170 @@
...
+
+- The user opens a table that has columns with a foreign key constraint applied.
+  - The column contains no values.
+  - When clicked, a list of records is displayed inside the [record selector](#record-selector) component. The list shows the first 50 records.
+  - An text input field is available within the dropdown to filter the records based on partial or complete value matches. As the user types in a value, the list narrows down the records. The system can search only the first five fields. Search across all types is supported.
+  - If there's a single match for the entered value, it should become highlighted, allowing the user to confirm the selection.
+
+### Scenario 5 Questions
+
+- What if the PK field is not one of the first five fields?
```
silentninja, line 150: There could be an option to do a search across all columns(wouldn't want to show this only if no rows are found as the returned row can be something the user is not looking for, but not sure how to restrict them from not using it often).

Having the user toggle the columns to search for can be another alternate

design/specs/foreign-key-constraints.md
mathemancer: I see that now.  I also like it.

design/specs/foreign-key-constraints.md
ghislaineguerin: I've added a note to add this to a future design issue. For now, we'll assume it's a single-column constraint.

design/specs/foreign-key-constraints.md
ghislaineguerin: I agree, I have removed the additional click target from the column header.

design/specs/foreign-key-constraints.md
ghislaineguerin: I will add a note to the spec regarding a future issue to define how we help users select appropriate columns. However, at this point, I don't want to block them or prevent them from choosing columns as they might not have set a unique constraint for a column, that in fact, has unique values. This might be a common thing when they start from an imported CSV. It will be related to how we might detect that and recommend that they apply a constraint etc. For now, we could assume that a column with no unique constraint could still be valuable as a search column.

design/specs/foreign-key-constraints.md
ghislaineguerin: Agree, I'll keep this in the spec. We can define how we want to apply color once design for Views is complete.

design/specs/foreign-key-constraints.md
ghislaineguerin: I agree. I think we'll need a separate issue for this, as @mathemancer had some concerns about editing foreign key constraints, rather than dropping them and creating new ones. So we need to figure out how exactly we want to handle these beyond just the interface.

design/specs/foreign-key-constraints.md
ghislaineguerin: A note about the usage of color will be added to the spec. Once we have a design for views we can discuss implementation details.

design/specs/foreign-key-constraints.md
ghislaineguerin: For now, the idea will be to point users towards changing the search columns if no results are found. I think your concern about a 'search across all' might encourage the behavior you describe, which might not be ideal for performance.

design/specs/foreign-key-constraints.md
kgodey: All constraints.

design/specs/foreign-key-constraints.md
```diff
@@ -176,6 +180,14 @@ The record preview allows users to identify records linked in other tables. It s
 - Multi-column unique constraints are potentially useful, but only if all columns are included.
 - Non-unique columns are never useful unless a unique column is included (that gives info for the user to choose between rows with duplicate entries in non-unique columns).
 
+### Linked Records for Multi-Column Foreign Key Constraints
+
+- The current UI does not take into account multi-column foreign key constraints and how those would be retrieved and selected.
```
kgodey, line 185: @ghislaineguerin Users might connect a database with multi-column foreign key constraints. How will we treat these? I assume we won't show any autocomplete behavior but it would be good to make that explicit.

ghislaineguerin: @kgodey can we say 'Unsupported table link' instead of showing the regular table link options on the column header menu

kgodey:
@ghislaineguerin I updated the status of this PR to review and added everyone to the assignee list.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,209 @@
...
+
+#### Steps for 5b
+
+- The user opens a table that has columns with a foreign key constraint applied.
+  - The column contains no values.
+  - When clicked, a list of records is displayed inside the [record selector](#record-selector) component. The list shows the first 50 records.
+  - An text input field is available within the dropdown to filter the records based on partial or complete value matches. As the user types in a value, the list narrows down the records. The system can search only the first five fields. Search across all types is supported.
+  - If there's a single match for the entered value, it should become highlighted, allowing the user to confirm the selection.
+  - If the table contains no records, the user sees an option to add a record.
+  - If there aren't any matches, the user sees an option to add a record for the newly entered value or try changing the 'Search Columns' table preference.
```
seancolsen, line 167: What happens after the user clicks "Add Record" here?

seancolsen (approved):
Aside from one clarifying question, this looks great!

design/specs/foreign-key-constraints.md
ghislaineguerin: @seancolsen we haven't defined that part of the flow yet. I assume we'll know more when we work on views as it also involves the creation of new records in other tables. I'll add a note for this.

mathemancer (commented):
Looks great overall.

Small comment:  I noticed in the video for Scenario 2 that we "remove" the foreign key constraint from some parts of the UI, but we "delete" the foreign key constraint from others.  I think these should be unified to one specific word/concept to avoid confusion, and that word/concept should be chosen carefully.  We want to make sure the User understands they're not deleting any data when they take the action, but only the attachment between the tables.

seancolsen:
@mathemancer @ghislaineguerin  I added a [discussion about "delete" vs "remove"](https://github.com/centerofci/mathesar/discussions/872)

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,209 @@
...
-- Click on the column header to show the menu and select the 'Link to Another Table' option
+### Scenario 1a: Automatically adding foreign key constraints from the 'Link Table' dialog
 
-**Note: This step does not cover the scenario where a user has a unique table, in which case they need to create an additional table to add a relationship.**
+#### Assumptions for 1a
 
-#### 2. The user selects a table from the list of available tables
-- The user selects the table on the 'one side of the 'one-to-many' relationship.
+- The user is not familiar with the concept of foreign keys.
+- The user has created tables within Mathesar, where the ID column is automatically generated and sent as the primary key.
```
kgodey, line 44: I don't think this is a good assumption to make. We should handle cases for all tables, even without primary keys.

We can solve this for the first version by not supporting creating links through scenario 1a. Perhaps we could have a little help message that says "Why don't I see all my tables in this list?" and explain that we only show tables with primary keys.

design/specs/foreign-key-constraints.md
kgodey: I think we should remove this option from the spec, we can add it in once we have it defined. It'll confuse implementation.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,209 @@
...
+
+### Record Preview
+
+The record preview allows users to identify records linked in other tables. It shows the first five field names and values. A specific limit for the value length needs to be defined so that space usage is optimal.
+
+## Other Considerations
+
+### Linked Records for Multi-Column Foreign Key Constraints
+
+- The current UI does not consider multi-column foreign key constraints and how those would be retrieved and selected.
```
kgodey, line 199: I think we should define this a bit, even if only to say that we treat it like a text column for now.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,209 @@
...
+
+### Usage of Color
+
+- A suggestion to use color as a means to differentiate table references has been included in this spec. However, there are implementation details that need to be discussed before the team can make a decision.
```
kgodey, line 203: Decision about what?

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,209 @@
...
+
+- The user opens the column menu for a column that has a foreign key constraint applied
+- The user clicks on the 'Table Link Preferences' option
+
+### Scenario 6b: From the record selector menu
+
+- The user opens the column menu for a column that has a foreign key constraint applied
+- The user clicks on the preferences link
+
+## Components
```
kgodey, line 181: What about the constraints modal component?

kgodey (changes requested):
@ghislaineguerin This looks great overall. Some (hopefully) minor comments.

design/specs/foreign-key-constraints.md
silentninja: I think it would be better if we show all the tables, but disable the table without a primary key with a text hint that would direct the user to create a primary key if they wish to link it to that particular table.

design/specs/foreign-key-constraints.md
ghislaineguerin: @kgodey I've updated to add more detail.

design/specs/foreign-key-constraints.md
ghislaineguerin: @kgodey I've added additional information.

design/specs/foreign-key-constraints.md
ghislaineguerin: I agree with @silentninja. I think it would be best to show the tables and prevent selection. I will add details for this in the spec.

kgodey:
@ghislaineguerin Please re-assign me when the spec is ready for re-review.

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,206 @@
...
+
+- The current UI does not consider multi-column foreign key constraints and how those would be retrieved and selected. Record search and selection via the dropdown component will not be supported. Instead, the cells will be limited to input and edit text values. The text values will have to match the primary keys of the referenced records and be unique.
+
+### Usage of Color
+
+- A suggestion to use color as a means to differentiate table references has been included in this spec. However, implementation details need to be discussed before deciding if the user interface will use color elements to create a visual connection between elements.
+
+### Terminology
+
+- `Remove` and `Delete` are used throughout this spec and should be applied consistently across the entire user interface as they are different affordances. `Remove` will be used for data that can be added back, while `Delete` will indicate data destruction.
```
kgodey, line 199: We actually decided to use Delete for everything in https://github.com/centerofci/mathesar/discussions/872

design/specs/foreign-key-constraints.md
```diff
@@ -2,75 +2,206 @@
...
-- Click on the column header to show the menu and select the 'Link to Another Table' option
+### Scenario 1a: Automatically adding foreign key constraints from the 'Link Table' dialog
 
-**Note: This step does not cover the scenario where a user has a unique table, in which case they need to create an additional table to add a relationship.**
+#### Assumptions for 1a
 
-#### 2. The user selects a table from the list of available tables
-- The user selects the table on the 'one side of the 'one-to-many' relationship.
+- The user is not familiar with the concept of foreign keys.
+- The user has at least two tables with primary key constraints set.
```
kgodey, line 43: I think we just need one table - the one that's being linked to. And it should be a single primary key and not a composite primary key.

kgodey (changes requested):
@ghislaineguerin Looks good overall, a couple of minor changes requested.

design/specs/foreign-key-constraints.md
ghislaineguerin: I agree. I've updated the assumptions.

design/specs/foreign-key-constraints.md
```diff
@@ -40,7 +40,7 @@ A user wants to link records from another table into their current table. Both t
 #### Assumptions for 1a
 
 - The user is not familiar with the concept of foreign keys.
-- The user has at least two tables with primary key constraints set.
+- The user has at least a tables with a primary key constraint set.
```
kgodey, line 43:
```suggestion
- The user has at least a table with a primary key constraint set.
```

kgodey (commented):
@ghislaineguerin I noticed a typo, made a suggested fix.

kgodey (approved):
Looks good other than my one suggested change.

PR #13, opened 2021-11-25
Handling Null Values Spec

ghislaineguerin:
This PR adds the specs to implement the design for [UX to allow user to set value to NULL within a cell](https://github.com/centerofci/mathesar/issues/780)

### DIFF
--- a/design/specs/specs-null-values.md
+++ b/design/specs/specs-null-values.md
@@ -10,7 +10,7 @@ dateCreated: 2021-11-25T09:00:00.201Z
 
 ## Context
 
-NULL values are used to indicate that a value does not exist in a database. It is important to make it clear to users that NULL is not the same as blank or zero. Ideally, we want users to think of NULL values as placeholders for unknown or missing values that are applicable.
+NULL values indicate that a value does not exist in a database. It is important to make it clear that NULL is not the same as blank or zero. Ideally, we want users to think of NULL values as placeholders for unknown or missing values that are applicable.
 
 ## Prototype
 
@@ -28,7 +28,7 @@ NULL values are used to indicate that a value does not exist in a database. It i
 
 ### Considerations for handling NULL Values
 
-When presented in the form of tabular data, NULL values should be visually distinct so that the user understands that they are not the same as other string types. The style should attempt to communicate the absence of a value, similar to how placeholders are used in text input fields.
+When presented in tables, NULL values should be visually distinct so that the user understands that they are not the same as other string types. The style should attempt to communicate the absence of a value, similar to how placeholders are used in text input fields.
 As for the process of transforming values to and from NULL, it will be important to have a consistent way of doing this across different data types.
 
 ## Scenarios
@@ -37,7 +37,7 @@ As for the process of transforming values to and from NULL, it will be important
 
 #### Scenario 1a: The user identifies a NULL value in a table
 
-#### Steps for 1a
+##### Steps for 1a
 
 1. The user opens a table that contains NULL values
 2. The user identifies a cell that contains a NULL value
@@ -48,7 +48,7 @@ As for the process of transforming values to and from NULL, it will be important
 
 #### Scenario 1b: The user identifies a NULL value in a record preview
 
-#### Steps for 1b
+##### Steps for 1b
 
 1. The user edits a value in a column that links to records from another table
 2. The user sees the list of records displayed in the record selection dropdown
@@ -61,11 +61,11 @@ As for the process of transforming values to and from NULL, it will be important
 
 #### Scenario 2a: The data type is text
 
-#### Steps for 2a
+##### Steps for 2a
 
 1. The user edits a NULL value in a column that is set to text data type
 2. The user double-clicks the value cell and enters edit mode
-3. The user enters a value and leaves edit mode by hitting enter or clicking outside of the active cell
+3. The user enters a value and leaves edit mode by pressing the enter key or clicking outside of the active cell
 4. The NULL value is replaced by the new value
 
 #### Scenario 2b: The data type is text, and the value is empty
@@ -74,50 +74,53 @@ As for the process of transforming values to and from NULL, it will be important
 
 1. The user edits a NULL value in a column that is set to text data type
 2. The user double-clicks the value cell and enters edit mode
-3. The user doesn't enter a value and leaves edit mode by hitting enter or clicking outside of the active cell
-4. The NULL value is replaced by an empty string
+3. The user sees a text input with a placeholder containing the word 'NULL.'
+4. The user doesn't enter a value, and the value replaces the placeholder text
+5. The user deletes the new value and leaves edit mode by hitting enter or clicking outside of the active cell
+6. The NULL value is replaced by an empty string
 
 #### Scenario 2c: The data type is boolean with checkbox display enabled
 
-#### Steps for 2c
+##### Steps for 2c
 
 1. The user edits a NULL value in a column that is set to boolean type with the checkbox display property enabled
     - The field contains a checkbox in an indeterminate state
+    - Hovering over the NULL checkbox will display a tooltip with the 'NULL' legend.
 2. The user clicks on the checkbox, going from indeterminate (NULL) to checked (TRUE).
-3. The user can click on the checkbox again to mark it as false. However, it does not go back to NULL.
+3. The user can click on the checkbox again to mark it as false and go back to NULL.
 
 #### Scenario 2d: The data type is number
 
-#### Steps for 2d
+##### Steps for 2d
 
-1. The user edits a NULL value in a column that is set to number type
-2. The user double-clicks the cell to enter edit mode
-3. The user exists in edit mode without making any changes
-4. The NULL value remains
+1. The user edits a NULL value in a column set to number type.
+2. The user double-clicks the cell to enter edit mode.
+3. The user exists in edit mode without making any changes.
+4. The NULL value remains.
 
 ### Scenario 3: User sets a value to NULL
 
 #### Scenario 3a: The column accepts NULL values
 
-#### Steps for 3a
+##### Steps for 3a
 
-1. The user selects a cell in a column that contains a value
-2. The user clears the content of the selected cell
-    - By pressing the delete key
-    - By opening the contextual menu and selecting the 'Set as NULL' option
-3. The selected cell now has a NULL value
+1. The user selects a cell in a column that contains a value.
+2. The user clears the content of the selected cell.
+    - By pressing the delete key.
+    - By opening the contextual menu and selecting the 'Set as NULL' option.
+3. The selected cell now has a NULL value.
 
 #### Scenario 3b: The column does not accept NULL values
 
-#### Steps for 3b
+##### Steps for 3b
 
-1. The user selects a cell in a column that contains a value
-2. The user attempts to clear the content of the selected cell
-    - By pressing the delete key
-        - The system prevents the insertion of a NULL value
-        - An error message indicates to the user that a NOT NULL constraint is applied to the column, and the system cannot insert NULL values
-    - By opening the contextual menu and selecting the 'Set as NULL' option
-        - The 'Set as NULL' option is disabled, and the user cannot select it
+1. The user selects a cell in a column that contains a value.
+2. The user attempts to clear the content of the selected cell.
+    - By pressing the delete key.
+        - The system prevents the insertion of a NULL value.
+        - An error message indicates that a NOT NULL constraint is applied to the column, and the system cannot insert NULL values.
+    - By opening the contextual menu and selecting the 'Set as NULL' option.
+        - The 'Set as NULL' option is disabled, and the user cannot select it.
 
 ## Related Discussions
 
--- a/design/specs/specs-null-values.md
+++ b/design/specs/specs-null-values.md
@@ -118,3 +118,7 @@ As for the process of transforming values to and from NULL, it will be important
         - An error message indicates to the user that a NOT NULL constraint is applied to the column, and the system cannot insert NULL values
     - By opening the contextual menu and selecting the 'Set as NULL' option
         - The 'Set as NULL' option is disabled, and the user cannot select it
+
+## Related Discussions
+
+- [UX for NULL Values](https://github.com/centerofci/mathesar/discussions/832)
--- a/design/specs/specs-null-values.md
+++ b/design/specs/specs-null-values.md
@@ -1,5 +1,5 @@
 ---
-title: Null Value Specs
+title: NULL Value Specs
 description: 
 published: true
 date: 2021-11-25T09:00:00.201Z
@@ -10,7 +10,7 @@ dateCreated: 2021-11-25T09:00:00.201Z
 
 ## Context
 
-NULL values indicate that a value does not exist in a database. It is important to make it clear that NULL is not the same as blank or zero. Ideally, we want users to think of NULL values as placeholders for unknown or missing values that are applicable.
+`NULL` values indicate that a value does not exist in a database. It is important to clarify that `NULL` is not the same as blank or zero. Ideally, we want users to think of `NULL` values as placeholders for unknown or missing values that are applicable.
 
 ## Prototype
 
@@ -26,102 +26,118 @@ NULL values indicate that a value does not exist in a database. It is important 
 
 ## User Experience
 
-### Considerations for handling NULL Values
+### Considerations for handling `NULL` Values
 
-When presented in tables, NULL values should be visually distinct so that the user understands that they are not the same as other string types. The style should attempt to communicate the absence of a value, similar to how placeholders are used in text input fields.
-As for the process of transforming values to and from NULL, it will be important to have a consistent way of doing this across different data types.
+When presented in tables, `NULL` values should be visually distinct so that the user understands that they are not the same as other string types. The style should attempt to communicate the absence of a value, similar to how placeholders are used in text input fields.
+As for the process of transforming values to and from `NULL`, it will be important to have a consistent way of doing this across different data types.
 
 ## Scenarios
 
-### Scenario 1: User identifies a NULL Value
+### Scenario 1: User identifies a `NULL` Value
 
-#### Scenario 1a: The user identifies a NULL value in a table
+#### Scenario 1a: The user identifies a `NULL` value in a table
 
 ##### Steps for 1a
 
-1. The user opens a table that contains NULL values
-2. The user identifies a cell that contains a NULL value
-    - The cell contains the word NULL in all uppercase letters
+1. The user opens a table that contains `NULL` values.
+2. The user identifies a cell that contains a `NULL` value.
+    - The cell contains the word `NULL` in all uppercase letters
     - The cell has a lighter background-color
     - The cell has a muted text color
     - The cell has a different font style (italic)
 
-#### Scenario 1b: The user identifies a NULL value in a record preview
+#### Scenario 1b: The user identifies a `NULL` value in a record preview
 
 ##### Steps for 1b
 
-1. The user edits a value in a column that links to records from another table
-2. The user sees the list of records displayed in the record selection dropdown
-3. The user identifies a record that contains a NULL value
-    - The field contains the word NULL in all uppercase letters
+1. The user edits a value in a column that links to records from another table.
+2. The user sees the list of records displayed in the record selection dropdown.
+3. The user identifies a record that contains a `NULL` value.
+    - The field contains the word `NULL` in all uppercase letters
     - The text has a muted text color
     - The text has a different font style (italic)
 
-### Scenario 2: User edits a NULL Value
+### Scenario 2: User edits a `NULL` Value
 
 #### Scenario 2a: The data type is text
 
 ##### Steps for 2a
 
-1. The user edits a NULL value in a column that is set to text data type
-2. The user double-clicks the value cell to enter edit mode
-3. The user enters a value and leaves edit mode by pressing the enter key or clicking outside of the active cell
-4. The NULL value is replaced by the new value
+1. The user edits a `NULL` value in a column that is set to text data type.
+2. The user double-clicks the value cell to enter edit mode.
+3. The user enters a value and leaves edit mode by pressing the enter key or clicking outside of the active cell.
+4. The `NULL` value is replaced by the new value.
 
 #### Scenario 2b: The data type is text, and the value is cleared
 
 #### Steps for 2b
 
-1. The user edits a NULL value in a column that is set to text data type
-2. The user double-clicks the value cell to enter edit mode
-3. In edit-mode, a text input box is displayed with a placeholder set to 'NULL'
-4. The user types in a value and the placeholder text is replaced by the new typed-in value
-5. The user clears the new value by hitting the backspace key until the value is completely cleared
-6. The user exits edit-mode
-7. The NULL value is replaced by an empty string and the cell is blank
+1. The user edits a `NULL` value in a column that is set to text data type.
+2. The user double-clicks the value cell to enter edit mode.
+3. In edit mode, a text input box is displayed with a placeholder set to `NULL`.
+4. The user types in a value, and the placeholder text is replaced by the new typed-in value.
+5. The user clears the new value by hitting the backspace key until the value is completely cleared.
+6. The user exits edit mode.
+7. The `NULL` value is replaced by an empty string, and the cell is blank.
 
 #### Scenario 2c: The data type is boolean with checkbox display enabled
 
 ##### Steps for 2c
 
-1. The user edits a NULL value in a column that is set to boolean type with the checkbox display property enabled
-    - The field contains a checkbox in an indeterminate state
-    - Hovering over the NULL checkbox will display a tooltip with the 'NULL' legend.
-2. The user clicks on the checkbox, going from indeterminate (NULL) to checked (TRUE).
-3. The user can click on the checkbox again to mark it as false and go back to NULL.
+1. The user edits a `NULL` value in a column that is set to boolean type with the checkbox display property enabled.
+    - The field contains a checkbox in an indeterminate state.
+    - Hovering over the `NULL` checkbox will display a tooltip with the '`NULL`' legend.
+2. The user clicks on the checkbox, going from indeterminate (`NULL`) to checked (TRUE).
+3. The user can click on the checkbox again to mark it as false and go back to `NULL`.
 
 #### Scenario 2d: The data type is number
 
 ##### Steps for 2d
 
-1. The user edits a NULL value in a column set to number type.
+1. The user edits a `NULL` value in a column set to number type.
 2. The user double-clicks the cell to enter edit mode.
 3. The user exists in edit mode without making any changes.
-4. The NULL value remains.
+4. The `NULL` value remains.
 
-### Scenario 3: User sets a value to NULL
+### Scenario 3: User sets a value to `NULL`
 
-#### Scenario 3a: The column accepts NULL values
+#### Scenario 3a: The column accepts `NULL` values
 
 ##### Steps for 3a
 
 1. The user selects a cell in a column that contains a value.
 2. The user clears the content of the selected cell.
-    - By pressing the delete key.
-    - By opening the contextual menu and selecting the 'Set as NULL' option.
-3. The selected cell now has a NULL value.
+    - By pressing the `delete` key.
+    - By opening the contextual menu and selecting the 'Set as `NULL`' option.
+3. The selected cell now has a `NULL` value.
 
-#### Scenario 3b: The column does not accept NULL values
+#### Scenario 3b: The column does not accept `NULL` values
 
 ##### Steps for 3b
 
 1. The user selects a cell in a column that contains a value.
 2. The user attempts to clear the content of the selected cell.
-    - By pressing the delete key.
-        - The system prevents the insertion of a NULL value.
-        - An error message indicates that a NOT NULL constraint is applied to the column, and the system cannot insert NULL values.
-    - By opening the contextual menu and selecting the 'Set as NULL' option.
-        - The 'Set as NULL' option is disabled, and the user cannot select it.
+    - By pressing the `delete` key.
+        - The system prevents the insertion of a `NULL` value.
+        - An error message indicates that a NOT `NULL` constraint is applied to the column, and the system cannot insert `NULL` values.
+    - By opening the contextual menu and selecting the 'Set as `NULL`' option.
+        - The 'Set as `NULL`' option is disabled, and the user cannot select it.
+
+## Interactions
+
+### edit mode Behavior
+
+The 'edit mode' state of a cell changes some interactions or provides additional functionality to users while editing data in a table. How users enter and exit 'edit mode' must be defined in better detail as more data types and cell-level functionality are introduced.
+
+For the implementation of this spec, we should consider the following notes:
+
+1. User highlights the cell with a single click. Pressing the `delete` or `backspace` key will delete the cell state and convert it to a `NULL` value.
+2. The user presses any key other than `delete` or `backspace` the cell enters edit mode.  Pressing `delete` or `backspace` will delete the text inside the cell.
+
+The same can be applied for boolean types too. The `checkbox` should be smaller than the cell.
+
+1. Clicking outside the `checkbox` would select the cell, and pressing `delete` would make the cell `NULL`
+2. Clicking inside the `checkbox` would toggle the value of the checkbox, similar to spreadsheets.
 
 ## Related Discussions
 
--- a/design/specs/specs-null-values.md
+++ b/design/specs/specs-null-values.md
@@ -116,7 +116,7 @@ As for the process of transforming values to and from `NULL`, it will be importa
 
 1. The user selects a cell in a column that contains a value.
 2. The user clears the content of the selected cell.
-    - By pressing the `delete` key.
+    - By pressing the `Shift`+`Backspace` or `Shift`+`Delete` keys.
     - By opening the contextual menu and selecting the `Set as NULL` option.
 3. The selected cell now has a `NULL` value.
 
@@ -126,7 +126,7 @@ As for the process of transforming values to and from `NULL`, it will be importa
 
 1. The user selects a cell in a column that contains a value.
 2. The user attempts to clear the content of the selected cell.
-    - By pressing the `delete` key.
+    - By pressing the `Shift`+`Backspace` or `Shift`+`Delete` keys.
         - The system prevents the insertion of a `NULL` value.
         - An error message indicates that a NOT `NULL` constraint is applied to the column, and the system cannot insert `NULL` values.
     - By opening the contextual menu and selecting the `Set as NULL` option.
@@ -136,7 +136,9 @@ As for the process of transforming values to and from `NULL`, it will be importa
 
 ### Select vs. Edit Mode Behaviors
 
-In `Edit Mode`, a cell provides additional functionality and interactions to users. `Edit Mode` is toggled by double-clicking on a cell. However, certain data types could have different interactions. Once specific behaviors and components for each data type are introduced, we must further define these interactions to ensure compatibility.
+In `Select Mode`, the cell is highlighted and set as the target for actions applying to the cell or elements associated with it, such as its row or column. `Select Mode` is toggled by clicking on a cell or by navigating to with the direction keys while `Select Mode` is active.
+
+In `Edit Mode`, a cell provides additional functionality and interactions to users. `Edit Mode` is toggled by double-clicking on a cell or by typing a text character or pressing backspace while in `Select Mode`. However, certain data types could have different interactions. Once specific behaviors and components for each data type are introduced, we must further define these interactions to ensure compatibility.
 
 We will create an additional issue to refine these interactions.
 
--- a/design/specs/specs-null-values.md
+++ b/design/specs/specs-null-values.md
@@ -1,0 +1,108 @@
+---
+title: Null Value Specs
+description: 
+published: true
+date: 2021-11-25T09:00:00.201Z
+tags: 
+editor: markdown
+dateCreated: 2021-11-25T09:00:00.201Z
+---
+
+## Handling NULL Values
+
+NULL values are used to indicate that a value does not exist in a database. It is important to make it clear to users that NULL is not the same as blank or zero. Ideally, we want users to think of NULL values as placeholders for unknown or missing values that are applicable.
+
+## User Experience
+
+### Considerations for handling NULL Values
+
+When presented in the form of tabular data, NULL values should be visually distinct so that the user understands that they are not the same as other string types. The style should attempt to communicate the absence of a value, similar to how placeholders are used in text input fields.
+As for the process of transforming values to and from NULL, it will be important to have a consistent way of doing this across different data types.
+
+## Scenarios
+
+### Scenario 1: User identifies a NULL Value
+
+#### Scenario 1a: The user identifies a NULL value in a table
+
+#### Steps for 1a
+
+1. The user opens a table that contains NULL values
+2. The user identifies a cell that contains a NULL value
+    - The cell contains the word NULL in all uppercase letters
+    - The cell has a lighter background-color
+    - The cell has a muted text color
+    - The cell has a different font style (italic)
+
+#### Scenario 1b: The user identifies a NULL value in a record preview
+
+#### Steps for 1b
+
+1. The user edits a value in a column that links to records from another table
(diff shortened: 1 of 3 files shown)


dmos62 (approved):

kgodey:
@dmos62 I unassigned you since you approved the spec.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,124 @@
...
+When presented in the form of tabular data, NULL values should be visually distinct so that the user understands that they are not the same as other string types. The style should attempt to communicate the absence of a value, similar to how placeholders are used in text input fields.
+As for the process of transforming values to and from NULL, it will be important to have a consistent way of doing this across different data types.
+
+## Scenarios
+
+### Scenario 1: User identifies a NULL Value
+
+#### Scenario 1a: The user identifies a NULL value in a table
+
+#### Steps for 1a
```
kgodey, line 40: Just a style issue, I would remove these "Steps" headers because they don't add any extra information and it's odd to have two headers at the same level.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,124 @@
...
+3. The user enters a value and leaves edit mode by hitting enter or clicking outside of the active cell
+4. The NULL value is replaced by the new value
+
+#### Scenario 2b: The data type is text, and the value is empty
+
+#### Steps for 2b
+
+1. The user edits a NULL value in a column that is set to text data type
+2. The user double-clicks the value cell and enters edit mode
+3. The user doesn't enter a value and leaves edit mode by hitting enter or clicking outside of the active cell
```
kgodey, line 77: I know we talked about this already in the discussion, but now that I see it in action in the video, I'm not sure about it. Sometimes, when I'm working with spreadsheets or tables, I just select a cell to highlight it for myself to help me think. In this case, if a user highlights the cell and stops highlighting it, then the value of that cell in the database will change, which I don't think users intend to do.

Instead, I think that we should only set the cell to an empty string if the user types something and then deletes it, not just if they enter edit mode and leave it. If they enter edit mode and leave it, it should stay NULL.

kgodey (changes requested):
@ghislaineguerin Please also add a link to this spec from the main specs page.

design/specs/specs-null-values.md
mathemancer: Is there a differentiation between "selecting" a cell, and "edit mode" on a cell?  For me, that's the most natural, and then this would be less of an issue.  @ghislaineguerin mentions "double-clicking" to enter edit mode, which is similar to how spreadsheets work (at least to my recollection).  If that's the case, it might make sense to somehow make the fact that you've entered "edit mode" very obvious and intentional.  For example, with LibreOffice, the cell highlight changes color and there's also an edit box in another place that activates when you're in "edit mode", but when you're just cursoring around, those things don't happen.  Just an idea.

I agree that we should try to make it an intentional act to set the cell to an empty string, though.  It's more normal to leave things as null than empty strings for most DB schemata.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,124 @@
...
+### Scenario 3: User sets a value to NULL
+
+#### Scenario 3a: The column accepts NULL values
+
+#### Steps for 3a
+
+1. The user selects a cell in a column that contains a value
+2. The user clears the content of the selected cell
+    - By pressing the delete key
+    - By opening the contextual menu and selecting the 'Set as NULL' option
```
mathemancer, line 107: I'm assuming this is in "normal/navigation mode" (i.e., not "edit mode").  Is that correct?  Is there a clear way for the user to see visually which mode they're in?  Many spreadsheet programs make a distinction between these two modes for cells, and that gives different behaviors for, e.g., the delete key depending on mode.  I'd be a bit grumpy (as a user) if I hit the delete key _in edit mode_, and the whole cell blanked out (as opposed to just the character to the right of the cursor).

On that front, if they're in "normal/navigation mode", would the backspace key have the similar "set to null" action as the delete key?  It seems like that would make sense if the two modes are clearly differentiated.

mathemancer (commented):
Overall, I think this is pretty good (especially the differentiation for viewing).  I do have a couple of questions w.r.t. modal editing, and whether the hints in that direction in this spec are intentional.  If they are, I think that should be made obvious (both in the spec, and also to the user ; maybe through some visual means).

design/specs/specs-null-values.md
seancolsen: > we should only set the cell to an empty string if the user types something and then deletes it, not just if they enter edit mode and leave it. If they enter edit mode and leave it, it should stay NULL

I agree. 

I think it will be rare for the user to intentionally change a cell value from NULL to empty string. However it will be more common for users to begin editing a cell (perhaps on accident) then stop editing that cell. We should avoid assigning a rarely desired outcome to a more commonly performed action.

**Further**, I think that when the user enters edit mode for a cell with a null value, the cell should _look different_ that it would with an empty string, even in edit mode. I think we should still display "NULL" (in italic) and place the user's cursor at the start of the text. Then when the user begins to type, the "NULL" text will disappear.

Converting a NULL value to an empty string would look like this:

1. Observe NULL in gray italic.
2. Double click.
3. Cell indicates the user is in edit mode by displaying thick border.
4. Cell indicates its value is NULL by displaying NULL in gray italic.
5. Cell places the user's cursor furthest to the left, before the NULL indicator text.
6. User enters any character, e.g. "a".
7. Cell displays "a". The NULL text disappears.
8. User presses `Backspace`.
9. Cell displays as empty -- no characters, no NULL indicator.
10. User presses `Enter` or blurs focus.
11. Cell exits edit mode and the value is saved as an empty string.

seancolsen (commented):
Very nice how you broke down these scenarios.

I added a couple minor comments, but generally this looks great!

design/specs/specs-null-values.md
pavish: I think everyone have covered this concern very well. I am on agreement with @seancolsen's suggestion to show null even during edit mode when user has not begun to edit yet.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,124 @@
...
+2. The user double-clicks the value cell and enters edit mode
+3. The user doesn't enter a value and leaves edit mode by hitting enter or clicking outside of the active cell
+4. The NULL value is replaced by an empty string
+
+#### Scenario 2c: The data type is boolean with checkbox display enabled
+
+#### Steps for 2c
+
+1. The user edits a NULL value in a column that is set to boolean type with the checkbox display property enabled
+    - The field contains a checkbox in an indeterminate state
```
pavish, line 85: It is not very clear to me what the indeterminate state represents looking at the UX.

* I think it would be better to show NULL next to the checkbox. Or,
* It might be better to show just NULL similar to other types, and only show the checkbox during focusing on the cell.

seancolsen: I had the same thought initially. However I came to accept the idea of the indeterminate checkbox for the following reasons:

- Showing some kind of checkbox is nice because it allows the user to change the value from NULL to TRUE.
- Showing "NULL" next to an indeterminate checkbox would increase the column width. Perhaps that's okay though.

For those reasons, I chose to suggest a hover tooltip.

I'm still somewhat on the fence about this.

design/specs/specs-null-values.md
ghislaineguerin: @kgodey It was mostly for consistency if I had other points that were not steps. So for some, I'd have steps and not for others.

design/specs/specs-null-values.md
ghislaineguerin: @mathemancer you are right, this is for navigation mode. In edit mode keys will have the default behavior

seancolsen (approved):

kgodey:
Assigned everyone except @seancolsen (since he already approved it) for re-review.

design/specs/specs-null-values.md
```diff
@@ -74,47 +74,54 @@ As for the process of transforming values to and from NULL, it will be important
 
 1. The user edits a NULL value in a column that is set to text data type
 2. The user double-clicks the value cell and enters edit mode
-3. The user doesn't enter a value and leaves edit mode by hitting enter or clicking outside of the active cell
-4. The NULL value is replaced by an empty string
+3. The user sees a text input with a placeholder containing the word 'NULL.'
+4. The user doesn't enter a value, and the value replaces the placeholder text
```
mathemancer, line 78: I don't understand what this is getting at.

mathemancer (commented):
There's one spot that I didn't quite understand the prose in the writeup.  However, the  associated video (for scenario 2) made sense to me.

design/specs/specs-null-values.md
ghislaineguerin: I have updated the text to clarify. It is mostly the behavior in which the cell turns blank rather than NULL if content was added and then cleared.

design/specs/specs-null-values.md
kgodey: I agree with @seancolsen here. I think showing `NULL` on hover will quickly teach users what an indeterminate checkbox is, even if the user doesn't get it at the first glance.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Scenario 2c: The data type is boolean with checkbox display enabled
+
+##### Steps for 2c
+
+1. The user edits a NULL value in a column that is set to boolean type with the checkbox display property enabled
+    - The field contains a checkbox in an indeterminate state
+    - Hovering over the NULL checkbox will display a tooltip with the 'NULL' legend.
+2. The user clicks on the checkbox, going from indeterminate (NULL) to checked (TRUE).
+3. The user can click on the checkbox again to mark it as false and go back to NULL.
```
kgodey, line 91: I'm not sure that we should support going back to `NULL` just by clicking on it. The general pattern with checkboxes is that clicking toggles TRUE/FALSE and I don't think we should violate that. I think we should only allow them to set this to NULL using the contextual menu.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Scenario 3: User sets a value to NULL
+
+#### Scenario 3a: The column accepts NULL values
+
+##### Steps for 3a
+
+1. The user selects a cell in a column that contains a value.
+2. The user clears the content of the selected cell.
+    - By pressing the delete key.
```
kgodey, line 110: I'm a little confused about how this will work with our current "spreadsheet like" editing for tables. Right now, the user only needs to click on a cell to edit it. This seems like it's introducing a pattern where the user needs to double click on a cell to edit it. I think this would hurt rapid data entry. I'd rather remove the feature where pressing `Delete`/`Backspace` would clear the cell and keep the single-click-to-edit pattern.

kgodey (changes requested):
@ghislaineguerin Looks good overall, added a couple of comments.

design/specs/specs-null-values.md
silentninja: > Right now, the user only needs to click on a cell to edit it
With the current master branch frontend interface, I have to double click to enter edit mode, clicking it once enters selection mode. Is this something part of the [Spreadsheet like editing milestone](https://github.com/centerofci/mathesar/milestone/100)

 Though we use a `column based` database and not a `row based` spreadsheet, we would still be needing a selection mode to reference, copy or drag and drop it as a reference. For any operation on a row rather than a row value, we would need a selection mode.

kgodey: I agree, I'm not arguing against the selection indicator. I don't see why clicking on a cell once can't do both actions - it can select the cell and also process any keystrokes while selected as editing the contents of the cell.

design/specs/specs-null-values.md
silentninja: Agree with kgodey, `null` should be seen as a state of the row and not as a value of a row., so an interface to change the value of a row shouldn't also change the state of the row.

design/specs/specs-null-values.md
silentninja: Makes sense, I would love to have something like this tbh, but I think the input element will mess up with the functionalities of the selection mode. For example, drag and drop would end up moving the text rather than the element itself.

kgodey: I'm not sure if that's necessarily the case. When I've implemented drag and drop in frontend, I think I had the flexibility to bubble the drag target up to the parent element. @seancolsen or @pavish would know better though.

seancolsen: However we spec this, I'm confident we'll be able to implement it. 

@silentninja is correct that the master branch currently requires double clicking before I can edit a cell. But I agree with @kgodey that we should change that behavior to be more consistent with spreadsheets. On a spreadsheet I can begin entering text even when the cell is not in edit mode. As soon as the first keystroke begins the cell switches to edit mode and begins replacing the contents with the key I just entered. That should be easy. I'm not seeing the relevant of drag and drop for this workflow, but I'm sure there will be edge cases to consider as we implement this. The good thing is that "spreadsheet like" editing is already fairly well-defined by other products so we just need to follow that UX. The fact that Google has come so far with their Sheets product on the web is part of what gives me confidence that we'll be able to implement our specs.

silentninja: I totally missed the fact that the cell turns into edit mode on keypress, I was thinking of not having an edit state visual cue at all. If that's the case then I think we don't have to remove the cell deletion upon the `Delete` keypress feature.

So an example scenario would be:
1. User highlights the cell with a single click, pressing the `delete` or `backspace` key will delete the cell state and convert it to a null value. 
2. Then the user presses any key other than `delete` or `backspace` the cell enters into an edit mode.  Pressing `delete` or `backspace` will delete the text inside the cell.

I think the same can be applied for boolean types too. The `checkbox` should be smaller than the cell
1. Clicking outside the `checkbox` would select the cell and pressing `delete` would make the cell null
2. Clicking inside the `checkbox` would toggle the value of the checkbox, similar to spreadsheets.

The only issue is with `dropdown's` though, we could have a null value as a dropdown option.

ghislaineguerin: What @mathemancer described here would be the correct behavior.
> As soon as the first keystroke begins the cell switches to edit mode and begins replacing the contents with the key I just entered. 

I think @silentninja is describing the same. I will update the spect to reflect this.

silentninja:
@ghislaineguerin Is it possible to document the null behaviour for dropdown cells?

kgodey:
@ghislaineguerin Please re-assign me when the spec is ready for re-review.

design/specs/specs-null-values.md
Joi: Not sure if this is too late in this conversation or appropriate, but I wonder if you could allow the user to enter edit mode and just hit `backspace` rather than have to create "fake text" and delete it in order to get an empty string.

design/specs/specs-null-values.md
Joi: Sorry about the misplaced comment earlier. Didn't read down all the way. So... regarding: 
> User highlights the cell with a single click, pressing the `delete` or `backspace` key will delete the cell state and convert it to a null value.
> Then the user presses any key other than `delete` or `backspace` the cell enters into an edit mode. Pressing `delete` or `backspace` will delete the text inside the cell.

Maybe too late and this is probably the right way to do it, but just intuitively, the idea of having to type something and then delete it to get an empty cell just feels "hacky" in sort of a bad way. Maybe there is a setting, but I'd rather, for example, do `⌘delete` or `⌃delete` for NULL and normal `delete` for empty string, for example...

silentninja: @Joi Upon single click, the cell gets selected, sort of like command mode in text editors. 

When any key excluding cell shortcut keys like delete, backspace is pressed, the cell goes into `edit mode`, which changes the appearance of cell from disabled grey to a text box, so keystrokes work like a shortcut for `enter key`. 

`delete` key will delete the words when in `edit mode`,  when in `command` mode it will delete the cell state.

Joi: @silentninja - ah, I see. So if I hit `delete` in `edit mode`, I can end up with an empty string without having to type anything other than `delete`? That makes sense. I misunderstood the interaction.

seancolsen: @Joi 

> So if I hit `delete` in `edit mode`, I can end up with an empty string without having to type anything other than `delete`?

If I'm understanding you correctly, I don't think this spec is headed (or should head) in that direction.

**When you're in edit mode**, you're editing text. You have a cursor. You can type new words. You can use the arrow keys to move your cursor. You can delete what you typed with keys like `Delete` and `Backspace`. Keys like `Delete` and `Backspace` (as well as key combinations with Command, Control, Shift, Option, Alt) have well established behavior (albeit with subtle differences across platforms) for editing text, and I don't think we should mess with that behavior. If you want to change the value of a cell from "Foo" to "" (empty string), then you'll need to double click to edit the cell. At this point your cursor will be placed _somewhere_ in the cell. We haven't defined that position yet. (For example LibreOffice places your cursor exactly where you clicked while Google Sheets places it at the end of the text.) In any case, at this point you'll need to manually _delete all the text_. There are may ways to do that. For example on Mac OS, `command+delete` will delete all the text between the cursor and the start of the line. That would delete all the text, but only if the cursor is at the end. Or `command+a` and then `delete` would select all and delete. Finally you'll need to press `Enter` to save. 

**When you're in select mode** you can use the arrow keys to move your selection between cells. You can press `Delete` or `Backspace` to set the cell value to NULL. There is no way to set the cell value to an empty string without entering edit mode and manually deleting all the text.

In general we're trying to discourage the use of empty string and encourage the use of NULL. 

Does my description make sense? And do you have any concerns with the direction we're heading? If so, it might be helpful for us to better understand the use case for empty string instead of NULL.

seancolsen: Further clarification...

When editing text:

- On Mac OS, pressing `Delete` deletes one character to the left. Pressing `Fn-Delete` deletes one character to the right.
- On Windows and Linux, pressing `Backspace` deletes one character to the left. Pressing `Delete` deletes one character to the right.

So when @silentninja said _"`delete` key will delete the words when in `edit mode`"_, I interpret that to mean _"the `delete` key will have its normal behavior (described above) when in `edit mode`"_.

silentninja: @Joi Sorry for the confusion I meant to say that delete will have its normal behaviour in `edit mode` which is to delete the letter not the word @seancolsen explains it much better.

pavish: @seancolsen's definition gives a clear idea of change in behaviour between select mode and edit mode.

There is a scenario we need to take into account regarding rapid data entry (@kgodey had a similar concern):

When the user is already in `edit` mode on a particular cell and then enters 'Tab', the active cell shifts and the immediate next cell enters into `edit` mode directly. This behaviour is similar to spreadsheets and it is what we have currently. When users are entering data rapidly, this would be their normal workflow.

In this scenario, users would never have to enter the `select` mode. So, setting NULL on a cell would be tiring. They would have to esc into `select` mode and set NULL and re-enter `edit` mode on the next cell.

@ghislaineguerin We probably need a keyboard combination to set NULL while the cell is in `edit` mode. Other spreadsheet applications do not have this problem (since they don't usually have the concept of NULL), so this would be something we can research and decide for ourselves.

pavish (approved):
@ghislaineguerin I think this looks good but keyboard interactions are not detailed anywhere, which would usually be the primary way of doing things.

@kgodey I do not think we should be spending more time on this at the moment, the spec looks good for mouse based interactions.

Perfecting this requires having to play around with the UI for all our usecases (including checkboxes, dropdowns, calendars and other input types), and we do not have visibility until we first figure out how data entry works on each of these types (using both keyboard and mouse interactions).

Setting cells to NULL for each individual scenario should be decided after we decide how to enter data on each of these cell types, which should probably be after our data types milestone is complete. For the time being I think the spec is good for mouse interactions and we can work on that alone.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,170 @@
...
+3. The user hovers over the indeterminate checkbox, and a tooltip with the legend `NULL` is displayed.
+
+### Scenario 2: User edits a `NULL` Value
+
+#### Scenario 2a: The data type is text
+
+##### Steps for 2a
+
+1. The user edits a `NULL` value in a column set to text data type.
+2. The user double-clicks the value cell to enter edit mode.
```
kgodey, line 75: I'm confused about this - I thought we had consensus in the comment thread that we should only require a single click to enter edit mode?

See these three comments: 
- https://github.com/centerofci/mathesar-wiki/pull/13#discussion_r765376340
- https://github.com/centerofci/mathesar-wiki/pull/13#discussion_r765211660
- https://github.com/centerofci/mathesar-wiki/pull/13#discussion_r770396298

kgodey (changes requested):
I'm confused about edit mode vs. select mode. I don't think we should require multiple clicks to edit a cell because it will be harder to do rapid data entry.

kgodey:
I'm fine with deferring keyboard interactions as @pavish suggested, @pavish or @ghislaineguerin could one of you create an issue so that we don't forget about it?

design/specs/specs-null-values.md
seancolsen: @pavish

## Product behavior

- I'm not sure we need a keyboard shortcut to set the cell value to NULL while in edit mode.
- In a spreadsheet, when you press `Tab` your selection moves to the right. But after that, you end up in select mode not edit mode. So if you want to set several adjacent cells in one column to NULL, you'd be able to do that by pressing `Delete`, `Tab`, `Delete`, `Tab` and so on.
- I can see how in some cases it would be useful to have a keyboard shortcut for setting a cell value to NULL while in edit mode. But it also introduces other UX questions that I don't think are worth spending the time to answer at this point. What happens _after_ I execute that keyboard shortcut? Am I still in edit mode? What if I mistakenly execute that keyboard shortcut? Will I be able to _undo_? It might be harder to support undo if we move to select mode. I suggest that we defer the _"set to NULL in edit mode"_ keyboard shortcut feature until a time when it becomes a more pressing need.

## Nomenclature

As an aside...

- I think in this conversation and spec (and others like it) it might be helpful if we make an effort refer to keys by the `key` value of the DOM `KeyboardEvent` (instead of by the text printed on the physical keyboard).
- On Mac OS, pressing the key labeled "delete" will send a `Backspace` event (and delete one character to the left). Pressing the keys labeled `Fn+delete` will send a `Delete` event and delete one character to the right.
- On Windows and Linux the physical key labels match their DOM event `key` values.
- That means that when we say _"press `delete`"_ in this discussion, it's somewhat ambiguous.
- If you execute `window.addEventListener('keydown', e => console.log(e.key))` in your browser console then you can press keys and see their respective DOM `key` values.

design/specs/specs-null-values.md
silentninja: @kgodey When you click on a cell once, the cell gets selected (ie) goes into the select mode, it would be a text box awaiting a text input rather than a greyed-out box without a cursor. But for convenience, when the user starts typing(any keystroke other than delete), the cell will change the mode, so it enters into edit mode with the cell turning into a text box, similar to what would have happened if you had double-clicked the cell

kgodey: @silentninja That makes sense to me, but then this part (later in the spec) becomes confusing to me:

> While in `Select Mode`, pressing the `delete` or `backspace` key will delete the cell state and convert it to a `NULL` value.

I would expect that if typing converts it to edit mode, delete/backspace will remove a single character, not convert it to `NULL`. It seems inconsistent to have special behavior for one key (that's also used in regular typing).

kgodey: @ghislaineguerin I think what @silentninja explained should also be explained better in the spec.

seancolsen: @kgodey 

I agree with @silentninja except for _"it would be a text box awaiting a text input rather than a greyed-out box without a cursor"_ (I find that part confusing). 

Here's my take on how a spreadsheet works and also how Mathesar should work:

| current mode | pointer action | pointer target | result |
| -- | -- | -- | -- |
| select | single click | current cell | nothing |
| select | single click | different cell | move selection to new cell |
| select | double click | current cell | switch to edit mode |
| select | double click | different cell | move selection to new cell, switch to edit mode |
| select | triple click | current cell | switch to edit mode, select all |
| select | triple click | different cell | move selection to new cell, switch to edit mode, select all |
| edit | single click | current cell | set cursor position |
| edit | single click | different cell | save changes to current cell, enter select mode, move selection to new cell |
| edit | double click | current cell | select work under cursor |

Now let's add in the `Delete` and `Backspace` behavior

| current mode | DOM `key` | result |
| -- | -- | -- |
| select | any text character | switch to edit mode, set cell contents equal to character entered |
| select | `Backspace` | set cell value to NULL |
| select | `Delete` | set cell value to NULL |
| select | `Shift+Backspace` | nothing |
| select | `Shift+Delete` | nothing |
| edit | `Backspace` | delete character to the left of cursor |
| edit | `Delete` | delete character to the right of cursor |
| edit | `Shift+Backspace` | _discussion pending_ |
| edit | `Shift+Delete` | _discussion pending_ |

kgodey: @seancolsen Thanks for the clear explanation. I find this behavior inconsistent - what if I want to edit the cell and only remove one character? It's confusing that typing `a` in a cell in select mode (for example) adds `a` to the contents of the cell, but typing `Delete` removes the entire contents.

I'd prefer to change the behavior to (only changed table rows are listed):

| current mode | DOM `key` | result |
| -- | -- | -- |
| select | `Backspace` | delete single character |
| select | `Delete` | delete single character |
| edit | `Shift+Backspace` | set cell value to NULL |
| edit | `Shift+Delete` | set cell value to NULL |

design/specs/specs-null-values.md
ghislaineguerin: @seancolsen that's a great suggestion on nomenclature, I've been struggling with maintaining consistency with that. I'll default to DOM `key` values from now on.

design/specs/specs-null-values.md
ghislaineguerin: @kgodey from my understanding if you are in select mode and type `a` you'd be replacing the cell content with the value `a`, not adding to the existing one.

seancolsen: @kgodey 

- One clarification:

    > typing `a` in a cell in select mode (for example) adds `a` to the contents of the cell

    No, typing a in `a` cell in select mode _replaces the entire cell contents_ with "a" (and also switches to edit mode). 

- Regarding your suggestion above... that's a bit different from the way that spreadsheets behave. In a spreadsheet, if I'm in select mode and I press `Delete` or `Backspace`, the cell value is set to an empty string.

kgodey: @ghislaineguerin I don't think we should do that either, I think we should err on the side of not entirely replacing data in the database based on quick actions. We care about data integrity. If the user accidentally types `a`, then they can remove the `a` if they make the mistake. If we end up replacing the content of the cell, then they can't recover their previous data, which sounds scary, especially since we don't have undo yet.

kgodey: @seancolsen I didn't see your comment before I responded, but what I said still applies. I don't think we should emulate a spreadsheet in that respect before we have undo functionality.

seancolsen: @kgodey good point about the risks given our lack of undo at this stage.

If we don't want `Backspace` and `Delete` to set the cell value to NULL in select mode, then I think we should leave the behavior void at this point so that we can add it in later (once we implement undo) without confusing users who have already been using Mathesar. 

So in that case it would be:

| current mode | DOM `key` | result |
| -- | -- | -- |
| select | `Backspace` | nothing |
| select | `Delete` | nothing |
| select | `Shift+Backspace` | set cell value to NULL |
| select | `Shift+Delete` | set cell value to NULL |

silentninja: I proofread it multiple times and still messed up, edited my comment with the correct answer

silentninja: > No, typing a in a cell in select mode replaces the entire cell contents with "a" (and also switches to edit mode).

I don't think we should replace the cell value, rather the entered character should be appended to the existing value.

seancolsen: > the entered character should be appended to the existing value

I don't think there are any spreadsheets that behave that way though.

silentninja: The behaviour I mentioned was the one I was expecting, but Spreadsheets have nailed the UX when it comes to editing, so wouldn't want to ditch that decision and I think it does make sense in a few cases. Do we have a undo shortcut and can it bring the data back?

kgodey: @seancolsen @silentninja I don't think we should perfectly emulate spreadsheets given that we do not have undo yet.

I think that `Backspace` or `Delete` in `select` mode should be similar to entering a text character:

| -- | -- | -- |
| select | any text character | switch to edit mode, append character to text content |
| select | `Backspace` | switch to edit mode, delete single character |
| select | `Delete` |  switch to edit mode, delete single character |

so that wou

kgodey: @ghislaineguerin I find this table view helpful, can we include it in the spec to explain select vs. edit mode?

seancolsen: @kgodey

Once we've implemented "undo", would you still want to keep the behavior you're proposing? Or would you want switch to the behavior I'm proposing at that point?

kgodey: @seancolsen I'm not sure, we can decide at that point. It should be after alpha so we'll have user feedback to help make the decision.

seancolsen: @kgodey I was asking because your answer would affect my opinion here.

@kgodey Given the consistency in UX across different spreadsheet products, I'm a quite wary of intentionally diverging from those well-established patterns. I understand and agree that we need to implement undo before we can safely implement all of the spreadsheet UX patterns. But I don't want to let the sequence of tasks in our roadmap influence our UX design. If we can't do it "the right way" right now, then I'd rather hold off until we can instead of implementing a UX pattern that will be unfamiliar to users. I can appreciate that data entry will be more cumbersome if I need to switch to edit mode (via double click or `F2`) before editing a cell. But I think that shortcoming is acceptable while we're still in alpha. For alpha I think it's more important to communicate _the direction the product is heading_ than to supply a product that people can rely on for real-world purposes. If we stray from the spreadsheet UX I think our alpha users will receive some subtle communication from the product that direction is not heading exactly towards spreadsheet-like editing, and that will lower peoples' expectations of our stabilized UX.

---

That said, I don't want to harp on this too much. It's a pretty minor detail after all.

If we _do_ go with your proposal, then I'd like to just make the following change 


| current mode | DOM `key` | result |
| -- | -- | -- |
| select | `Delete` | nothing |

Why? Because `Delete` is _forwards_ delete. If we delete the last character in the cell, that would be as if we had placed the cursor one character before the final character, which is weird and arbitrary. If we delete the _first_ character in the cell that seems weird too and inconsistent with any products I can recall using.

We could potentially have `Delete` trigger a "switch to edit mode" action. But for simplicity's sake I'd suggest we leave it unassigned to any action.

kgodey: @seancolsen I see your point but I am unconvinced that the spreadsheet pattern is the "right" way of doing things or that we will end up doing things that way in the end. I think it might be useful for users to understand that we _aren't_ quite a spreadsheet and that we prioritize data integrity. I'd rather take a risk on this new pattern in our alpha release to see if it works well for users. If it doesn't, then we can always switch to the well-established spreadsheet pattern.

So I'd like to go with my proposal. I agree with your amendment, My suggestion was based on `Delete` being the same key as `Backspace` (because that's what it's called on Macs), I agree that forwards `Delete` should do nothing.

seancolsen: @kgodey Sounds good :grin:

seancolsen: To summarize where we're at now:

| current mode | DOM `key` | result |
| -- | -- | -- |
| select | any text character | switch to edit mode, cursor at end, append character to end of cell contents |
| select | `Backspace` | switch to edit mode, cursor at end, delete last character |
| select | `Delete` | nothing |
| select | `Shift+Backspace` | set cell value to NULL |
| select | `Shift+Delete` | set cell value to NULL |
| edit | `Backspace` | delete character to the left of cursor |
| edit | `Delete` | delete character to the right of cursor |
| edit | `Shift+Backspace` | _discussion pending_ |
| edit | `Shift+Delete` | _discussion pending_ |

design/specs/specs-null-values.md
mathemancer: The process @seancolsen described matches what I'd expect personally.  I also agree that it's not that necessary to have a keyboard shortcut while in edit mode to set `NULL`.

design/specs/specs-null-values.md
mathemancer: I'm a little late, but I'm really concerned about a detail here that was assumed, but glossed over:  In spreadsheets, if I enter edit mode, then press `Esc` (without navigating by clicking, tabbing, or returning), the cell returns to its original state.  It seems like we are not planning to have that.  Shouldn't that be the way it works?  We don't actually need "undo" for that, and it's very expected by common users of spreadsheets.  Kind of a freebie "mini-undo".  All we'd need to do is avoid sending any requests until the appropriate navigation has taken place (which we should do anyway).

Given that, I don't see the danger of following the expected "type from select replaces cell value with new character".  All they need to do to "undo" is escape, and users are pretty familiar with that (I tried both Google Sheets and LibreOffice; they each use that setup).

For "delete from select", It's much more efficient to have a LibreOffice sheets style dialog pop up "delete contents?" than to have to delete each character.  This would allow quickly deleting cells without spamming the delete key, but still avoid destroying data without confirmation.  If we want to use a confirmation when "delete" (or backspace) is pressed, it might then make sense to also have a keyboard shortcut for users who know what they're doing to "nullify" a cell without confirmation from select (perhaps shared with edit mode).

kgodey: @mathemancer Please put your comment in https://github.com/centerofci/mathesar/issues/902, we'll fully define keyboard actions for editing then. I think we should go ahead with the proposal here to unblock this spec.

seancolsen: @mathemancer the `Esc` has been on my mind too throughout this conversation and I almost brought it up. I would definitely like to implement the `Esc` behavior as part of https://github.com/centerofci/mathesar/issues/902. It certainly mitigates the risk of data loss, but in my opinion it doesn't provide enough mitigation to make me feel comfortable implementing the "type from select replaces cell value with new character" UX pattern in spreadsheets. Why? Because I think we'll also want to implement "save cell contents on blur", meaning if the user clicks away from the cell while in edit mode, we'll save their edits. That makes it pretty easy to mess up.

@kgodey this whole conversation has me elevating my mental prioritization of "undo" quite significantly as it seems to be a crucial factor in implementing the UX patterns which which users are most familiar.

kgodey: Undo has always been high on my mental prioritization for that reason - I don't think users will be able to fully explore and discover features without being able to fix mistakes. But it's also complicated to implement so I think we made the right decision to push it to post-alpha.

seancolsen: @kgodey I had a couple more thoughts on this thread last night.

- @ghislaineguerin I'm curious to hear your opinion here. To summarize, we've preliminarily agreed to follow Kriti UX which deviates from standard spreadsheet UX as follows

    | Current mode | DOM `key` | Spreadsheet UX | Kriti's UX |
    | -- | -- | -- | -- |
    | select | any text character | switch to edit mode, **replace** cell contents with character | switch to edit mode, place cursor at end, **append** character to end of cell contents |
    | select | `Backspace` | **delete cell contents** | switch to edit mode, place cursor at end, **delete last character** |

    I'd like us to adhere to the spreadsheet UX, but wait until we have implemented undo before we implement those actions. However, for the sake of moving forward, I've agreed to Kriti's proposal.

    I just want to make sure we're not losing your voice here, Ghislaine since you're the UX person!

- Kriti said:

    > I'd rather take a risk on this new pattern in our alpha release to see if it works well for users. If it doesn't, then we can always switch to the well-established spreadsheet pattern.

    This had me thinking. If we take a risk on a UX pattern that is unfamiliar to users, I think we're going to need to be very aware of this divergence and intentionally seek feedback on it during our user testing. I doubt we'll have the resources to dedicate towards the type of user testing the Microsoft has done with Excel or that Google has done with Sheets. I'm sure those companies have run paid focus groups to observe new users interacting with the product. In my experience building UIs, it's absolutely essential to watch users with the product when seeking feedback on theses subtle UX decisions, but it takes time, especially with a large product. New users will basically never express opinions with the specificity necessary for actionable change on small UX decisions. But _watching_ them use the product will inevitably uncover problems. If we diverge from well established UX patterns that type of focus group testing becomes even more important so we'll need to keep this in mind and consider allocating some budget towards it later this year.

kgodey: >  In my experience building UIs, it's absolutely essential to watch users with the product when seeking feedback on theses subtle UX decisions, but it takes time, especially with a large product. New users will basically never express opinions with the specificity necessary for actionable change on small UX decisions.

It's always been my plan to schedule sessions with users where we watch them use the product after release. I have some experience with running user tests, I read _[Rocket Surgery Made Easy](https://www.amazon.com/Rocket-Surgery-Made-Easy-Yourself/dp/0321657292)_ at my last job, which is a great guide to usability testing and I highly recommend it.

ghislaineguerin:
@kgodey I've updated the spec. I will create the additional issue for select and edit mode.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,176 @@
...
+3. The user hovers over the indeterminate checkbox, and a tooltip with the legend `NULL` is displayed.
+
+### Scenario 2: User edits a `NULL` Value
+
+#### Scenario 2a: Replacing `NULL` with a new value
+
+##### Steps for 2a
+
+1. The user edits a `NULL` value in a column.
+2. The enters `edit mode`.
```
kgodey, line 76:
```suggestion
2. The user enters `edit mode`.
```

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Scenario 3: User sets a value to `NULL`
+
+#### Scenario 3a: The column accepts `NULL` values
+
+##### Steps for 3a
+
+1. The user selects a cell in a column that contains a value.
+2. The user clears the content of the selected cell.
+    - By pressing the `delete` key.
```
kgodey, line 119: We decided to make this shift-delete instead.

kgodey: or shift-backspace.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,176 @@
...
+    - By opening the contextual menu and selecting the `Set as NULL` option.
+3. The selected cell now has a `NULL` value.
+
+#### Scenario 3b: The column does not accept `NULL` values
+
+##### Steps for 3b
+
+1. The user selects a cell in a column that contains a value.
+2. The user attempts to clear the content of the selected cell.
+    - By pressing the `delete` key.
```
kgodey, line 129: We decided to make this shift-delete or shift-backspace instead.

design/specs/specs-null-values.md
```diff
@@ -0,0 +1,176 @@
...
+        - The system prevents the insertion of a `NULL` value.
+        - An error message indicates that a NOT `NULL` constraint is applied to the column, and the system cannot insert `NULL` values.
+    - By opening the contextual menu and selecting the `Set as NULL` option.
+        - The 'Set as `NULL`' option is disabled, and the user cannot select it.
+
+## Interactions
+
+### Select vs. Edit Mode Behaviors
+
+In `Edit Mode`, a cell provides additional functionality and interactions to users. `Edit Mode` is toggled by double-clicking on a cell. However, certain data types could have different interactions. Once specific behaviors and components for each data type are introduced, we must further define these interactions to ensure compatibility.
```
kgodey, line 139: We decided to also enable `Edit Mode` if the user types a text character or backspace while in `Select Mode`.

kgodey: Would be good to also define `Select Mode` since it's referenced in the table below.

kgodey (approved):
@ghislaineguerin Added some comments and suggestions for some minor missing items. Please feel free to merge after you fix them, no need for an additional review.

Also, I haven't reviewed the Loom videos to see if there are any changes needed. Could you please review them and make sure they follow the same behavior as described in the spec? We've changed some things based on discussions since the videos were recorded.

design/specs/specs-null-values.md
kgodey: @ghislaineguerin and I talked about this on a call, we'll deal with this in the follow up issue #917.

PR #16, opened 2021-12-20
Listing Views Spec

ghislaineguerin:
Specs to implement the design for [Support for listing database views](https://github.com/centerofci/mathesar/issues/455)


seancolsen (commented):
### Scenario 1 -- "Schema detail view" 

- What is the purpose of the "Schema detail view" page? Why do we need it?
- How does the user navigate to this page?
- What should the URL look like for this page?
- What does "Last access" mean? The time I viewed it? The time anyone viewed it? The time someone modified it?

ghislaineguerin:
@seancolsen thanks for the feedback. 
> What is the purpose of the "Schema detail view" page? Why do we need it?

At some point, as we've discussed in the issue for 'Listing Views' we'll have limits on how many views or tables we can show, and having an overview of the schema might be helpful for this.

> How does the user navigate to this page?

The user can navigate to it by clicking on the top nav's schema name.

> What should the URL look like for this page?

I'm assuming just `database/schema` without the selected object at the end. But now I'm wondering if the 'explorer' view allows a 'nothing selected' state and if that's a state that needs to be designed as well. I'll think about this and update.

> What does "Last access" mean? The time I viewed it? The time anyone viewed it? The time someone modified it?

Last accessed was added in replacement for 'last updated' as we have no way of knowing that. Accessed would mean the object was opened in the explorer view.

seancolsen:
@ghislaineguerin Here are some thoughts based on your above clarifications:

- I don't think we need the "Schema detail view" page. Can we wait to add that page until the need for it becomes stronger? Broadly speaking, I can see the usefulness of the "last accessed" field as a way to quickly find the objects I've been working with recently. But I think we could safely defer this feature until after Alpha. Would you agree?
- If we _do_ add that page, then:
    - You said:

        > The user can navigate to it by clicking on the top nav's schema name.

        Currently clicking on the schema name will select the schema. So with your proposed functionality, the user would have no way to toggle between viewing separate schemas without first clicking through the "Schema detail view", correct? That seems like a (very minor) UX regression. Would the user have some other way to select a schema without navigating to the "Schema detail view" page?

    - You said:
    
        > Accessed would mean the object was opened in the explorer view.

        I have more questions:
    
        - Is it the time that _the current user_ last opened the object in the explorer view? Or the time that _any user_ last opened the object in the explorer view?
        - @kgodey do we already have support for this "last accessed" functionality at the API level? I'm curious what events we're using (or would plan to use) at the API level to qualify as as "accessing" the object for this purpose. Maybe we would rely on a GET request to the `records` API? Or would the be a different mechanism? Perhaps a dedicated endpoint? I'm sure we'll have a more detailed conversation at implementation time, but for now I'm just trying to understand the meaning behind "last accessed", for the purpose of UX.

ghislaineguerin:
@seancolsen I've introduced the 'last accessed functionality based on [this comment.](https://github.com/centerofci/mathesar/issues/455#issuecomment-961755582) The requirements specified 'last updated' which we couldn't determine.

ghislaineguerin:
@seancolsen I agree we can wait to add the detail view once the need becomes stronger and there's more to support it, besides a limit for objects displayed in the sidebar.

seancolsen:
@ghislaineguerin thanks for the additional context. Using localStorage as Pavish suggested in your linked comment makes sense and answers my questions. If we're going ahead with the "Schema detail view" page, the it would be nice to add that info to the spec that the time will get set in the browser's localStorage whenever the user opens the object in explorer view.

kgodey:
> it would be nice to add that info to the spec that the time will get set in the browser's localStorage whenever the user opens the object in explorer view.

This is a technical implementation detail, I don't think it makes sense in the UI/UX design spec.

design/specs/listing-views.md
```diff
@@ -1,27 +1,109 @@
...
 ## Scenarios
 
-### User sees a list of all views and can filter or sort the list items
-A user can see a list of all views for a schema in the sidebar area. They can filter the list using the search input or sort them by alphabetical order or last modified. Last modified is the default sorting for the list of views.
+### Scenario 1: User browses through a list of all available views in a currently open schema
+
+#### Scenario 1a: From the schema details view
+
+- User navigates to the schema details page by navigating directly to a URL that points to this page or by clicking on the schema link at the top navigation bar.
+- User sees a list of all tables and views for the current schema.
```
kgodey, line 43: Views and Tables are going to have pretty different functionality so I think we should separate them into two different sections on the schema detail page (rather than showing them as the same list) to make the distinction clearer to the user.

I do think we can have a unified search for both, although maybe we'll still want to split search results into sections.

design/specs/listing-views.md
```diff
@@ -1,27 +1,109 @@
...
+  - By default, the list is sorted by the last accessed object.
+- Above a certain number of objects, the list will be divided into pages. The user will then have to navigate the pages to access the entire list.
+
+#### Scenario 1b: From the schema explorer view's sidebar
```
kgodey, line 47: I think that it would be better to have a dropdown to select between "All", "Tables", and "Views" rather than tabs - it will make the sidebar be narrower.

I'm also not sure about "All" being the default and showing everything in a single list. I think we need to make the distinction between tables and views more clear to the user.

design/specs/listing-views.md
kgodey: It might also be a good idea to show some metadata about the tables and views. e.g. we can show the number of records or rows, and maybe some details about the column names? It would help users identify what they're looking for if they don't have great names for their DB objects.

kgodey (changes requested):
@ghislaineguerin Looks good overall, added some comments.

kgodey:
I also want to second @seancolsen's question here:

>  Currently clicking on the schema name will select the schema. So with your proposed functionality, the user would have no way to toggle between viewing separate schemas without first clicking through the "Schema detail view", correct? That seems like a (very minor) UX regression. Would the user have some other way to select a schema without navigating to the "Schema detail view" page?

I think it would make more sense to access the schema detail view by clicking the schema name from the schema list page, which we already have implemented.

---

@seancolsen it looks like you have the answer about "last accessed", let me know if you have further questions for me.

design/specs/listing-views.md
```diff
@@ -1,27 +1,109 @@
...
 
-### User sees a list of all views and can filter or sort the list items
-A user can see a list of all views for a schema in the sidebar area. They can filter the list using the search input or sort them by alphabetical order or last modified. Last modified is the default sorting for the list of views.
+### Scenario 1: User browses through a list of all available views in a currently open schema
+
+#### Scenario 1a: From the schema details view
+
+- User navigates to the schema details page by navigating directly to a URL that points to this page or by clicking on the schema link at the top navigation bar.
+- User sees a list of all tables and views for the current schema.
+  - By default, the list is sorted by the last accessed object.
```
pavish, line 44: The intention behind my earlier suggestion of the 'recently accessed' section was if we display only a limited set of tables & views on the side bar. I personally don't find value in providing sort by last access if we are always going to be displaying the entire list.

On the implementation aspect, sorting would be a mess to implement if the backend does not provide the option. Displaying a few recently accessed items on the sidebar is a straight forward implementation but sorting the entire list based on data stored in localStorage would be a mess (each browser would show a different order), while it is still possible. I'm not sure if it's worth it to spend time implementing that.

design/specs/listing-views.md
pavish, line 47: I dont think we need the 'All' option. As mentioned in the previous comment, if we are going to be showing all the tables and views, then I see no major value in having a 'last accessed' sort.

If we show a limited set of the list on the sidebar (since we cannot practically use the sidebar with >100 tables & views), then we could have a 'Recently accessed' section with around 10 to 15 entries.

design/specs/listing-views.md
```diff
@@ -1,27 +1,109 @@
...
+- Above a certain number of objects, the list will be divided into pages. The user will then have to navigate the pages to access the entire list.
+
+#### Scenario 1b: From the schema explorer view's sidebar
+
+- User navigates to the schema explorer view by opening any object in the details page or navigating to an object URL.
+- User sees a list of all tables and views for the current schema.
+  - The list shows both object types by default but can be filtered to only show tables or views using the filter bar at the top of the sidebar.
+
+### Scenario 2: User encounters an error in loading a list of views for a currently open schema
```
pavish, line 53: We need UX to show the error on the sidebar as well.

There is a possibility that the entries fail to load on the sidebar while the schema detail page loads successfully.

pavish (changes requested):

pavish:
@ghislaineguerin 

I agree with both @seancolsen and @kgodey that we should not be moving to the schema details page when clicking on the top navigation menu.

It would be much better to do it from the 'Manage schemas' (schema listing) page as @kgodey  suggested. Optionally, we could provide an additional icon with each schema in the top navigation menu, which will directly lead to the schema detail page, if you think that provides value to the user.

design/specs/listing-views.md
```diff
@@ -1,27 +1,109 @@
...
+### Videos
+
+- Scenario 1: [Video](https://www.loom.com/share/179070149f00419e9125bd8fba675aa4)
+- Scenario 2: [Video](https://www.loom.com/share/e32500bc744246e5b926852231577fd7)
+- Scenario 3: [Video](https://www.loom.com/share/d553e7edbb28481b8aed02b055270154)
+- Scenario 4: [Video](https://www.loom.com/share/91a20c4770ef4439860425a1bcd84db6)
+
+## User Experience
+
+### Listing of Views and Searching by Name
```
pavish, line 30: @ghislaineguerin 
Tables and views are currently treated as if they are on the same level i.e equal objects to the user, however that's not the case.

The user should be able to identify that views are a higher layer over tables, and tables are more of a raw data store. This should be made evident in both the sidebar and the schema details page.

I'm not sure how we would be able to get this message across effectively, but I'm certain that representing them both on the same level would only get the wrong message across.

@kgodey Do we need to worry about this for the alpha or should we wait until after alpha? I assume it's going to be very tricky to figure out an UX for this.

kgodey: @pavish I had the same comment as you, I think it's important to solve it for the alpha since it's a fundamental concept for the product.

pavish: @kgodey I saw your comment on separating them out and I agree with it for the alpha, but I also think that we should also find a way to represent the difference in level to the user.

* If we keep them as separate tabs in the schema details page or on the sidebar, they appear like they are equal level objects which are just different from each other.
* We need to represent that they are not equal and views are a higher layer over tables and I think it would require a lot more thought and experimentation. This should be made clear to the user from the very beginning of using the product.

kgodey: I see what you mean. If @ghislaineguerin has a design concept that doesn't take too much work, we can figure it out for the alpha. Otherwise, I think it's fine to defer until afterwards.

design/specs/listing-views.md
ghislaineguerin: @kgodey I had them separated initially for the same reason, but as I was thinking about the workflow I imagined a user would be jumping from view to table, to view again especially if they are creating a view from an existing table and I liked the idea of having them as a unified view. Maybe this could be solved by tabs. I'll consider this and make changes.

kgodey: I think there's other ways to solve users being able to jump between tables and views easily. It's really important for users to be able to tell the difference between Views and Tables.

design/specs/listing-views.md
```diff
@@ -1,27 +1,124 @@
...
+Users of Mathesar will be able to connect multiple databases containing multiple schemas and navigate between them. This prototype assumes there's a single schema and database. For that reason, the interface presented is limited in functionality.
 
-### User sees a default view when they open a schema
-A view can be set as the default that a user will see when they open a schema. Users can set the default view from the 'View Options' menu in a view.
+### Regular vs. Materialized Views
 
-### User can see a list of views referencing a table
-A user from a table can access a list of views referencing that table. This menu also contains a brief description of how users should use views. 
\ No newline at end of file
+No differentiation has been introduced for materialized views, as we concluded that the functionality would be similar. A different icon was proposed, which can be added to the UI improvements design work.
```
mathemancer, line 124: I apologize for not catching this earlier.  Materialized views are more similar to tables than to non-materialized views.  In fact, they _are_ tables, just annotated with the query that created them as well as enhanced with some functionality to update or refresh them.

From a list of materialized views, I think that it should be possible to:
- see if a materialized view is being (non-concurrently) refreshed from the listing pane, since that would produce a lock on the view making it unavailable.
- trigger a refresh of all materialized views in some list by clicking something.
- long-term, we need to show the last time each materialized view was refreshed in the list, since this is an important piece of info when deciding whether to interact with (refresh) the view.

To reiterate, the distinction between materialized and non-materialized views is crucial, since the non-materialized ones are stateless (I discussed this in another comment) and the materialized ones are stateful.  This leads to quite different use cases, as different as between tables and stateless views.

mathemancer (commented):
I left some more specific comments, but the highlights are:
- I think @pavish is right that we need to distinguish between tables and views more
- I think that we _shouldn't_ present views as an abstraction layer over tables
- I think that we need to distinguish materialized and non-materialized views, since they have different use-cases and need different interactions when listing.

design/specs/listing-views.md
dmos62: > the non-materialized ones are stateless (I discussed this in another comment) and the materialized ones are stateful

I take it that a materialized view's state is only how out-of-date it is. You can't actually mutate the view without mutating what it references, or am I wrong?

kgodey: @mathemancer The product goals for the alpha release are to ensure that materialized views that are already present in a database are shown as such to the user. We will not have any other functionality for creating or working with materialized views from the UI.

I don't think it is worth spending much design effort into materialized views until we have better support for them in the product.

kgodey (commented):
@ghislaineguerin Looks good! Please assign me again when it's ready for re-review.

design/specs/listing-views.md
mathemancer: @dmos62 The state of the materialized view is the data.  It's "frozen in time", and maintains its own copy.  The "last updated" part is metadata.  As an example, if you create a materialized view, then change the table it was made from, the materialized view will _not_ change to show the update, until you refresh it.

@kgodey Regardless, we'll be missing functionality that we have planned for both views and tables (e.g., inserting, updating) for materialized views, which can't be altered _except_ by refreshing them.  Even if we don't want to include refresh functionality, I maintain they're as different from normal views as normal views are from tables, and should be separated.  It's unfortunate that they have the name "view", since they're not really the same.

kgodey: @mathemancer we had a conversation about updating and inserting into views in https://github.com/centerofci/mathesar/issues/834 and came to the conclusion that we would have pretty much identical functionality for both views and materialized views.

Having materialized views be a separate section has major product implications – we would then need to explain to non-technical users what this third type of object is and what it's meant to be used for. Since they are going to have pretty much identical functionality to views, I'd rather lump them into views with a different icon and an explanation that they're not up to date when opened (or something similarly basic). The other alternative I can see is to not support them at all or show them in the UI, but that seems worse when we'll already have most of the work done to support them.

design/specs/listing-views.md
```diff
@@ -1,27 +1,120 @@
...
-# User Experience
 ## Scenarios
 
-### User sees a list of all views and can filter or sort the list items
-A user can see a list of all views for a schema in the sidebar area. They can filter the list using the search input or sort them by alphabetical order or last modified. Last modified is the default sorting for the list of views.
+### Scenario 1: User browses through a list of all available views in a currently open schema
+
+#### Scenario 1a: From the schema details view
+
+- User opens a schema and sees the [schema details page](#schema-details-page).
```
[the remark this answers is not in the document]
ghislaineguerin: @kgodey I've added a description below. The link on that line should take you there. The details show when no tabs are open. It's like a "get started" kind of page.

design/specs/listing-views.md
```diff
@@ -1,27 +1,120 @@
...
+- User sees a list of views [recently opened](#recently-opened) in Mathesar for the current schema.
+
+#### Scenario 1b: From the schema explorer sidebar
+
+- User opens a schema and sees the [schema explorer sidebar](#schema-explorer-sidebar).
+- User sees a list of all tables and views for the current schema.
+  - The list shows both object types by default but can be filtered to only show tables or views using the [filter bar](#sidebar-filters) at the top of the sidebar.
```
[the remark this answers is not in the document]
ghislaineguerin: I think maybe the name filter is misleading. I'll update that to tabs instead. I'd be concerned about having multiple instances of filter and confusing users.

design/specs/listing-views.md
ghislaineguerin: I agree that it would not be helpful for a large collection of tables, especially if the accessed timestamp is not available for all. I have removed this from the prototype.

design/specs/listing-views.md
```diff
@@ -1,27 +1,120 @@
...
+- The list is filtered to show objects whose name matches the search term.
+- The list contains metadata associated with the objects like record and column count.
+  - If no results are found, the system displays a notice to indicate the lack of results and potential causes.
+- Users can clear the search by clearing the input box content or clicking on the `Clear Search` link displayed along with the results summary.
+
+## Views
+
+### Schema Details Page
+
+The schema details page will be shown when a schema is open and notables are selected. It will contain the schema's name, the total number of tables and views, and additional sections that we can use to present the user with relevant information or links. Examples of content that could go into this section are:
```
kgodey, line 96: Do you mean "no tables" instead of "notables"? I was trying to figure out what a "notable" was. :)

design/specs/listing-views.md
```diff
@@ -1,27 +1,120 @@
...
+
+- Recently opened views or tables
```
kgodey, line 98: Can you add a note here that these are examples and not meant to be part of the implementation of this spec.

kgodey: If any of these are not examples and should be implemented (maybe recently opened?), please indicate those.

design/specs/listing-views.md
```diff
@@ -1,27 +1,120 @@
...
+
+### Sidebar Tabs
+
+Sidebar tabs will allow users to filter the sidebar's contents to show only a particular object type. When set to ' All ', the content will include both tables and views. Setting the tabs to `Tables` will display only table objects, while `Views` will display only view objects.
+
+## Other Considerations
+
+### Regular vs. Materialized Views
+
+Regular views need to be differentiated from materialized views as the use cases might differ. However, we don't want to create a rigid distinction between both as it could confuse users being introduced to the concept of views. For this purpose, an indicator, such as the letter 'M' can be added next to the view icon to show that it is of the type 'Materialized'.
```
kgodey, line 116: What does "can be added" mean here? Should it be added or not?

kgodey (changes requested):
@ghislaineguerin Looks good, some minor changes requested.

design/specs/listing-views.md
ghislaineguerin: @kgodey I did add an indicator to the design. I believe it should be added, I will update the spec to remove the ambiguity.

design/specs/listing-views.md
ghislaineguerin: Yup, that was it, I'll fix it. Thanks

kgodey (approved):
Looks good @ghislaineguerin!

PR #19, opened 2022-01-06
Left Sidebar Improvements

ghislaineguerin:
This spec goes over the design and implementation notes for [Left Pane Improvements](https://github.com/centerofci/mathesar/issues/723)


seancolsen (commented):
This looks great. It seems like it's very similar to the VS Code sidebar which I like -- not because our users will necessarily be familiar with VS Code -- but because I think part of the reason VS Code has gained popularity so quickly is that its UX is very approachable.

## Questions

- In the explorer, when I enter a search query like "foo" with the filter below set to "Tables", will it also show _views_ that match "foo"? Or will the search respect the filter?

## Suggested formatting changes

_for this spec and for future ones too_

- I'd prefer to see just one heading per scenario. Here we have three headings per scenario which seems superfluous.
- I'd prefer the scenarios to be a flat list. Instead of `1a`, `1b`, I'd prefer to have `1`, `2`, and so on.

pavish (approved):
Looks good to me

ghislaineguerin:
@kgodey I didn't include any special UI for partial loading of content in the case of long lists and we decided not to address that now (when we worked on the design for listing views). We will have basic scrolling and if there are both tables and views, then the user should see both sections, meaning that a section doesn't get pushed down. That's implemented in the prototype.

silentninja:
I like the sidebar design, glad to see that it won't be used just for navigation. Will it be possible to see the information of the view such as the table and columns it is derived from in a split panel below the view listings? By split panel, I mean the place where tables are listed when viewing the `All` section

kgodey:
> @kgodey I didn't include any special UI for partial loading of content in the case of long lists and we decided not to address that now (when we worked on the design for listing views). 

@ghislaineguerin I thought we weren't handling it in listing views because we were going to handle it here, it's part of the issue description for https://github.com/centerofci/mathesar/issues/723. If you're not handling all the requirements from https://github.com/centerofci/mathesar/issues/723, please open a separate issue for whatever you're not handling and link it from the original issue. I would also put a note in the spec so that people who are comparing the spec and the issue are not confused about why it doesn't match up with all the requirements.

PR #22, opened 2022-01-13
Product spec for Views

kgodey:
I was thinking through how to represent Views in Mathesar and ended up having a lot of thoughts on the data model and how to translate DB queries to Views. I figured we could have the most granular discussion here instead of a GitHub Discussion.

## Previous discussion and context
- Meta issue for current Views issues: https://github.com/centerofci/mathesar/issues/442
- https://github.com/centerofci/mathesar/discussions/838

## Additional notes
I also fixed a bug with the organize images script, it couldn't handle images with spaces in the filename.


kgodey:
@dmos62 @ghislaineguerin @mathemancer @pavish @seancolsen @silentninja I've assigned all of you to review. As usual, please unassign yourself when you have left feedback and don't intend to leave any more or if everything looks good to you.

product/specs/2022-01-views/02-modeling-views.md
```diff
@@ -0,0 +1,84 @@
...
+
+Column attributes are defined [in a separate page](/product/specs/2022-01-views/03-modeling-view-columns.md).
+
+## Rows
+These are the output rows of the View's Query. Rows are **not required**, Views can contain 0 rows.
+
+## Filters
+Views can have filters applied. Unlike Tables, view filters are not necessarily related to the columns that are present in the view.
+
+Using the example table above, imagine a view created from the query `SELECT ID, Title FROM Movies WHERE Year > 2000;` This will return this view: which is filtered by Year even though it's not a column in the View.
```
pavish, line 36: Do we intend to show this filter in the 'filters dropdown' on the frontend?

I think it's best if we consider this just as part of the query for the view, and not as a filter that can be manipulated.

Reasoning:
* The view is specifically created by applying this filter on the parent table.
* When the user wants to apply filters on the view, I assume they'll want to only further filter the view with the columns shown rather than being able to edit the value for Year.
* Based on our UX, it would not be possible to add this filter for Year back if the user removes it, unless they close and reopen the view.
* It would also confuse non-technical users if a random filter is present on a view which does not contain that column.

Having said that, if we do not intend to show it on the UI, I think it's best to explicitly mention it here.

product/specs/2022-01-views/02-modeling-views.md
```diff
@@ -0,0 +1,84 @@
...
+
+| ID | Title |
+|-|-|
+| 2 | Don't Look Up |
+| 4 | Jason Bourne |
+
+## Sorting
+Views can have sorting applied. Unlike Tables, view sorting is not necessarily related to the columns that are present in the view.
+
+Using the example table above, imagine a view created from the query `SELECT ID, Title FROM Movies ORDER BY Year;` This will return this view, which is ordered by Year even though it's not a column in the View.
```
pavish, line 46: Based on the same reasoning as the filters, I do not think we should be showing this sorting in the 'sort dropdown' on the frontend.

If the user sorts the view, I would assume the backend would sort it in addition to the existing sorting for Year. In essence, if they sort by Title, the sorting would be prioritized first by Title and then by Year (which would function as kind of a default sort).

As mentioned in the filters comments, if we do not intend to show the sorting applied in the query on the UI, I think it's best to explicitly mention that here.

product/specs/2022-01-views/03-modeling-view-columns.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Data Sources
+ - **Definition**: This is the set of source columns that are used to generate the data in the current View column.
+- **Allowed values**: references to other Table or View columns, including other columns in the same View.
+- **Optional**: This could be empty for purely calculated columns (e.g. using the Postgres `random()` function and putting the output in a column)
+
+### Data Formula
+- **Definition**: This is the formula used to generate data in for this column.
+- **Allowed values**: SQL function + operators
+- **Optional**: Columns that are direct copies of other columns from tables or views won't have a formula.
```
pavish, line 26: Should Data sources and Data formula column types be specific to views?

I assume tables would also need them for generated columns created using GENERATED ALWAYS AS, and for auto-incremented columns.

For tables, we'd allow editing all columns if they are not a primary-key and it would fail if the connected db has tables with such columns. We should atleast have a way to identify them for tables to disable the input on the frontend.

product/specs/2022-01-views/04-ui-requirements-for-views.md
```diff
@@ -0,0 +1,78 @@
...
+- **For alpha release**:
+	- Users should be able to see what sorts are applied to their View.
+	- Users should be able to edit and delete sorts applied to their view, including basic use cases for columns that are not visible in the View.
+- **Potential future features**:
+	- Improvements to what columns can be used for sorts to support more complex use cases.
+
+## Aggregations
+- **For alpha release**:
+	- Users should be able to see what aggregations are applied to their View.
+	- Users should be able to edit and delete basic aggregations applied to their view (such as `DISTINCT` and `GROUP BY`)
```
pavish, line 56: Fir reasons mentioned above for filters and sort, I don't think we should support this.

We should probably only show what aggregations are applied when the user views the query.

product/specs/2022-01-views/02-modeling-views.md
kgodey: > the 'filters dropdown' on the frontend?
> [...]
> Based on our UX

Please don't assume any particular design for Views as yet. This spec is aimed at clarifying product requirements which will then influence the requirements for design.

kgodey:
it might be useful to look through Chart.io docs while reviewing the spec, I took some inspiration from them. they have the a good example of breaking down SQL query building visually, which is essentially what we're doing to create and interact with Views:
- https://chartio.com/docs/visual-sql/
- https://chartio.com/docs/visual-sql/start-a-query/visual-mode/

product/specs/2022-01-views/02-modeling-views.md
kgodey: e.g. see this UX for adding a filter in Chart.io https://chartio.com/docs/visual-sql/start-a-query/visual-mode/#add-a-filter.

Even if we don't allow the user to manipulate the filters, it seems like it would be useful to show the filter applied (without them having to look at the query or understand SQL), otherwise non-technical users may get confused about why the View is not showing all the data they expect. And if we're figuring out how to represent the filter, that's most of the work, so why not let them change it?

product/specs/2022-01-views/02-modeling-views.md
kgodey: We can discuss this in one place under filtering, I think whatever conclusion we get to will apply.

product/specs/2022-01-views/04-ui-requirements-for-views.md
kgodey: We can discuss this in one place under filtering, I think whatever conclusion we get to will apply.

product/specs/2022-01-views/03-modeling-view-columns.md
kgodey: I think it would make sense to add Data Sources and Data Formula to generated columns in tables, I hadn't thought of that.

seancolsen (approved):

product/specs/2022-01-views/02-modeling-views.md
pavish: Okay, that makes sense.

I've been explaining to people lately that tables are a structured representation of raw data and views are reports generated from those tables. When I look at views as reports, it kind of makes them seem immutable, which is true from a db standpoint.

But considering that we intend to represent views as mutable entities on the frontend, allowing users to edit the base filter makes sense.

I'm only concerned on how well we would be able to represent the differences between tables and views to the user.

product/specs/2022-01-views/04-ui-requirements-for-views.md
```diff
@@ -0,0 +1,78 @@
...
+## Query
+- **For alpha release**: Users should be able to see the query associated with a View after it has been loaded.
+- **Potential future features**:
+	- Users can edit the query associated with a View.
+	- Users can create a new View using a query.
+
+## Columns
+- **For alpha release**: Users should be able to see all columns associated with a view. Each column should show:
+	- Data Type (non-editable)
+	- Data Sources (editable if there's no Formula)
```
mathemancer, line 24: I don't understand the goal of making the data sources editable.  Would the intention be to drop a different column in in place of the one shown?  Wouldn't it be simpler / smoother to remove that column and add a different one?

product/specs/2022-01-views/04-ui-requirements-for-views.md
```diff
@@ -0,0 +1,78 @@
...
+	- Data Formula (editable in simple cases)
+- **Potential future features**:
+	- Show virtual columns involved in view creation (via CTEs, subqueries, etc.)
+	- Allow editing data formula using SQL
+
+## Rows
+- **For alpha release**: 
+	- Users should be able to see the rows associated with a view.
+	- If a cell is a direct representation of a record, users should be able to edit that record via that cell.
+	- If a column is a direct representation of a record, users should be able to add a new record via that cell.
```
mathemancer, line 34: To double-check; the intent would be to pop up some "record input" UI when trying to edit the cell, correct?  I.e,. since the entire table wouldn't be there, they'd need some interface to input the rest of the row in the associated table.  What if the column is the join column?  Would they edit the record in both tables, or just one?  (Or does that not count as a direct representation?)

mathemancer (commented):
I think most of this makes sense.  I've noted some spots that may be some trouble in specific comments.

product/specs/2022-01-views/02-modeling-views.md
mathemancer: Argh, I couldn't see this conversation while doing my own review, so there's some overlap.  To reiterate what I said in other spots, I think we're missing clarity between the filter that defines a view, and a filter applied to a view.  I think we need both, and since from SQLAlchemy a view is a table (more-or-less), filtering a previously created view is the same as filtering a table.  I think we should be crystal clear on this point, lest we create confusion for users.

This applies to all transformations that can be used to define a view.

product/specs/2022-01-views/04-ui-requirements-for-views.md
ghislaineguerin: I think that if users want to reuse a view w a different table, this might come in handy. I think removing/adding a new column is probably easier w SQL than it is through the interface.

ghislaineguerin (approved):

product/specs/2022-01-views/02-modeling-views.md
mathemancer: I'm not sure that helps.  For me most natural would be:
- fiddle with a table (or tables) to get things set up like I want, then
- save that as a view, then
- recurse (i.e., do the same thing, but on the view).

When it comes to all the interactions that create a view (filtering, grouping, joining, choosing columns, etc.), a view has (at the DB level) precisely the same functionality as a table.  I think reflecting that in the UI is really empowering, since it lets a user iterate over different views to get what they want instead of needing to foresee everything when initially creating the view.

mathemancer: The problem, then, is how to show the previous filter (or sequence of filtration steps) in a user-friendly way.  It'd be visible in the query if you know what you're looking for, but that won't be very friendly to non-technical types.

product/specs/2022-01-views/04-ui-requirements-for-views.md
mathemancer: Now I definitely don't understand.  Do you want them to use sort of the same processing, but on a different data set?  I.e., maybe apply the same sequence of filters and grouping?  Do you have an example of what you're envisioning there (i.e., with actual table names and fake data)

product/specs/2022-01-views/02-modeling-views.md
kgodey: I think that rather than having two different types of "filters", it would be better to just have one concept of filters and either portray them as "saved" or "unsaved". 

"Saved" filters are part of the query. "Unsaved" filters are only visible in the UI.

product/specs/2022-01-views.md
```diff
@@ -0,0 +1,30 @@
...
+editor: markdown
+dateCreated: 2022-01-13T19:49:54Z
+---
+
+**Date**: 2022-01
+**Author**: Kriti Godey
+
+# Introduction
+
+Fundamentally, **Views** are saved database queries. This means that in order to work with Views in Mathesar, we need to translate every concept that can be used in [PostgreSQL queries](https://www.postgresql.org/docs/14/queries.html) to our end users in a user-friendly way.
```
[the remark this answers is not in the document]
kgodey: I figured we'd be able to break down views previously defined by some DB query into the concepts defined in this spec (as long as we have access to the query, which I assume we would if we had access to the view). Is this inaccurate?

product/specs/2022-01-views/01-assumptions.md
```diff
@@ -0,0 +1,17 @@
...
+tags: 
+editor: markdown
+dateCreated: 2022-01-13T19:49:54Z
+---
+
+## Assumptions
+I'm making the following assumptions in the rest of the spec about how we want to work with Views in Mathesar.
+
+- We **do not** need to support creating or editing Views based on every conceivable database query in Mathesar. We will be focusing on allowing common use cases.
+- We **do** need to support viewing Views based on any conceivable database query correctly, even if they can't be edited. Users should be able to connect a database with existing Views to Mathesar and have those Views show up correctly.
```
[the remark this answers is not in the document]
kgodey: Yes, that's the idea. If there's a filter we don't understand, I think we would show it as an unknown filter in the UI and not allow them to edit that part.

product/specs/2022-01-views/01-assumptions.md
```diff
@@ -0,0 +1,17 @@
...
+- At the moment, we **only** care about the final output of the views. If a view uses a subquery, CTE, union, intersection, etc. internally, we will not be representing those to the user in the UI (unless they look at the underlying SQL query).
```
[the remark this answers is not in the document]
kgodey: That is correct. But if a view was created outside of Mathesar and involved a CTE creating some derived columns that don't reflect in the final tabular output of the view, the user will not see anything about those derived columns.

product/specs/2022-01-views/02-modeling-views.md
```diff
@@ -0,0 +1,84 @@
...
+title: Views: Modeling Views
+description:
+published: true
+date: 2022-01-13T19:49:54Z
+tags: 
+editor: markdown
+dateCreated: 2022-01-13T19:49:54Z
+---
+
+Here's how I think we should model views in our API and UI. Each heading represents an attribute of Views.
```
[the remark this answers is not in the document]
kgodey: Please see comment about "saved" and "unsaved" filters above.

product/specs/2022-01-views/03-modeling-view-columns.md
```diff
@@ -0,0 +1,28 @@
...
+- **Definition**: This is the final data type of the content of the column after any computations etc. are applied.
+- **Allowed values**: same as Table data types.
+- **Required**. Data type should always be set, at the very least, we can treat unknown data types as text.
+
+### Data Sources
+ - **Definition**: This is the set of source columns that are used to generate the data in the current View column.
+- **Allowed values**: references to other Table or View columns, including other columns in the same View.
+- **Optional**: This could be empty for purely calculated columns (e.g. using the Postgres `random()` function and putting the output in a column)
+
+### Data Formula
```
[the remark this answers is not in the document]
kgodey: The formula would include the sources, yes.

I'm borrowing Element's UI to illustrate the idea, imagine Matrix channels are data sources.

Formula would be something like:
Screen Shot 2022-01-20 at 3 52 16 PM

Sources would just be list of variables used in the formula:
Screen Shot 2022-01-20 at 3 51 34 PM

product/specs/2022-01-views/04-ui-requirements-for-views.md
kgodey: I don't think we can reuse views with different tables, at least not in any easy way that's worth doing for the alpha release.

I'll remove data sources being editable, it would be better to drop and create new columns.

product/specs/2022-01-views/04-ui-requirements-for-views.md
kgodey: Yes, that's the idea.

I think if it's a join column, it should be updated in both tables. I'll specify.

product/specs/2022-01-views/04-ui-requirements-for-views.md
```diff
@@ -0,0 +1,78 @@
...
+- **Potential future features**:
+	- Users can edit the query associated with a View.
+	- Users can create a new View using a query.
+
+## Filters
+- **For alpha release**:
+	- Users should be able to see what filters are applied to their View.
+	- Users should be able to edit and delete filters applied to their view, including basic use cases for columns that are not visible in the View.
```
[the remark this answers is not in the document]
kgodey: Both, I'll add more details about "saved" vs. "unsaved".

product/specs/2022-01-views/03-modeling-view-columns.md
kgodey: Added an issue to track product work for this: https://github.com/centerofci/mathesar/issues/1002

product/specs/2022-01-views/04-ui-requirements-for-views.md
kgodey: Actually, on second thought, join columns should not count as direct representations. Only columns with a single data source count.

product/specs/2022-01-views/03-modeling-view-columns.md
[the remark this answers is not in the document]
dmos62: I like how you made that illustration, Kriti. Creative. I'm leaning towards dropping Data Sources, and querying the Data Formula for what columns it references. That way there would be a single source of truth for this. Having Data Sources at the same level of interface as Data Formula is a bit in conflict with the fact that one is derived from the other.

product/specs/2022-01-views/02-modeling-views.md
silentninja: > "Saved" filters are part of the query. "Unsaved" filters are only visible in the UI.

This would get rid of concept of the views and users would just see see it as "Saved filters" and can be misleading, if we happen to cache or store the filters locally. Moreover it might be confusing for views that are very deep. 
`View name` should be enough to know about the data that belongs to the view. For example, `views` for the table `Movies` would be named as "Horror Movies" enough to know what data to expect.

 I would rather have a dependency chart that would tell me the information how the table was derived than have a cluttered filters

kgodey: > This would get rid of concept of the views.

I don't see how this would get rid of the concept of views. Could you elaborate?

> users would just see see it as "Saved filters"

Views _are_ saved filters, among other things. Views can be used as:
- a way to save filters/sorts of a single table.
- a way to see data from multiple tables joined together.
- a way to aggregate data (either across a single table or multiple tables)
- a way to have computed/derived columns

...and probably more.

> `View name` should be enough to know about the data that belongs to the view. For example, `views` for the table `Movies` would be named as "Horror Movies" enough to know what data to expect.

I don't think we can rely on users always naming views well. I'm also not sure how this point relates to the product spec or filters.

> I would rather have a dependency chart that would tell me the information how the table was derived than have a cluttered filters

I'm not sure what you mean here by "a dependency chart" or "cluttered filters". Could you explain or maybe do a quick wireframe of what you're imagining for both of those things?

product/specs/2022-01-views/03-modeling-view-columns.md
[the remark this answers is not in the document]
kgodey: We won't always have a Formula, e.g. a view that's built on the query `SELECT Title, Release Year, Rating FROM Movies;` won't have any Formula for `Title`, `Release Year` or `Rating`. It will have a Source for each column, though. 

We also may not have a Source for a column with a Formula - e.g. a column might be generated using the `RANDOM()` function, which would be the Formula. There's no source data for it, though.

product/specs/2022-01-views.md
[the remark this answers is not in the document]
mathemancer: That might be possible, but would be very (_very_) difficult under all but the simplest circumstances.  Moreover, my point is that even if we could do that, we don't need to in order to be able to work with a view.  Having access to the underlying query isn't necessary, since we can treat a view as a table for any of the operations we support.

product/specs/2022-01-views/02-modeling-views.md
mathemancer: The problem with just having saved filters is that they compose, and the composition order can be important (e.g., for "is distinct" filtering).  So, a view might be a filtered version of a join of two views, each of which is formed by filtering some table or another.  I think any UI that attempts to show the arbitrary composition of filters (or other operations used to create a view) is going to have the potential to become pretty cluttered pretty quickly.  I don't know if a dependency graph is the right solution, but I think it's crucial that a user understands when they're applying a filter to a view as opposed to an underlying view or an underlying data set.  I don't think "saved filter" gets that point across.

silentninja: > >"Saved" filters are part of the query. "Unsaved" filters are only visible in the UI.

> I don't see how this would get rid of the concept of views. Could you elaborate?

Views are more than just saved filters. They can be referenced by other queries/views,  can have their own set of filters/sorts, more of hierarchical structures which are closer to a table in characteristics. If we call them unsaved filters, it would be misleading as `unsaved filter` looks like something that cannot be extended up, something that either can be persisted or can't be and if persisted can be concatenated to existing filters
In the case of very deep Views(table1-> view1->view2), they need to show that hierarchy and users should be able to determine where certain operations like filter/sort took place.

silentninja: > Views are saved filters, among other things. Views can be used as:

Hierarchy and composition are the reason why I would not call them saved filters.

mathemancer: @silentninja 

> Hierarchy and composition are the reason why I would not call them saved filters.

This is a succinct way to put it; I agree.

silentninja: > I don't think we can rely on users always naming views well. I'm also not sure how this point relates to the product spec or filters.

Views should be treated as similar to a table. so a view name could help with understanding what the data the view holds instead of having to look where or how it was generated. This is more of a convention rather an accurate description.  If named properly, we don't have to worry about showing how the view was generated unless required, where we could do complex diagrams like dependency graphs.

 Generally, when working with a `View` I wouldn't worry about how it was generated, rather would be focusing on what to do with the data it holds. This again is a convention, so it wouldn't apply to all the situations just like you mentioned

silentninja: > I'm not sure what you mean here by "a dependency chart" or "cluttered filters". Could you explain or maybe do a quick wireframe of what you're imagining for both of those things?

Dependency graph: 
img

As for cluttered filters, I don't have visualisation other than our existing filter dropdown, but the reason why I think it would be cluttered is that we could end up with too many filters as a list(maybe along with the entity name they came from) in case of a deep view

product/specs/2022-01-views/04-ui-requirements-for-views.md
```diff
@@ -0,0 +1,100 @@
...
+		- "Direct representation" means that the record has only one data source and no formula.
+	- If a column is a direct representation of a record, users should be able to add a new record of the same type via the View UI.
+- **Potential future features**:
+	- Users can edit the query associated with a View.
+	- Users can create a new View using a query.
+
+## Filters
+- **For alpha release**:
+	- Users should be able to see what filters are applied to their View.
+	- Users should be able to edit and delete filters applied to their view in the UI, including basic use cases for columns that are not visible in the View.
```
silentninja, line 48: What does "including basic use cases for columns that are not visible in the View" mean?

product/specs/2022-01-views/03-modeling-view-columns.md
```diff
@@ -0,0 +1,43 @@
...
+date: 2022-01-13T19:49:54Z
+tags: 
+editor: markdown
+dateCreated: 2022-01-13T19:49:54Z
+---
+
+Here's how I think we should model view columns in our API and UI. Each heading represents an attribute of a View Column.
+
+### Data Type
+- **Definition**: This is the final data type of the content of the column after any computations etc. are applied.
```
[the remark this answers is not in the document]
mathemancer: You can create a function like that, and even use it to generate the column, but if PostgreSQL can't figure out a non-pseudo-type for the column, it'll throw an error.  All polymorphic types are pseudo types.

https://www.postgresql.org/docs/13/datatype-pseudo.html

So, you can't have a column of type, e.g., `ANYARRAY`

product/specs/2022-01-views/03-modeling-view-columns.md
```diff
@@ -0,0 +1,43 @@
...
+- **Allowed values**: same as Table data types.
+- **Required**. Data type should always be set, at the very least, we can treat unknown data types as text.
+
+### Sources
+ - **Definition**: This is the set of source columns that are used to generate the data in the current View column.
```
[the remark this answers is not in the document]
kgodey: Yes, it will reference the immediate parent. If the parent is another view, you'll have to go look at that view to find the source.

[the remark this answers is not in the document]
kgodey: I do like the idea of a dependency graph eventually but I don't think it's worth doing before the alpha release.

product/specs/2022-01-views/04-ui-requirements-for-views.md
kgodey: Assuming `SELECT ID, Title FROM Movies WHERE Year > 2000;` is the query for the view, here the "filter" would be `Year > 2000`. Users should be able to edit that even though `Year` is not a visible column in the View.

kgodey:
@silentninja @mathemancer I've updated the spec to have a new page for modeling View Queries. Queries have their own filters, sorts, and aggregations, which should only be accessible when you're editing the query (if the query is editable). This is separate from filters. sorts, and groups on view _data_.

The "saved filter" concept has been removed, although there will still be functionality to apply whatever filters you have in the UI to the view query.

kgodey:
I'm going to go ahead and merge this. I'll open a new discussion once it's up on the wiki for any further thoughts.

PR #26, opened 2022-01-19
Add more front end code standards

seancolsen:


seancolsen:
@pavish Can you review this?

engineering/architecture/front-end-standards.md
```diff
@@ -77,4 +77,219 @@ dateCreated: 2022-01-14T19:02:19.924Z
...
+
+Notes:
+
+- Not all of our components adhere to this guideline yet.
+
+
+## `type` vs `interface`
+
+- Prefer `interface` when possible.
+- Use `type` when necessary.
```
seancolsen, line 236: @pavish this is some content that we haven't discussed before, but from reading your code I've deduced that we're on the same page about this (which is also the direction that I think the large TS community is going).

engineering/architecture/front-end-standards.md
```diff
@@ -77,4 +77,219 @@ dateCreated: 2022-01-14T19:02:19.924Z
...
+
```
seancolsen, line 237: @pavish this is some content that we haven't discussed before, but from reading your code I've deduced that we're on the same page about this (which is also the direction that I think the large TS community is going).

pavish (approved):
@seancolsen Looks good to me!

PR #28, opened 2022-01-20
Create backend-codebase-organisation.md

### LINKED ISSUE #27: Document the structure of the backend codebase
## Problem

Mathesar a lots of moving parts and it would be better if we have documentation that gives a bird's eye view of the codebase making it easier for someone new to the codebase

## Proposed Solution

[Django](https://docs.djangoproject.com/en/4.0/) has good docs which addresses similar problem, we could build on top of that

silentninja:
Fixes #27

Adds a file to the Engineering section of the wiki

### DIFF
--- a/engineering/architecture/backend-codebase-organisation.md
+++ b/engineering/architecture/backend-codebase-organisation.md
@@ -1,0 +1,20 @@
+How the documentation is organized
+Mathesar has lots of components. A high-level overview of how it’s organized will help you know where to look for certain things:
+
+#### DB Layer:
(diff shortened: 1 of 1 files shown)


kgodey:
@silentninja Is this ready for review? It looks incomplete.

Please also add a link to this from the Architecture page, otherwise no one will be able to find this from the wiki. It's probably also useful to add a link from our README so contributors can find it, see @seancolsen's recent PR https://github.com/centerofci/mathesar/pull/977 for an example.

silentninja:
Deprioritized as of now

PR #33, opened 2022-02-04
Add three new front-end coding standards

seancolsen:


seancolsen:
@pavish Can you review this?

See ba4c1a1e3004e20f2ca65b2779cb53593ef5c9aa for the standards I've added (after doing some preliminary cleanup).

I'm proposing three new standards

- "CSS units" -- we've talked about before and I anticipate you'll be in agreement here.
- "API type definitions" -- we've also talked about this
- "Component spacing and layout" -- is a result of [this code](https://github.com/centerofci/mathesar/blob/db9fa80e7ffdf2520e20b3f5bdbf8a4c4dfb3d64/mathesar_ui/src/sections/table-view/header/header-cell/type-options/database-options/DbTypeIndicator.svelte#L15-L15) I noticed while reviewing https://github.com/centerofci/mathesar/pull/926

    I like to avoid margin on components because it makes the easier to compose.

    If you're in agreement with me on this standard, then for the sake of time I'd suggest we still leave that code in-question for now and not even take the time to open a new ticket for it. We can refactor it to become in conformance with the standard when/if it presents a problem when composing that component.

pavish (approved):
@seancolsen I am in agreement with all the points mentioned.

Regarding avoiding margin on components, I think it should be *strictly enforced* only on common components.

For narrow components used only for specific purposes, such as the one in [the code that is referenced](https://github.com/centerofci/mathesar/blob/db9fa80e7ffdf2520e20b3f5bdbf8a4c4dfb3d64/mathesar_ui/src/sections/table-view/header/header-cell/type-options/database-options/DbTypeIndicator.svelte#L15-L15), it would be good to avoid margins on the root element, but I think it should be a guideline to follow, instead of a strict imposition.

PR #36, opened 2022-02-18
Update number type to support custom currency format

ghislaineguerin:
Fixes https://github.com/centerofci/mathesar/issues/1063


silentninja:
Based on our [discussion on Matrix](https://matrix.to/#/!lHwCNrQmvPOrQQgpmG:matrix.mathesar.org/$TVbsvAbeC7Idysb7YYxHdaAkayMdIEt0MRRIaaZ5iCI?via=matrix.mathesar.org&via=matrix.org), we should have the UX in such a way that `Postgres Money` type is discouraged, since it does not bring any advantage over the `numeric` type + display options, rather has [limitations](https://www.postgresql.org/docs/9.1/datatype-money.html)

kgodey:
@ghislaineguerin This spec does not address this design problem from https://github.com/centerofci/mathesar/issues/1063

> Users may be confused by two different types to represent money (Money, and Number + display options). We should figure out how to represent this to the user in a way that's easy to understand.

Some quotes from the Matrix discussion about this topic:

@pavish:

> If a non-technical user wants to add a money column, how will they know which to use? We provide two separate options, they might add one wrongly or might want to switch from one to another, in which case the display options of number (eg., choosing currency symbol) will be lost.
>
> Their first instinct will be to go to the Money type, and we could have limitations and warnings there, but how do we get the message across that it's a database issue and they'll have to tweak db config, especially considering that they're non-technical.

> We would like them to use the display settings of Number, but our UX shows Money more prominently and display options are hidden. 
>
>  It is a guarantee that the user will always go to the Money type, and how do we redirect it effectively?

kgodey:
Potential solutions, not all of these are mutually exclusive.
- Display money type as "Dollars" (or whatever the locale is set to)
- Show a warning message when using Money type that we recommend that users use the Number type with currency formatting instead
- Create a custom "money" type that's actually just a alias for `NUMERIC` and will need display options set to be useful. Move currency formatting options there.
- Don't allow users to change column type to the Money type, only show the Money type if there's an existing column that uses it.

Other solutions proposed were:
- Remove currency formatting from number types entirely (this is not an option since we still need a way for users to store columns with different currencies)
- Remove support for Money type entirely (this seems unwise since we want to support all default Postgres types)

design/specs/money-data-type.md
```diff
@@ -9,31 +9,72 @@ dateCreated: 2021-08-05T09:14:25.201Z
...
+
+## Update 23 Feb 2022
+
+### Compatibility with Postgres Money Type
+
+Mathesar will allow users to view columns set to Postgres money type as a money type column. However, it will not allow users to set an existing column to Postgres money data type. To change from Postgres money data type to the custom Mathesar money type, the user would have to set the column to Number or other data type and then choose money again.
+
+#### Postgres Money Type Prototype
+
+[Figma Prototype - Postgres Money Type](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=7552%3A83433&node-id=7646%3A84762&viewport=241%2C48%2C0.33&scaling=contain)
```
kgodey, line 56: The warning should say:
> "The currency setting of this database type cannot be changed."

_If_ the custom Money type is available, it should say:
> To use a different currency, please change this type to Number and back to Money.

_If_ the custom Money type is not available, it should say:
> Please contact an administrator if you'd like to change the currency setting.

The frontend will know whether the custom Money type is available, so it can use the appropriate message depending on the availability.

design/specs/money-data-type.md
```diff
@@ -9,31 +9,72 @@ dateCreated: 2021-08-05T09:14:25.201Z
...
+
+### Support for Multi-Currency through custom Mathesar Money Type
+
+Mathesar's custom money data type will allow users to set any currency and customize various display options. The custom money data type will be the only money type option that can be set in Mathesar. With Postgres money type being available only for existing database columns set to this type.
+
+#### Custom Mathesar Money Type Prototype
+
+[Figma Prototype - Custom Mathesar Money Type](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=7552%3A83433&node-id=7590%3A84021&viewport=241%2C48%2C0.46&scaling=contain)
```
kgodey, line 64: Looks good to me. The only change I'd like is to change the "Symbol Location" options to Beginning and End, I think it's clearer.

@pavish Can you take a look and make sure this works for frontend formatting?

design/specs/money-data-type.md
```diff
@@ -9,31 +9,72 @@ dateCreated: 2021-08-05T09:14:25.201Z
...
+- The user sets a column type to `Money.`
+- The user opens the data type menu and goes to the `Display` section.
+- The user sees the default currency selected and [additional options](#currency_format_options) are displayed.
+- Changing any of the default options for a currency will set the currency to custom. For example, changing the symbol for US dollar from `$` to `USD`.
+- Changing to any other currency will update the values of the formatting options to those corresponding to the currency.
+
+#### Currency Format Options
 
+- Select Symbol: Users can set the symbol that will be displayed along with the number.
+- Symbol position: Users can select the position for the currency symbol. The options are start or end.
```
kgodey, line 77: These don't seem to match what's in the prototype.

kgodey (changes requested):
Looks good @ghislaineguerin, some minor changes requested.

design/specs/money-data-type.md
pavish: The options look good, it would suffice for frontend formatting.

The number of inputs is many for the display options, it will definitely exceed the window size. We'll need a scrollbar in this pane or change the width for the dropdown. The max-height needs to be fixed.

@ghislaineguerin It would be better to have a prototype with scrollbars.

design/specs/money-data-type.md
ghislaineguerin: I've fixed this to match the prototype.

design/specs/money-data-type.md
ghislaineguerin: This has been updated.

design/specs/money-data-type.md
ghislaineguerin: This has been updated.

design/specs/money-data-type.md
kgodey: @ghislaineguerin I don't see a scrollbar in the prototype.

kgodey (changes requested):
@ghislaineguerin Looks good, the only comment is that I don't see a scrollbar in the prototype (per @pavish's comment).

PR #37, opened 2022-03-10
Updated common-issue.md

sahilsaini110:
Added solution to error getting while running docker-compose up.
here is the preview of the above.

Screenshot from 2022-03-10 12-07-53


seancolsen:
Thanks @sahilsaini110 :grin:

PR #38, opened 2022-03-14
Data Explorer Specs

ghislaineguerin:
Fixes #1065


design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+`Data Explorer` is meant to be a part of multiple flows like the creation of new views, the inspection of existing ones, and as a way to explore data for goals beyond view creation. For this reason, different starting points are considered from which `Data Explorer` might be accessed.
+
+### User opens the Data Explorer to create a new view
+
+#### From Scratch (Top Navigation)
+
+- In the `Schema Browser` view, a user clicks on the `Data Explorer` navigation item located in the top navigation bar.
+- `Data Explorer` opens, and the layout changes. The schema explorer sidebar is hidden, as well as any open table.
+
+#### From a Table (Sidebar Navigation or Open Table)
```
kgodey, line 30: @ghislaineguerin I think this scenario is out of scope for this issue, there are many details to consider here. I made a separate issue for this, see: https://github.com/centerofci/mathesar/issues/782

Secondly, we need to make the distinction between queries and views more apparent. Anything opened in the data explorer is still at the query stage, it has not yet been saved as a view. If we have a button called "Create View", then the user will expect us to have created a view. If we just want to open the table in the data explorer, then we should name the button differently. But we can handle all of that in https://github.com/centerofci/mathesar/issues/782, it seems better to remove this scenario and keep this spec focused on the query builder interface rather than how to get to it. Let me know if you disagree.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+- From an open table, a user clicks the `Create View from [Table Name]` button located in the toolbar. This action will open `Data Explorer` and automatically set the table as `Base Table` and select all direct columns.
+- In the `Schema Browser` view, a user opens the contextual menu for a table navigation item and selects `Create view from Table`. This action launches `Data Explorer` and automatically sets the selected table as `Base Table` and adds its direct columns to a `Select Column` step.
+- The system will also add link columns if present in the `Base Table`. However, it won't include link columns referencing the `Base Table` from other tables. The system will set the values for link columns to the referenced table's primary key by default.
+- The user can then choose to remove or add additional columns.
+
+### User opens the Data Explorer to inspect an existing view
```
kgodey, line 37: There's also a separate issue for this: https://github.com/centerofci/mathesar/issues/783.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+Views require a base table to be set so that the available columns and links can be listed for selection.
+
+### User sets the base table for a new view
+
+- For users starting a new view from scratch, `Base Table` selection will be the first required step.
+- Once `Data Explorer` is launched the user will be prompted to make a selection. The system will list all tables from the active schema. See *Wireframe 1* for an example of the base table selection prompt.
+- Once the `Base Table` is selected, a second step for `Select Columns` is automatically added and the user is prompted to continue with the view creation process.
+- At this point, the result table is empty.
+
+[Wireframe 1](https://share.balsamiq.com/c/jp7WFu663j6sbfKNknuq5Z.png)
```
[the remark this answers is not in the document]
kgodey: Idea: show the skeleton of the layout of the query builder with some placeholder steps and data. That way, it'll show the user the interface and give them the context for selecting the base table.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+### User inspects the base table for an existing view in Data Explorer
+
+- When inspecting an existing view, a user can find the `Base Table` as the first step of the `Workflow Sidebar`.
+- The `Base Table` step cannot be removed, only edited, causing all subsequent steps to be cleared.
+
+[Wireframe 2](https://share.balsamiq.com/c/wTEzQcBamoBBWBSSF5Cmuy.png)
```
kgodey, line 66: I'm wondering if we need this to be an explicit "step" given that the user cannot remove it. Here is a rough idea:

2 - Base Table Columns

Reasoning:
- I don't think we need the user to learn the term "base table" I'd like to have an interface where the user needs to learn as few terms as possible. 
- I think this also conveys that the base table affects the entire exploration (query), which is something we want users to learn.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+### User opens a view where the base table is missing
+
+- If a user deletes the base table for a view, they should be able to open the view in `Data Explorer` and either resolve the error by replacing the base table with an identical one or clear the workflow and start over.
```
kgodey, line 70: This scenario will never happen. If the user tries to delete the base table, depending on the options, it will either:
- refuse to delete the base table because it has dependent objects (i.e. the view)
- delete the view along with the table.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Worflow: Column Selection
+
+In the column selection step, users select columns from or linked to the base table of a view.
+
+### User adds a direct column
+
+- Direct columns will be those with values contained directly in the `Base Table` rather than linked from other tables via foreign key relationships.
```
kgodey, line 78: I was thinking that direct columns would be any column with a value coming directly from _any_ table, not just the base table. For example, if a Movie had an FK to Language, then language name would still be a direct column since the value comes from a table.

kgodey: If you want a distinction between base table columns and columns from links, I would call them "base table columns" instead of "direct columns".  That will let us use "direct columns" for any non-formula columns.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+- Direct columns will always be listed at the top of the column list in the same order as the base table.
+
+[Wireframe 3](https://share.balsamiq.com/c/hNWzHEM9ksGGp5ZcPixLrA.png)
+
+### User adds a linked column
+
+- To add a linked column, the user can click on any column listed under the `Base Table Links` or `Linked to Base Table` sections. Linked columns are listed under the tables for which a link to or from the `Base Table` exists.
+- The linked column can then be expanded to show all possible values for that field. The default value points to the primary key of the referenced table, however, the user might choose to display values from any other column.
+
+[Wireframe 4](https://share.balsamiq.com/c/hNWzHEM9ksGGp5ZcPixLrA.png)
```
kgodey, line 89: I assume you'll add a wireframe to show the expansion.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+### User adds multiple select column steps
```
kgodey, line 91: I'm still not convinced about allowing multiple steps for column selection. I think it would be better to allow users to add columns in a single step and dynamically update what columns are available to use based on the columns that have already been added.

For example, imagine user adds direct column A. We would only show base table columns + linked columns as options as available columns to use. 

Then, imagine they add formula column B. Then, we show base table columns + linked columns + column A from the query as available columns to use.

Then, imagine they add formula column C. Then, we show base table columns + linked columns + column A and B from the query as available columns to use.

I think this is clearer than having multiple select column steps.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Results Table and Previewing
+
+### User previews the result table for all steps
+
+- The result table (preview) by default contains the output of all steps in the workflow sidebar. Changing any of the steps or adding new ones will automatically affect the table's contents.
+
+### User previews the output of a specific step
+
+- Users can temporarily change the preview to show the workflow's output at a particular point.
```
kgodey, line 152: I like this feature!

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+### User adds a summarization step
+
+- A user can summarize the result table at any point in the workflow by adding a `Summarization` step.
+- The summarization step will let users set a column as the summary column. By doing so, The system will aggregate the rest of the columns.
+- The summary column might or might not be a part of the result table—any formula columns created before the summarization step can be set as a summary column.
+- Once a summary column is set, the aggregated columns will be listed, each with a suggested aggregation function. Users can change the suggested functions.
+- A user can add more than one summarization step.
+
+[Wireframe 7](https://share.balsamiq.com/c/2e6TYLVzqf4pssmfpNqacx.png)
```
kgodey, line 142: This wireframe doesn't have anything related to summarization, I assume you'll update it.

design/specs/visual-query-builder.md
ghislaineguerin: @kgodey what if the base table was deleted outside of Mathesar?

design/specs/visual-query-builder.md
ghislaineguerin: @kgodey I understand that we want to simplify, but I feel that we are mixing two different models here... the "step based" and this more "dynamic list" approach.
My concern, is that we need to persist the original selection somewhere, as those are feeding into the next steps. In the same way, a formula column created in step 8 for example, cannot be selected in step 1.

[1. Set base table ]
[2. Select Columns   movie.title.  movie_crew.actorId(as person.full_name)  ]
[3. Add Formula:  actorInitials.  text from:  movie_crew.actorId(as person.full_name)]
[4. Summarize by ActorInitials  and aggregate sum(movie.title)]

I now have a table with Initials and a count of movies per actor name initial. I no longer want the person full name

[5. Select Columns : Actor Initials and sum(movie.title)]

I want to filter the actor name list, before they were initials... at this point I no longer have access to that column, I need to filter before the column selection.

In the dynamic list approach, everything would always be listed, even in steps before they are created (in the case of formulas) so it would stop being a true step based UX. I feel those inconsistencies will confuse users more than multiple select steps. And I'm not sure how, if we don't follow the step based solutions, we'll be able to have more than one filter or summary step.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+### User deletes a formula column
+
+## Filtering
+
+In the `Filter` step, users apply filters to columns in the result table.
+
+### User adds a filter step
+
+- A user can filter selected columns at any point in the workflow by adding a `Filter` step.
```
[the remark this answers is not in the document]
ghislaineguerin: @kgodey In the step-based approach, how is input filtering different from output? aren't those just filters set at different points in the workflow?

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,192 @@
...
+
+- Users can temporarily change the preview to show the workflow's output at a particular point.
+
+[Wireframe 8](https://share.balsamiq.com/c/5zwJEcEV7ob2TTwTU3TBgX.png)
+
+### User creates a view from the output of a step
+
+## Layouts
+
+### Data Explorer Layout
```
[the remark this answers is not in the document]
ghislaineguerin: @kgodey in your second wireframe, how would aggregation at the column level work compare to summarization?

design/specs/visual-query-builder.md
kgodey: Not responding here since @ghislaineguerin and I discussed the overall flow on a call.

design/specs/visual-query-builder.md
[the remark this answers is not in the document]
kgodey: Not responding here since @ghislaineguerin and I discussed the overall flow on a call.

design/specs/visual-query-builder.md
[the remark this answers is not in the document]
kgodey: Not responding here since @ghislaineguerin and I discussed the overall flow on a call.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,140 @@
...
+
+Once a selection has been made, the user will be prompted to proceed with the input table configuration, starting with column selection.
+
+[BASE TABLE SELECT WIREFRAME]
+
+## 2. Creating the Input Table
+
+The input table is created by selecting columns from the base table or creating new ones via formulas.
+
+Various options are available for input columns, including filters or aggregations that can help users control which data they want to show and how.
```
kgodey, line 45: Filters and aggregations are only available for columns that have multiple records.

For example, if you have the following structure:

---
### Movie

| ID | Title | Runtime | Year |
|-|-|-|-|
| 1 | Encanto | 109 | 2021 |

### Actor

| ID | Name |
|-|-|
| 462 | Stephanie Beatriz |
| 463 | John Leguizamo |

### Movie Actor Map

| ID | Actor ID | Movie ID | Character |
|-|-|-|-|
| 34 | 462 | 1 | Mirabel Madrigal |
| 35 | 463 | 1 | Bruno Madrigal |

---

Imagine the user sets the base table to "Movie" and adds "Movie > Title" as the first column. 

Now if they add "Movie > Runtime", there won't be any filters or aggregations available for that column because each movie has only one runtime, there's nothing to filter or aggregate.

Instead, if they add "Actor > Name", there will be filters and aggregations available since there are multiple actors associated with a single movie.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,140 @@
...
+
+These options can be set for all input columns, including formulas. Some options like aggregations might only be available for columns that link to multiple values. The system will automatically detect this and present the available options to users.
```
kgodey, line 47: I think you should specify that filters and aggregations specifically will only be available for columns that involve pulling data from multiple records. It's clearer when implementing.

It might be good to document all of the options available for a column and rules for when they are available.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,140 @@
...
+
+When selecting input columns, the user can add them using the list controls or drop the columns directly into the result table from the column selector.
+
+The column selector component will display all columns from the base table, as well as columns from any table linked or containing links to the base table.
+
+#### Direct Columns
+
+Direct columns are columns from the base or linked tables that contain the actual stored values, these can be selected directly and added as input columns.
+
+Any non-direct column from the base table that was set as a link in the base table will not be available for selection.
```
kgodey, line 59: I'm not sure what this means.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,140 @@
...
+
+[COLUMN SELECTOR WIREFRAME]
+
+### 2.2 Adding Formulas
+
+During column selection, the user can add a new column by using formulas to generate its values. Available formulas will be included as part of the column selector, and dropping them into the result table will prompt users to set the formula settings for the selected formula.
+
+Depending on the selected formula, different settings will be available. Formulas will use the column selector component for settings that require the selection of columns. However, the system only list those with data types accepted by the formula.
+
+[WIREFRAME SHOWING FORMULA COLUMN]
```
kgodey, line 79: I think eventually we will need UX (wireframes) for all the different variable types supported by formulas.

kgodey: It can either be in this spec or a new spec.

kgodey (approved):
@ghislaineguerin This looks great overall. I added some comments but they're fairly minor.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+Columns that link to multiple records can have filters applied to them in order to retrieve only values that match user-specified criteria. Multiple filters are allowed for each input column.
+
+Filter options will be determined by the data type of the input column.
+
+Filters can be added by clicking on the `Add Filter` option from the query configuration panel or directly from the results table column header menu.
+
+---
+Wireframes
+
+[Added Input Filter](https://share.balsamiq.com/c/tK2hZy6FB4zVYpN56ejpBk.png)
```
kgodey, line 98: Looks good. Nitpick: The column name in the table and inspector are different (Total Actors vs. Actors)

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+### 2.4 Aggregating Input Column Values
+
+Input columns can be aggregated in cases where a link might be referencing multiple values. In such cases, and based on the data type of the referenced column, the system will apply an automatic aggregation. Users can change the aggregation type at any moment.
+
+[Added Input Aggregation](https://share.balsamiq.com/c/fNcbvQ1q5xMyni5JxBg3Rc.png)
```
kgodey, line 104: I do think it makes sense to represent the aggregation as a formula (to avoid having two separate concepts). Turning the aggregation off would be removing the formula.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+### 2.5 Applying a Formula to an Input Column
+
+Direct input columns can also be transformed into formulas. The available formulas will depend on the input column type.
+
+To transform the input column into a formula, the user will select the `Apply Formula` option in the column header menu.
```
kgodey, line 110: Can we have a wireframe for this?

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+## 3. Transforming the Output Table
+
+The output table refers to the resulting table from all input columns selected and added, including filters and aggregations. Transforming this output can be done by applying output filters and summarizations to the result table.
+
+### 3.1. Filtering the Output Table
+
+Users can filter the output table by selecting any column from the result table and applying filters. The column selector, in this case, will only allow users to select input columns rather than the complete column list from the column selection step.
+
+[Output Filter](https://share.balsamiq.com/c/dAVYnp8VnG2r2HzkSZ3rgp.png)
```
kgodey, line 120: Looks good

Should we have the ability to add another filter step after summarization?

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+### 3.2. Summarizing the Output Table
+
+Users can summarize the output table to get an aggregation of the column values grouped by a summary column.
+
+To summarize a table, a user selects the `Summarize` option and sets a summary column. The rest of the columns are automatically assigned an aggregation function which users can change at any point. The aggregations are inferred based on the data type of the output column.
+
+[Output Summary](https://share.balsamiq.com/c/dAVYnp8VnG2r2HzkSZ3rgp.png)
```
kgodey, line 128: This is the same wireframe as filters, can we have a summary-specific wireframe?

kgodey: We should have the ability to add multiple summarization steps.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+## 5. Saving the Query as a View
+
+Under `Save Options`, the user can name the resulting query and choose to save it. This would save the view as a query, that would be run everytime the user opens it.
+
+## 6. Troubleshooting and Resolving Errors
+
+### The result table has no rows
+
+### The are duplicate values
```
kgodey, line 142:
```suggestion
### There are duplicate values
```

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+## 7. Alerts and Error Prevention
+
+### User attemps to delete a base table
```
kgodey, line 146:
```suggestion
### User attempts to delete a base table
```

kgodey (commented):
@ghislaineguerin Looks good overall. I added a couple of comments.

I think the only bigger change is that I think the user should be able to add multiple output filter and summarization steps. For example, imagine

After adding columns:

| title | actor | num_actors |
|-|-|-|
| Venom | Tom Hardy | 4 |
| Venom | Michelle Williams | 4 |
| Venom | Riz Ahmed | 4 |
| Venom | Jenny Slate | 4 |
| Venom: Let There Be Carnage | Tom Hardy | 5 |
| Venom: Let There Be Carnage | Michelle Williams | 5 |
| Venom: Let There Be Carnage | Woody Harrelson | 5 |
| Venom: Let There Be Carnage | Naomie Harris | 5 |
| Venom: Let There Be Carnage | Stephen Graham | 5 |

Output filter 1: `actor` does not contain `Jenny`

| title | actor | num_actors |
|-|-|-|
| Venom | Tom Hardy | 4 |
| Venom | Michelle Williams | 4 |
| Venom | Riz Ahmed | 4 |
| Venom: Let There Be Carnage | Tom Hardy | 5 |
| Venom: Let There Be Carnage | Michelle Williams | 5 |
| Venom: Let There Be Carnage | Woody Harrelson | 5 |
| Venom: Let There Be Carnage | Naomie Harris | 5 |
| Venom: Let There Be Carnage | Stephen Graham | 5 |

Summarization 1: unique `actor`

| title list | actor | num_actors |
|-|-|-|
| Venom, Venom: Let There Be Carnage | Tom Hardy | 4, 5 |
| Venom, Venom: Let There Be Carnage | Michelle Williams | 4, 5 |
| Venom | Riz Ahmed | 4 |
| Venom: Let There Be Carnage | Woody Harrelson | 5 |
| Venom: Let There Be Carnage | Naomie Harris | 5 |
| Venom: Let There Be Carnage | Stephen Graham| 5 |

Output filter 2: length of `title_list` = 2

| title list | actor | num_actors |
|-|-|-|
| Venom, Venom: Let There Be Carnage | Tom Hardy | 4, 5 |
| Venom, Venom: Let There Be Carnage | Michelle Williams | 4, 5 |

--

You can see that the second output filter needs to happen after the summarization since it operates on the summary.

We can also extend this to further steps, like deduplication in #1028

design/specs/visual-query-builder.md
kgodey: One of the things we also need to specify is that aggregation and deaggregation will happen on a "per-table" basis. e.g. if there are two columns A and B from a linked table and A is de-aggregated, B will be too. (e.g. if you have both actor name and actor DOB, de-aggregating actor name will then also deaggregate DOB)

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+
+Links are presented as expandable sections containing the columns from the linked tables.
+
+When adding input columns, the user can add them using the list controls or drop the columns directly into the result table.
+
+---
+Wireframes
+
+[Adding Columns](https://share.balsamiq.com/c/oyxTXxqSh8rLqU3JY71DWY.png)
+[Added Column](https://share.balsamiq.com/c/7U9yahZtYWX2G5obkyivoz.png)
```
kgodey, line 72: We need to figure out how to represent the source's relationship to the base table (similar to how we do it in the column selection flow)

kgodey:
@ghislaineguerin Another thing I just realized is that we seem to have lost the option to add sorts.

kgodey:
@ghislaineguerin I wanted to also follow up on two design elements from my previous suggestions:

Screen Shot 2022-03-22 at 3 35 22 AM

This is the "column details", the things from here I don't see represented are:
- the "source" section showing the relationship to the base table
- the view of all the transformations applied to a column

Screen Shot 2022-03-22 at 3 35 32 AM

And this shows the workflow of adding different kinds of output transformation steps as mentioned in my above comment.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,174 @@
...
+While adding columns, the user can choose to create a new column by using formulas. Available formulas will be included as part of the column selector, and dropping them into the result table will prompt users to set the formula settings for the selected formula.
+
+Depending on the selected formula, different settings will be available. Formulas will use the column selector component for settings that require the selection of columns. The list of columns will be populated based on the selected formulas.
+
+However, the system only lists those with data types accepted by the formula.
+
+---
+Wireframes
+
+[Added Formula](https://share.balsamiq.com/c/vWfJ9sUYWJxed5Zo5WtPGi.png)
```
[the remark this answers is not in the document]
ghislaineguerin: @kgodey I think so, a different issue would be best.

kgodey:
@ghislaineguerin I'm wondering if it makes sense to close this and open a new PR so that my comments are not distracting, since they refer to previous iterations.

PR #39, opened 2022-03-15
Goals and strategies for displaying numbers and accepting user input of numbers

seancolsen:
## Why I opened this PR

In working on [#970](https://github.com/centerofci/mathesar/issues/970), I'm building the component with which the user will interact while editing numbers. This work has surfaced many interesting edge cases regarding numbers. To help guide my implementation, I'd like to first establish some clear over-arching product goals and strategies with regard to numbers in Mathesar.

## Who should review it

If you've also been thinking about the display and input of numbers (and the edge cases we face at various level of our stack), then you might want to look at this.

## Notes

This PR does _not_ say which product goals will be implemented pre-alpha vs post-alpha. I figure we can make those decisions after we've established our vision.

### DIFF
--- a/engineering/architecture.md
+++ b/engineering/architecture.md
@@ -30,7 +30,7 @@ These documents are notes about a specific feature written to get feedback befor
 
 - [Import Previewing](/engineering/architecture/import-preview-api)
 - [Syncing Database Objects](/engineering/architecture/syncing)
-- [Numbers](/engineering/architecture/numbers)
+- [Number Display and Entry](/engineering/architecture/numbers)
 - [Human Readable Dates and Times](/engineering/architecture/dateandtime)
 
 ## Related Resources
--- a/engineering/architecture.md
+++ b/engineering/architecture.md
@@ -30,6 +30,7 @@ These documents are notes about a specific feature written to get feedback befor
 
 - [Import Previewing](/engineering/architecture/import-preview-api)
 - [Syncing Database Objects](/engineering/architecture/syncing)
+- [Numbers](/engineering/architecture/numbers)
 - [Human Readable Dates and Times](/engineering/architecture/dateandtime)
 
 ## Related Resources
--- a/engineering/architecture/numbers.md
+++ b/engineering/architecture/numbers.md
@@ -1,0 +1,106 @@
+---
+title: Number display and entry
+description: Goals and strategies for the display and entry of numerical data
+published: true
+date: 2022-03-15T00:00:00.000Z
+tags: 
+editor: markdown
+dateCreated: 2022-03-15T00:00:00.000Z
+---
+
+## Product goals
+
+- **(Goal A)** Mathesar is capable of displaying and accepting entry for all [numbers which Postgres is capable of storing](https://www.postgresql.org/docs/current/datatype-numeric.html).
+
+- **(Goal B)** Aside from the numbers, `NaN`, `-Infinity`, and `Infinity`, users are able to type all numbers into Mathesar exactly as Mathesar displays those numbers to them.
+
+- **(Goal C)** When a user enters data which cannot be accurately stored in the column exactly as-entered, then Mathesar stores the nearest acceptable value and clearly communicates to stored value to the user.
+
+    For example, a `float8` column cannot accurately store the value `9999999999999999` (because it will instead be stored as `1e+16`). In this example, Mathesar will store `1e+16`, displaying that value back to the user after entry.
+
+- **(Goal D)** Mathesar displays numbers and accepts entry of numbers in the user's desired locale, including the correct decimal separator and grouping separators.
+
+- **(Goal E)** Mathesar auto-formats numbers as the user types.
+
+- **(Goal F)** Mathesar accepts entry of numbers in scientific notation and displays numbers in scientific notation when the user is likely to prefer it. For example `1E100` instead of 1 followed by 100 zeros.
+
+## Implementation strategies
+
+### API
+
+- The API transmits numbers as strings because most JSON parsers are not capable of accurately parsing high precision numbers or special Postgres numbers stored plainly in JSON. The format for such strings is described below.
+
+### Canonical stringified number format
+
+Throughout our stack, we may want to represent numbers as strings. When doing so, we follow a format very similar to numbers in JSON.
+
+- The decimal separator is an ASCII dot.
+- The minus sign is an ASCII hyphen.
+- The exponent separator is an uppercase "E".
+- Grouping separators are disallowed.
+- The strings `"NaN"`, `"-Infinity"`, and `"Infinity"` are used to represent special Postgres numbers. Note that this is one way in which our stringified number format differs from JSON -- raw values like `NaN` are _not_ valid inside JSON.
+- The string `"NULL"` should _not_ be used. In JSON, the _value_ `null` should be used instead.
+- Examples:
+
+    - `"NaN"`
+    - `"-Infinity"`
+    - `"Infinity"`
+    - `"0"`
+    - `"1234.5"`
+    - `"99999999999999999999999999999999"`
+    - `"-1.2E-15"`
+
+
+### Locale
+
+- To satisfy Goal D, we infer the user's desired locale from their _browser_. The back-end is not aware of the user's locale unless the front end passes data about its in the API request.
+
+- Mathesar is limited to only displaying and accepting entry for numbers using the latin numbering system. This means sacrificing Goal A for the locales  "bn", "fa", and "mr" (Bengali, Persian, Marathi) which use non-latin numbering systems by default.
+
+- The minus sign is always displayed using an ASCII hyphen, regardless of locale (even though some locales like 'li' use the Unicode minus sign). The Unicode minus sign (and similar characters) will be accepted in all locales while _entering_ numbers though.
+
+- For display of scientific notation, the exponent separator is always displayed using the ASCII "E" regardless of locale even though some locales use other exponent separators. For input, other strings are accepted too, such as "e" (lowercase) and "Е" (U+0415) and variations of "x10^".
+
+### Parsing
+
+- The front end parses **user input**, by performing string transformations to normalize the input into canonical stringified form. For example, all characters matching the current locale's decimal separator are converted to ASCII dots.
+
+- The front end parses **canonical stringified numbers** (as produced above, or received from the API) by attempting to produce a `number` or `BigInt` which can be re-formatted exactly to match the input (without loss of precision).
+
+- The front end will **fail** to parse canonical stringified numbers like `"99999999999999.99"` (which cannot be accurately represented by any of JavaScript's built-in data types) because they will fail to be reformatted without loss of precision.
+
+### Formatting
+
+- The front-end formats numbers for display in the browser's locale.
+
+- Values from the API are formatted (to satisfy Goal D), and values from the user are formatted as the user types (to satisfy Goal E). In both of these cases we begin with a string.
+
+- When the front end can parse the input into a `number` or `BigInt`, it uses `Intl.NumberFormat` for formatting, which allows robust handling of grouping separators across various locales.
+
+- When the front end cannot parse the canonical stringified number, it falls back to a simplistic formatter which only performs string transformation to ensure only that the decimal separator matches the user's locale. This produces a formatted result which lacks grouping separators, partially sacrificing Goal D.
+
+### UX
+
+- The front-end allows users to enter the Postgres numbers `NaN`, `-Infinity`, and `Infinity` using a context menu -- and it displays those numbers stylized (similar to `NULL`).
+
+- While the user is entering a number, Mathesar _partially_ reformats the user's entry in order to match the final display of the number as closely as possible (to satisfy Goal E). For example, with a "en-US" locale, the entry `1234` is reformatted to `1,234` as the user types. However, some entries cannot be _fully_ reformatted as the user types. For example, an entry of `1.` would be fully formatted to `1` but must preserve the trailing decimal so that the user may type additional characters.
+
+    After the user shifts focus away from the number input, Mathesar _fully_ reformats their entry. For example, `1.` will be reformatted to `1` (removing the trailing zero).
+
+### Scientific notation
+
+- To satisfy Goal F, Mathesar delegates to the front-end the responsibility to decide between displaying a number in standard notation or scientific notation.
+
+    The algorithm for making this decision is as follows.
+    
+    - If the front-end cannot parse the number, then it retains the notation used in the stringified number.
+    - If the front-end can only produce a `BigInt`, then it will use standard notation because: (a) `Intl.NumberFormat` is not capable of accurately formatting very large `BigInt` values due to its upper limit on `maximumFractionDigits` of `20`, and (b) standard notation makes it clearer to the user that the value is an integer.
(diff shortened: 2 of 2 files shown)


engineering/architecture/numbers.md
```diff
@@ -0,0 +1,106 @@
...
+    - If the front-end can produce a `number`, then it will use some yet-to-be-determined heuristics to make a best guess as to whether the user would prefer to see the number in standard notation or scientific notation.
+
+### Loss of precision
+
+- To satisfy Goal C, Mathesar delegates responsibility to the back-end for transforming precise values entered by the user into less-precise values stored by Postgres. The front-end does not attempt any rounding or truncation. The back-end responds with the value saved in Postgres and the front-end displays this value. At this point, it is the responsibility of the user to notice that their value has been rounded or truncated. (This behavior is consistent with `psql`, spreadsheets, and AirTable.)
+
+## Future consideration
+
+- At some point we may add display options for numbers, which allow users to customize things like: percentages, minimum fraction digits, maximum fraction digits, parentheses for negative numbers, forcing scientific notation on/off, forcing grouping separators on/off.
```
pavish, line 106: @seancolsen We already have some of these as part of our goals for the alpha. Here's the [figma link](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?page-id=4260%3A37440&node-id=4308%3A41575&viewport=324%2C48%2C0.35&scaling=min-zoom&starting-point-node-id=4270%3A40031&show-proto-sidebar=1), in which you could find display options within Number.

The display options currently only includes percentages and ability to choose the format from a list of locales. We will not be defaulting to the client's locale all the time. If the client's locale is present in the list we allow, we'd choose it, if not, we'd choose en-us as the default. We would also not have to worry about non-latin numbering systems since we do not intend to support them.

@kgodey Let me know if my assumption is correct on this.

engineering/architecture/numbers.md
```diff
@@ -0,0 +1,106 @@
...
+
+- For display of scientific notation, the exponent separator is always displayed using the ASCII "E" regardless of locale even though some locales use other exponent separators. For input, other strings are accepted too, such as "e" (lowercase) and "Е" (U+0415) and variations of "x10^".
+
+### Parsing
+
+- The front end parses **user input**, by performing string transformations to normalize the input into canonical stringified form. For example, all characters matching the current locale's decimal separator are converted to ASCII dots.
+
+- The front end parses **canonical stringified numbers** (as produced above, or received from the API) by attempting to produce a `number` or `BigInt` which can be re-formatted exactly to match the input (without loss of precision).
+
+- The front end will **fail** to parse canonical stringified numbers like `"99999999999999.99"` (which cannot be accurately represented by any of JavaScript's built-in data types) because they will fail to be reformatted without loss of precision.
```
pavish, line 70: Is there a reason why we need to parse it to a number type on the frontend, after we normalize the string? What advantage does having a number type provide us, compared to directly using the normalized string? 

I'm only currently considering the usecase of entering numbers on the table cell, and not regular usage of the component.

seancolsen: Right -- yeah good question. I considered the idea of keeping numbers only as strings on the front end. Broadly there are two reasons:

1. Formatting a cell's numerical value for display. There are a variety of smaller goals here.

    - Likely the most important goal is to format the decimal separator in accordance with the user's locale (dot or comma). That's easy to do with strings, so if that's all we care about, then yeah, we can stick with strings.
    - Adding grouping separators is _nice_, but not crucial. Anecdotally, at my previous job I pushed a feature which added grouping separators as users typed into a field where they were frequently entering larger numbers -- and we got a lot of positive feedback about it immediately. Even if the separators are applied after the user finishes typing, I think they offer a significant UX improvement in some situations. If we stick exclusively to strings, it's actually quite hard to add the grouping separators in a locale-aware way. The `Intl` API takes care of it for us, but only if we have a `number` or `BigInt` -- so that requires parsing.
    - Similar to above, `Intl` offers a bunch of other [formatting options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) that we'll likely want to utilize at some point in the future, like `style: 'percent'`, `style: 'currency'`, `minimumFractionDigits`, and `maximumFractionDigits`. That stuff is trivial to implement when using `Intl`, but harder to do with raw strings.

1. Accepting a user-entered number in other contexts.

    This is the "regular usage of the component", as you say. This concern is not actually a pressing need for the features we're currently building, but more of a concern about building a robust and intuitive component library (which lowers the priority here, but still warrants consideration as we make these low-level architectural decisions). For editing a cell, I doubt we'll ever need access to the _numerical value_ of the cell on the front end (beyond the need to format the number for display). But we likely will need access to numerical values of user input on the front end in the future. For example, when/if we allow the user to configure the minimum number of fraction digits, I'd want to take that user input and feed it back into `Intl.NumberFormat` to display a formatted preview of a sample number. I think we'd also want client-side validation to ensure that `minimumFractionDigits` is within `0` and `20`, as required by `Intl.NumberFormat`.

Parsing the number is more work, but not _that_ much more, especially given the code I already have in-place in #1155. To finish out that PR, my plan is as follows:

- Move the majority of the code in the `NumberFormatter` class into a class called `UniversalNumberFormatter` which, instead of implementing `InputFormatter<number>`, implements `InputFormatter<UniversalNumberParseValue>` where `UniversalNumberParseValue` is as follows:

    ```ts
    export type UniversalParseResultType = 'number' | 'bigFloat' | 'bigInt';
    interface BaseParseValue {
      /**
      * A canonical string representation of the number, like you'd see in a JSON
      * string. No grouping separators. ASCII only. Dot for decimal separator.
      */
      string: string;
    }
    export interface SimpleNumberParseValue extends BaseParseValue {
      type: 'number';
      number: number;
    }
    export interface BigIntNumberParseValue extends BaseParseValue {
      type: 'bigInt';
      bigInt: bigint;
    }
    export interface BigFloatParseValue extends BaseParseValue {
      type: 'bigFloat';
    }
    export type UniversalNumberParseValue =
      | SimpleNumberParseValue
      | BigIntNumberParseValue
      | BigFloatParseValue;
    ```

- Then, by composing `UniversalNumberFormatter`, I'll build a very simple `NumberFormatter` class which implements `InputFormatter<number>` and a very simple `StringifiedNumberFormatter` which implements `InputFormatter<string>`. I'll use those classes to build very simple components `NumberInput.svelte` and `StringNumberInput.svelte`. For cell editing, we'd use `StringNumberInput` which handles high-precision floats. For other use-cases in the future where the front end needs access to numerical values from user input, we'd use `NumberInput` which prevents the user from entering values which cannot be accurately represented by `number`. Does that make sense?

pavish: Okay, sounds good to me.

engineering/architecture/numbers.md
kgodey: We should continue with approach 1. Our plan is for the database layer to be locale-agnostic, and I think that's most in line with our goal of allowing databases created using Mathesar to be used by other clients and tools.

> If the back end is aware of the locale, then the back end can do the formatting and parsing work. 
>
> This is the approach suggested by the Figma specs. The back end knows locale already from a setting persisted at the column level.

Implementing the locale display option for numbers does _not_ mean the backend takes responsibility for formatting and parsing work. Display options are meant to be a cache for the frontend to store preferences, none of the display options (for any data type) affect how the backend works. This is why we have a distinction between "database options" (affects the backend) and "display options" (frontend only).

Implementing the display option would allow the locale used for formatting/parsing to be changed on a per-column basis, but it would still be the frontend's responsibility to use the locale to handle formatting/parsing.

@seancolsen I know when we talked last week, I said we wouldn't be customizing locale per column, I had forgotten about this display option. I don't see how it would change implementation, though. Instead of always defaulting to the browser locale, we would check the locale in the display option and use the browser locale if there's no locale set there.

---

I'm assuming we are in agreement about not supporting non-Latin numbering systems and supporting display options for percentages.

engineering/architecture/numbers.md
kgodey: @seancolsen I think your comment here would be a good addition to the spec for future reference in case someone has the same question as @pavish.

engineering/architecture/numbers.md
pavish: @seancolsen 

>  That decision underpins so many other decisions and has the potential to render my entire previous week's worth of work as moot.

I do not see how this is the case. Am I missing something?

As @kgodey mentioned, we would just use a locale stored and sent from the backend and if it's not present we'd default to the browser locale. Additionally, we would also have a list of supported locales, so if the browser locale is not part of that list, we would choose en-us.

The frontend would always be incharge of formatting numbers for display here, and we would interact with the backend only using the canonical format.

seancolsen: @kgodey said:

> Display options are meant to be a cache for the frontend to store preferences

I see. Got it. This is sort of a separate hybrid approach, call it `4.`, which builds upon `1.`. Sounds good. 

> I don't see how it would change implementation, though.

Agreed. We can use all the work we've done thus far.

seancolsen: @pavish 

> Am I missing something?

I think we're all in agreement on sticking with `1.`, so I don't think it's worth hashing out approaches `2.` or `3.` in great detail. But just to offer some clarity...

The concerns surrounding imports seem to me to weigh in favor of `2.` or `3.` which is why I'm somewhat ambivalent overall and why I've raised this for discussion before continuing my work. If we _did_ decide to go with `2.` or `3.` (parsing and formatting on the back-end), we'd have much less of a need to do parsing and formatting on the front end. If we abandon the goal of formatting numbers _as the user types_, then the front end really could deal with the numerical cells entirely as strings. If that were the case, then I'd probably want to throw out a lot of the code I wrote last week, or maybe port it to Python. That's what I mean by "underpins so many other decisions".

engineering/architecture.md
```diff
@@ -30,6 +30,7 @@ These documents are notes about a specific feature written to get feedback befor
 
 - [Import Previewing](/engineering/architecture/import-preview-api)
 - [Syncing Database Objects](/engineering/architecture/syncing)
+- [Numbers](/engineering/architecture/numbers)
```
kgodey, line 33: Could we name this the same thing as the page ("Number display and entry") for clarity?

engineering/architecture/numbers.md
```diff
@@ -0,0 +1,106 @@
...
+
+- **(Goal C)** When a user enters data which cannot be accurately stored in the column exactly as-entered, then Mathesar stores the nearest acceptable value and clearly communicates to stored value to the user.
+
+    For example, a `float8` column cannot accurately store the value `9999999999999999` (because it will instead be stored as `1e+16`). In this example, Mathesar will store `1e+16`, displaying that value back to the user after entry.
+
+- **(Goal D)** Mathesar displays numbers and accepts entry of numbers in the user's desired locale, including the correct decimal separator and grouping separators.
+
+- **(Goal E)** Mathesar auto-formats numbers as the user types.
+
+- **(Goal F)** Mathesar accepts entry of numbers in scientific notation and displays numbers in scientific notation when the user is likely to prefer it. For example `1E100` instead of 1 followed by 100 zeros.
```
kgodey, line 25: This is engineering notation, not scientific notation ([quick link explaining the difference in case anyone else is interested](https://www.medcalc.org/manual/scientific-notation.php)).

engineering/architecture/numbers.md
```diff
@@ -0,0 +1,106 @@
...
+    - `"Infinity"`
+    - `"0"`
+    - `"1234.5"`
+    - `"99999999999999999999999999999999"`
+    - `"-1.2E-15"`
+
+
+### Locale
+
+- To satisfy Goal D, we infer the user's desired locale from their _browser_. The back-end is not aware of the user's locale unless the front end passes data about its in the API request.
```
kgodey, line 56: I assume you'll update this to include display options.

kgodey (approved):
Looks good overall, added some comments to be addressed before merge.

engineering/architecture/numbers.md
seancolsen: Is it? My understanding is that `1E100` is scientific notation whereas the same number would be represented with `10E99` in engineering notation (to make the exponent a multiple of three). Though realistically, nobody would want to use engineering notation for a number that large because the [SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes) (the main reason to use engineering notation, in my experience) stop well before 99.

seancolsen: Another way to frame my understanding... I'd say that engineering notation is a form of scientific notation, but not necessarily the other way around.

seancolsen: Also, here is how the `Intl` API behaves:

```txt
> Intl.NumberFormat('en-US', { notation: 'scientific' }).format(1E100)
"1E100"
> Intl.NumberFormat('en-US', { notation: 'engineering' }).format(1E100)
"10E99"
>
```

kgodey: Never mind, I got `E` notation and Engineering notation mixed up (because Engineering starts with an E).

i do think it would be cool to eventually support rendering non-E-scientific notation, e.g. 35.2 × 10<sup>-7</sup>. Probably not anytime soon, though, it sounds complicated.

seancolsen: Yeah we could eventually have a display option for rich formatting. Though (to be pedantic), I don't think we'd ever format the number in your example using an exponent of `-7` -- instead it would be be 3.52 × 10<sup>-6</sup> (in both scientific and engineering notation).

We may even want a display option to use engineering notation at some point.

For `notation`, we could have:

- `'auto'` (default)
- `'standard'`
- `'scientific'`
- `'scientific-rich'`
- `'engineering-rich'`

With the code I have so far, this would all be easy to support.

seancolsen: Interestingly, for some locales `Intl` actually behave more akin to the format you were expecting. For example:

```ts
Intl.NumberFormat('sv', { notation: 'scientific' }).format(1E100) // --> "1×10^100"
```

That's why I wrote in the specs that the exponent separator is to always be "E", regardless of locale.

engineering/architecture/numbers.md
seancolsen: @kgodey 

> your comment here would be a good addition to the spec

Fixed in 8d79053876b7deb1b86951f93e66414db099686d

engineering/architecture.md
seancolsen: Fixed in d0a1ef4fcb7ae6b2690d80ebdf2631aadb58e008

engineering/architecture/numbers.md
seancolsen: Fixed in 86b2315ec28d088372c8c0e42bc492f8745b2cd8

engineering/architecture/numbers.md
kgodey: > Though (to be pedantic), I don't think we'd ever format the number in your example using an exponent of `-7`

I didn't even notice that (I stole the example from the MedCalc docs I linked above), I agree the exponent should be `-6`.

I'm glad the current implementation will be extensible for all sorts of formatting, I like your suggestion notation options.

seancolsen:
@kgodey want to give this another look and merge if you're satisfied?

kgodey (approved):
Looks good!

PR #42, opened 2022-03-28
Data explorer specs

ghislaineguerin:
Fixes [#1065](https://github.com/centerofci/mathesar/issues/1065)


design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Context
+
+This document specifies the design for a visual query builder named `Data Explorer` within the Mathesar application. Through using `Data Explorer`, users will be able to analyze data across one or multiple tables of a schema and save the result of those queries as views.
+
+Additionally, users will be able to open existing views in `Data Explorer`, as long as the query commands and functions used to generate it are supported.
+
+For a more detailed overview of this feature, read [Views Product Spec](https://wiki.mathesar.org/en/product/specs/2022-01-views).
+
+## Scenarios
```
kgodey, line 19: We are missing scenarios for navigating to and navigating away from the data explorer (from the usual view with tabs and left sidebar).

kgodey (changes requested):
@ghislaineguerin I added some comments. Some general questions?

(1) What happened to the alerts and error prevention and onboarding sections?
(2) What is the plan for putting this in the design prototype? Will you still be doing that or are we going to stick with the low definition wireframes for now? There are a lot of errors and interactions that are missing here but I think if we're moving this to the prototype, those are better handled there. If there's no plans to put this in the prototype, then I'll start listing them out here.

> Fixes #1065

Could you link the PR description to the correct issue (it's in the main repo, not this one).

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,202 @@
...
+Selecting the base table will enable the interface controls for additional steps (e.g., column selection, filters), which were initially disabled. No action is available before a base table is selected.
+
+Once a selection has been made, the user will be prompted to proceed with the input columns configuration, starting with column selection.
+
+---
+Wireframes
+
+[Selecting a Base Table](https://share.balsamiq.com/c/esUftomoFsZiRhZDMYdoPP.png)
+
+[Base Table Selected](https://share.balsamiq.com/c/me5CfGGeX3R35x2otLqbpu.png)
```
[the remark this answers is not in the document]
ghislaineguerin: @kgodey Those numbers applied to the base table. I might add that information to the table selector menu instead. I feel it's important as an additional reference point when selecting tables, as the name only might not suffice.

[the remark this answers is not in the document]
ghislaineguerin: I've updated the wireframe to include those numbers inside the table selector.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,202 @@
...
+[Adding Columns](https://share.balsamiq.com/c/oyxTXxqSh8rLqU3JY71DWY.png)
+
+[Added Column](https://share.balsamiq.com/c/7U9yahZtYWX2G5obkyivoz.png)
+
+### 2.2 Adding Formulas
+
+Formulas are used to generate new columns based on different parameters. To access the list of formulas, the user start the `Add Column` process and selects the option `From Formula` at the top of the inspector panel. Selecting a formula will open a form that users can fill out to determine the values of the new column.
+
+Depending on the selected formula, different settings will be available.
+More on formulas and specific details for each will be covered on a separate issue.
```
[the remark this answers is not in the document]
ghislaineguerin: Done https://github.com/centerofci/mathesar/issues/1252

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,202 @@
...
+### 2.5 Applying a Formula to an Input Column
+
+Direct input columns can also be transformed into formulas. The available formulas will depend on the input column type.
+
+To transform the input column into a formula, the user will select the `Apply Formula` option in the column header menu.
+
+---
+Wireframes
+
+[Applying a Formula](https://share.balsamiq.com/c/nEzYaNKNTSv9EFwzwtfSof.png)
```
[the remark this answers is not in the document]
ghislaineguerin: Once they save it, yes. I have added the additional step.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,202 @@
...
+Links are presented as expandable sections containing the columns from the linked tables.
+
+When adding input columns, the user can add them using the list controls or drop the columns directly into the result table.
+
+---
+Wireframes
+
+[Adding Columns](https://share.balsamiq.com/c/oyxTXxqSh8rLqU3JY71DWY.png)
+
+[Added Column](https://share.balsamiq.com/c/7U9yahZtYWX2G5obkyivoz.png)
```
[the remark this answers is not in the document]
ghislaineguerin: I've added a section specific to sources.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,202 @@
...
+Direct input columns can also be transformed into formulas. The available formulas will depend on the input column type.
+
+To transform the input column into a formula, the user will select the `Apply Formula` option in the column header menu.
+
+---
+Wireframes
+
+[Applying a Formula](https://share.balsamiq.com/c/nEzYaNKNTSv9EFwzwtfSof.png)
+
+## 3. Transforming the Output Table
```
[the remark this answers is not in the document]
ghislaineguerin: Added to the specs along w wireframes.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+---
+Wireframes
+
+[Output Summarization](https://share.balsamiq.com/c/rPMwwETnuQ8a4ut8NfmcDB.png)
+
+[Summarization Options](https://share.balsamiq.com/c/x2iwVBT3NwzmKyhnxSiJku.png)
+
+### 3.3 Sorting the Output Table
+
+Users can sort the output table by applying a sort to any of the result table columns.
```
mathemancer, line 169: Does this include other aggregations? I'm not sure how this should work for non-numeric aggregations.

mathemancer (commented):
Great work overall. I left a few specific questions.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+Columns generated from formulas will display a formula icon indicator in the column header.
+
+---
+Wireframes
+
+[Added Formula](https://share.balsamiq.com/c/vWfJ9sUYWJxed5Zo5WtPGi.png)
+
+### 2.4 Filtering Input Column Values
+
+Columns that link to multiple records can have filters applied to them to retrieve only values that match user-specified criteria. Multiple filters are allowed for each input column.
```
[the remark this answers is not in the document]
kgodey: @mathemancer Could you provide an example of what you're envisioning here? I'm having a hard time understanding the issue from your description.

seancolsen:
## Caveats about my review

- For the sake of time, I have not read the other reviews. Apologies if some of this feedback overlaps with other comments.

- For the sake of time, I'm withholding critique of all the "Formulas" features because (based on conversations with Kriti) my understanding is that we're seeking to defer implementation of formulas until after alpha.

---

## General questions and critique

- `(1)` For our own product terminology, what's the difference between "Data Explorer" and "Query Builder"?

- `(2)` How do I navigate to the Data Explorer? What is the URL? From the Data Explorer, how do I navigate to view a table?

    I think the "Select Table" step should be its own page. When I'm selecting a table, I don't need to see any of the query-building UI. And when I'm building my query, I'd like to see a button somewhere like "Start over" or "New Query".

- `(3)` If I build a query, close my browser tab (without saving to a view), then return to the data explorer, is my query saved? Or do I need to start over?

    This may seem like an implementation detail, but I think we should figure it out during the spec process because there are some underlying architectural decision that affect UX and may require more UI.

    I see the following approaches:

    - "Ephemeral queries" - If I close the tab, I lose my work. The benefit to this approach is that it's simple to implement and easy for the user to build two queries side-by-side in separate tabs. The downside of course is that users may inadvertently lose their work.

    - "Server persistent queries via a custom data structure" - During every step of the process, we'd save the query to the server. This is a nice safeguard, but significantly complicates the use-case of running multiple queries side-by-side without saving them to views.

    - "Server persistent queries via views" - Automatically save to a view, but provide a mechanism mark the view as an incomplete query.

    I'm torn on which approach I think is best. There may be other approaches too.

- `(4)` A substantial amount of screen space is dedicated to the top header bar with "Exploring" and "Movie". Can we get rid of this header?

- `(5)` Why are limit and offset different from the pagination on tables?

- `(6)` Like Pavish mentioned on the call, I'm also interested in merging the left and right sidebars into one.

    As far as I can tell, this seems pretty straightforward...

    1. We'd have one sidebar. It could be on the left or right. I don't care.
    
    1. The sidebar would display one (and only one) "pane" at a time.

    1. We'd have three possible panes:

        - "Builder" - (with sections: "Columns", "Transform Result", "Query Details", and "Save Options")

        - "Source Selector" - shown titled "Add Columns" [here](https://share.balsamiq.com/c/oyxTXxqSh8rLqU3JY71DWY.png)

        - "Source Config"

        (The pane names would not be displayed to the user -- they only exist so that I can refer to the panes here.)

    1. The sidebar would show the "Builder" pane by default.

    1. When any non-default pane displays, a "Back" button would appear at the top, taking the user to the previously displayed pane.
    
    1. Clicking the "Add Column" button would show the "Source Selector" pane.

    1. Selecting a column would display the "Source Config" pane.
    
- `(7)` Can the user re-order the columns? How? Drag/drop? From the "Columns" section? From the result table? Both?

- `(8)` There are lots of text boxes that seem like they're probably editable but don't have a "Save" button. For example: filter values, "Limit", "Offset", column names, etc. How do I apply the changes I make within these text boxes? E.g. changing the name of a column.

- `(9)` Is the following query possible? If so, how do I do I add a filter as an OR condition across two columns?

    ```sql
    SELECT "Title"
    FROM "Movies"
    WHERE
      "sequel_id" IS NOT NULL OR
      "prequel_id" IS NOT NULL;
    ```

## 2.2 Input column sources

- `(10)` Within the "Input Column Source" [wireframe](https://share.balsamiq.com/c/ewCxvkjS8VoScKfgb9hFos.png), I find the "Link" info to be a bit confusing.

- `(11)` Here's an alternate idea for how to present the entire "Input column source" panel:

    image

    I'm not tied to this idea, but it's a quick stab at presenting the info in a manner more akin to my own mental model.

- `(12)` If we don't want to use my idea above, then I'd suggest that within the current wire frame we display "Table" before "Column" because table is a higher-level entity than column and I need to know the table first before I know the column.

## 2.4 Filtering Input Column Values

- `(13)` The highlighted "Filter" section in [this wireframe](https://share.balsamiq.com/c/tK2hZy6FB4zVYpN56ejpBk.png) took me a while to understand. I see it's purpose now, and see why we need it. But I think this needs a little polish before it will be intuitive to users. I like how we're using separate terms for "Aggregation" and "Summarize". We might want to think of different terms for the filter in this case and the filter that gets applied within "Transform Result". Perhaps a help bubble next to the section heading would be sufficient. I'm not sure.

- `(14)` Can I use the Data Explorer to build this query?

    ```sql
    SELECT
      "Person"."full_name" AS "James Bond Actor",
      array_agg("Movie"."Title") AS "Films",
    FROM "Person"
    JOIN "Movie Cast Member"
      ON "Movie Cast Member"."actor_id" = "Person"."id" AND
        "Movie Cast Member"."character_name" = 'James Bond'
    JOIN "Movie"
      ON "Movie"."id" = "Movie Cast Member"."movie_id"
    GROUP BY "Person"."id";
    ```

    What's notable here is that `"character_name" = 'James Bond'` is essentially a multi-record column filter, but then I'm aggregating on a different column that's based on that filtered join.


## 2.5 Aggregating Input Column Values

- `(15)` What are the options for the "Aggregation" dropdown? 

- `(16)` I'm curious how we intend to handle `JOIN` vs `LEFT JOIN`.

    _As an aside: my experience is primarily with MySQL. The MySQL syntax supports many different join keywords, but I've come to understand that when writing SQL, I only need to choose between `JOIN` and `LEFT JOIN` -- and words like `INNER`, `OUTER` and such are superfluous. I'm not certain the same is true for Postgres. And I'm aware that Postgres supports other joins which MySQL does not, such as `FULL JOIN`. So I hope this part makes sense to Postgres people. I can clarify if needed._

    If I don't want to aggregate my values, I may want to choose between a `JOIN` (which might eliminate rows from the base table) and a `LEFT JOIN` (which might have `NULL` values for the columns in my joined table). Maybe we should give users the choice between the two join types? I'm not sure the best way to present that choice to non-technical users though. It's a difficult concept.

    With an aggregation, I _think_ a `LEFT JOIN` would suffice in all cases, but there may be scenarios I'm not considering.

## Transform Result

- `(17)` Can we get wireframes for the UI of each transform?

- `(18)` For simplicity's sake, what if we only allow the user to delete the last transformation step? If this is not acceptable, then I have some further concerns which are more complex and will take more time to articulate.


## 6. Saving the Query as a view

- `(19)` After I save to a view, what happens? Does Mathesar keep me on the edit screen so that I can continue tweaking the query? Or does Mathesar navigate me to view the view?

- `(20)` If I use the query builder to edit a view, how do I update the view definition? Does every step within the query builder automatically re-save to the view? (I would hope not.) But some users might expect auto-save. How will users know their view definition is _not_ saved?


## An input column used in a formula has been deleted

- `(21)` The specs say:

    > If a user deletes an input column utilized in a formula, then that formula column needs to display the error and prompt the user to resolve it. 

    This is tricky. For the sake of simplicity, I'd rather not support this kind of situation that gracefully. I have some additional questions which are dependent on answers to other questions I've already asked. Fully explaining all my thoughts here will take a while, so **I'd like discuss this on our next call about views**.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+Filter options will be determined by the data type of the input column.
+
+Filters can be added by clicking on the `Add Filter` option from the query configuration panel or directly from the results table column header menu.
+
+Columns with filters applied will display a filter icon indicator in the column header.
+
+---
+Wireframes
+
+[Added Input Filter](https://share.balsamiq.com/c/tK2hZy6FB4zVYpN56ejpBk.png)
```
pavish, line 117: The filter is applied on `actor.date_of_birth`, while the selected column is `actor.id`. If we add another column `actor.name`, would we show this applied filter in both columns?

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+### 2.6 Applying a Formula to an Input Column
+
+Direct input columns can also be transformed into formulas. The available formulas will depend on the input column type.
+
+To transform the input column into a formula, the user will select the `Apply Formula` option in the column header menu.
+
+---
+Wireframes
+
+[Applying a Formula](https://share.balsamiq.com/c/nEzYaNKNTSv9EFwzwtfSof.png)
```
pavish, line 139: I think it's best to have a single workflow:

* Either user adds a formula and chooses the columns (My choice), or
* User adds a column and chooses the formula..

A formula can modify the column entirely, and already applied filters and aggregations would no longer be applicable. A user adding a formula to an existing column might not understand that the filters & aggregations would get removed.

Overall, I think this additional flexibility would be troublesome for the user and for us to maintain in the long run.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+The result table refers to the resulting table from all input columns selected and added, including filters and aggregations. Transforming this output can be done by applying output filters, sorts, and summarizations to the result table.
+
+### 3.1. Filtering the Output Table
+
+Users can filter the output table by adding a filter step to any column from the result table. The column selector, in this scenario, will only allow users to select input columns rather than the whole column list from the column selection stage.
+
+---
+Wireframes
+
+[Output Filter](https://share.balsamiq.com/c/dAVYnp8VnG2r2HzkSZ3rgp.png)
```
pavish, line 152: I find using the text `Filter` for both input and output to be confusing. Can we rename this to be more descriptive or add additional help content in both places describing what they do?

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+To delete a result transformation step, the user can click on the `Delete` button present in each step item. Deleting a step means that the system will undo all transformations made through that step, and the result table will reflect the updated output.
+
+In circumstances where subsequent steps rely on columns resulting from a deleted step, the system will display error warnings for every failed step. The table output will only reflect the output generated by those stages without errors.
+
+---
+Wireframes
+
+[Deleting a Transformation Step](https://share.balsamiq.com/c/XeNAKGz1UnDMfscSTLDVp.png)
+
+## 5. Previewing the Query Results
```
pavish, line 198: I think the preview needs more screen space. It would be better if we can explore options to merge both panes into one.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+
+A preview, or query result table, should be visible while the user is in `Data Explorer`. The result table will change based on the different configurations, for example, if a user applies a filter, the system should refresh the table to show the output with the filter applied.
```
pavish, line 200: It would be better if this preview table is similar to normal tables that we display, with pagination options at the bottom.

I dont think `limit` and `offset` should be part of transformations.

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+[Column Sources](https://share.balsamiq.com/c/ewCxvkjS8VoScKfgb9hFos.png)
+
+### 2.3 Adding Formulas
+
+Formulas are used to generate new columns based on different parameters. To access the list of formulas, the user start the `Add Column` process and selects the option `From Formula` at the top of the inspector panel. Selecting a formula will open a form that users can fill out to determine the values of the new column.
+
+Depending on the selected formula, different settings will be available.
+More on formulas and specific details for each will be covered on a separate issue.
+
+Columns generated from formulas will display a formula icon indicator in the column header.
```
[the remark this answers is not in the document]
ghislaineguerin: Yes, if a formula parameter is a column, the sources are displayed.

design/specs/visual-query-builder.md
ghislaineguerin: For the time being, non-numeric aggregations can only be displayed as lists.

ghislaineguerin:
@seancolsen Thanks for the input. Here are my answers, I will also ask @kgodey to provide additional details where needed.


> `(1)` For our own product terminology, what's the difference between "Data Explorer" and "Query Builder"?

They are the same. We renamed the query builder to data explorer to make it more user-friendly and goal-oriented.

> `(2)` How do I navigate to the Data Explorer? What is the URL? From the Data Explorer, how do I navigate to view a table?

The data explorer will be accessible via the top navigation as well as the toolbar for views and tables. Views can be opened in data explorer to inspect or modify their settings, and tables can be used as the starting point for a new query.

> I think the "Select Table" step should be its own page. When I'm selecting a table, I don't need to see any of the query-building UI. And when I'm building my query, I'd like to see a button somewhere like "Start over" or "New Query".

I believe that this is a reasonable recommendation, given that changing the base table will result in the entire query being deleted. It will also be more clear that table selection is not something that can be modified.

>`(3)` If I build a query, close my browser tab (without saving to a view), then return to the data explorer, is my query saved? Or do I need to start over?

In my opinion, "ephemeral inquiries" make sense as a starting point. Rather than just saving views, we wanted users to be able to explore data at any time and use this as a way of assessing, and reviewing data in their tables. It is possible that having new objects created every time they explore will discourage them from using the tool in this manner.

> `(4)` A substantial amount of screen space is dedicated to the top header bar with "Exploring" and "Movie". Can we get rid of this header?

It's possible that we could get rid of it if we implement the changes you've recommended earlier. We'd still need a way to make it clear which table we were looking at.

> `(5)` Why are limit and offset different from the pagination on tables?

According to my understanding, pagination is merely a display option, but limit and offset impact the actual number of records in the table

> `(6)` Like Pavish mentioned on the call, I'm also interested in merging the left and right sidebars into one.

Although it's worth looking into and your suggestions make sense, some of the interactions in the inspector panel could lead to a more comfortable user experience. We are preparing an interactive prototype and will be testing different configurations.
    
> `(7)` Can the user re-order the columns? How? Drag/drop? From the "Columns" section? From the result table? Both?

They should be able to do so from both locations.

> `(8)` There are lots of text boxes that seem like they're probably editable but don't have a "Save" button. For example: filter values, "Limit", "Offset", column names, etc. How do I apply the changes I make within these text boxes? E.g. changing the name of a column.

When a user leaves the field, the update should be reflected. The assumption is that the result table is updated when the user changes the input columns.

> `(9)` Is the following query possible? If so, how do I do I add a filter as an OR condition across two columns?

    ```sql
    SELECT "Title"
    FROM "Movies"
    WHERE
      "sequel_id" IS NOT NULL OR
      "prequel_id" IS NOT NULL;
    ```
I think we only support AND conditions. @kgodey might have a better answer.

> `(10)` Within the "Input Column Source" [wireframe](https://share.balsamiq.com/c/ewCxvkjS8VoScKfgb9hFos.png), I find the "Link" info to be a bit confusing.

We are attempting to create a more natural approach to describing the links that does not rely on diagrams. `For example, "Displaying values from the `title` column in `Movies`, which is linked via the `recommendationId` field in `movie recommendations`". Definitely, I believe it can be made clearer.

> `(12)` If we don't want to use my idea above, then I'd suggest that within the current wire frame we display "Table" before "Column" because table is a higher-level entity than column and I need to know the table first before I know the column.

This makes sense. However, we started with a column to better identify the source of the new column values. Because we will be displaying sources in places where the selection is done by column rather than a table, the order made more sense. I'm not sure what the best approach is here; I'll think about it more and respond with a better answer.

> `(13)` The highlighted "Filter" section in [this wireframe]

Rather than calling it filter, we could call it `filter records` or `include only records that match this condition`.

> `(14)` Can I use the Data Explorer to build this query?

This, I believe, is possible. In the interactive prototype, I will include an example of this.

> `(15)` What are the options for the "Aggregation" dropdown? 

https://wiki.mathesar.org/en/product/specs/2022-01-views/04-formulas/4a-record-aggregations

- `(16)` I'm curious how we intend to handle `JOIN` vs `LEFT JOIN`.

@kgodey, I'll defer to you.

> `(17)` Can we get wireframes for the UI of each transform?

Will be included in the interactive prototype.

> `(18)` For simplicity's sake, what if we only allow the user to delete the last transformation step? If this is not acceptable, then I have some further concerns which are more complex and will take more time to articulate.

We talked about it, and while your approach makes sense, not all step deletions will result in errors. We'd have to assess how simple it would be for users to update the remaining steps that cause errors. Imagine if a user only wants to remove an initial filter and can't do it without losing all subsequent steps?

> `(19)` After I save to a view, what happens? Does Mathesar keep me on the edit screen so that I can continue tweaking the query? Or does Mathesar navigate me to view the view?

It should remain and disable the save button until changes have been made so that users know it has been saved. I believe we'll need to provide a way to get to the actual view.

> `(20)` If I use the query builder to edit a view, how do I update the view definition? Does every step within the query builder automatically re-save to the view? (I would hope not.) But some users might expect auto-save. How will users know their view definition is _not_ saved?

See the previous answer.

- `(21)` The specs say:

    > If a user deletes an input column utilized in a formula, then that formula column needs to display the error and prompt the user to resolve it. 

I fully agree that this is a tricky scenario. It would be a good idea to go over this further on our next call.

design/specs/visual-query-builder.md
ghislaineguerin: According to my understanding, pagination is merely a display option, but limit and offset have an effect on the number of records in the table. How do you see users distinguishing between the two?

design/specs/visual-query-builder.md
```diff
@@ -0,0 +1,236 @@
...
+The columns in `Column Selector` are listed in a hierarchical structure based on the links that exist in both directions between the base table and other tables. The icons indicate whether the linked records are a single or multiple record collection.
+
+Links are presented as expandable sections containing the columns from the linked tables.
+
+When adding input columns, the user can add them using the list controls or drop the columns directly into the result table.
+
+---
+Wireframes
+
+[Adding Columns](https://share.balsamiq.com/c/oyxTXxqSh8rLqU3JY71DWY.png)
```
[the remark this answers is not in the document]
ghislaineguerin: Yes, they can add a column again. They may, however, change the column names and even the data types of the columns that have been selected (through aggregations or summarization). The 'In use' section would allow users to quickly see which columns are being used in the view. Consider a user inspecting an existing view that they did not create and simply wanting to list all of the sources.

design/specs/visual-query-builder.md
ghislaineguerin: It would only affect the input column with the filter.

design/specs/visual-query-builder.md
ghislaineguerin: Agreed. See my reply to Sean's same question below:
> Rather than calling it filter, we could call it filter records or include only records that match this condition.

design/specs/visual-query-builder.md
kgodey: @ghislaineguerin As we discussed yesterday, I think we should remove formulas from this spec entirely and have a separate spec for integrating formulas once we finish the design for https://github.com/centerofci/mathesar/issues/1252.

design/specs/visual-query-builder.md
kgodey: I don't think merging both panes is a good idea. The left pane represents the structure of the query, the right pane is an "inspector" focusing on context-specific details for the current task the user is doing. 

I do agree that the user might want more space for the preview. Making both panes collapsible seems like a better option for this.

design/specs/visual-query-builder.md
kgodey: As @ghislaineguerin said, limit and offset are applied to the query definition, this is not pagination. @ghislaineguerin It would be good to explain the limit/offset somehow since new users may not understand the difference. I also think the wireframes (or prototype) should include pagination.

kgodey:
Responses to comments from @seancolsen and @ghislaineguerin:

> I think the "Select Table" step should be its own page. When I'm selecting a table, I don't need to see any of the query-building UI. And when I'm building my query, I'd like to see a button somewhere like "Start over" or "New Query".

This was integrated into the main query builder page so that if a new user clicked on "Data Explorer", they could have an idea about what was going to happen next. If we just ask them to select a table in a separate page (we had this in a previous wireframe version), they won't know what the context is.

@ghislaineguerin It would be good to integrate reasoning like this into the spec so that if we go back and look at the spec, we'll remember why we made the design decisions we did.

> - `(4)` A substantial amount of screen space is dedicated to the top header bar with "Exploring" and "Movie". Can we get rid of this header?

We need to have the base table in a prominent place because the rest of the query depends on it. The columns available are based on relationships to the base table, the source information displayed is relative to the base table. I think it helps orient users to have it displayed _over_ everything else. I do think we could put in an explanation there about why it's relevant.

We could also use the header for other purposes (e.g. if we're using the data explorer to edit an existing view, we could put information about the view there). 

@ghislaineguerin It would be good to document that the main purpose of the header to orient the user and provide context for the whole query.

> `(6)` Like Pavish mentioned on the call, I'm also interested in merging the left and right sidebars into one.

Please see my response to @pavish's comment here: https://github.com/centerofci/mathesar-wiki/pull/42#discussion_r843378111 I think that making the panes collapsible is a better option than trying to merge them because having a separation of "where am I in the query" and "what am I currently working on" would help orient users.

> `(9)` Is the following query possible? If so, how do I do I add a filter as an OR condition across two columns?

@ghislaineguerin I do think we should support `OR` filters similarly like we do for tables. We should have a consistent filter component everywhere we use filters so that the user knows what to expect when they see the word "filter".

> `(16)` I'm curious how we intend to handle `JOIN` vs `LEFT JOIN`.
>
>  If I don't want to aggregate my values, I may want to choose between a `JOIN` (which might eliminate rows from the base table) and a `LEFT JOIN` (which might have `NULL` values for the columns in my joined table). Maybe we should give users the choice between the two join types? 

We're not currently giving users this option, we're defaulting to `LEFT JOIN`. I think this is fine for the alpha release. If there are queries that users want to make that are not served by this version, we'll figure that out later.

> - `(19)` After I save to a view, what happens? Does Mathesar keep me on the edit screen so that I can continue tweaking the query? Or does Mathesar navigate me to view the view?
> 
> - `(20)` If I use the query builder to edit a view, how do I update the view definition? Does every step within the query builder automatically re-save to the view? (I would hope not.) But some users might expect auto-save. How will users know their view definition is _not_ saved?

These are out of scope for this spec, they will be handled in:
- https://github.com/centerofci/mathesar/issues/781
- https://github.com/centerofci/mathesar/issues/783

> Further, the user may actually want to page through the Data Explorer results, so perhaps we add "limit" and "offset" transform steps _and_ add pagination to the Data Explorer results (which does not affect the saved view).

I think this is a good idea. I also commented about this here: https://github.com/centerofci/mathesar-wiki/pull/42#discussion_r843378682

> Most operations within the Data Explorer will need "save" buttons, so using them everywhere provides a more consistent UX, which I think is important with a complex tool like this.

I agree with this.

> What if we only allow the user to delete steps that don't result in errors? If the step will cause errors, then the delete button would be disabled.

@ghislaineguerin I think it would be fine to only allow users to delete the last step. I don't think we can figure out which steps will cause errors and which won't, since the query is built in-order. Even if we can figure it out, it will be a hard problem to solve technically and I don't think it's worth putting effort into for the alpha release.

The other alternative I see is that deleting a step would also delete all subsequent steps, but that seems like a worse UX.

design/specs/visual-query-builder.md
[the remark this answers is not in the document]
mathemancer: Suppose we're aggregating actors in movies into lists, but we're only interested in actors whose first name starts with 'P'. Then filtering might produce some movies with _no_ actors after filtering for only actors fitting the description. I assume we'd want to show `NULL` for that case. What if instead, we're aggregating the number of Academy awards held by the casts of movies? This would be a sum, rather than a list. But then the value shown for a movie with a cast none of whom ever won an Oscar should probably be `0`, not `NULL`. Is this the intended UI in these cases? Are there other cases  to consider?

Now, since we're opening the possibility of ending up with _no_ results for some movies, what if we have some many-to-one relationship (maybe movie studios as a toy example)? Each movie has one studio, and for some reason we're only interested in studios starting with the letter 'P'. Why not give the possibility of filtering the linked studio column so that we show 'Pixar' and 'Paramount', but only `NULL` where there would have been 'Universal'? My question there is whether the restriction to only allowing filtering of linked records in cases of columns that link multiple records is needed.

design/specs/visual-query-builder.md
mathemancer: My question in that case is whether we want to sort by lists. That may be an awkward concept for users, since sorting by a list isn't a very natural thing to do, and there are a number of slightly weird options for how the sorting would actually work.

kgodey: Maybe we don't allow users to sort by aggregated columns? I can see how that would get confusing.

@ghislaineguerin what do you think?

design/specs/visual-query-builder.md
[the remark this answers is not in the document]
kgodey: I understand your question now. I'm concerned that allowing filtering for those types of use cases would overload the concept of filters. I think users should use formulas for use cases like that instead, e.g. [the `Is True` formula](https://wiki.mathesar.org/en/product/specs/2022-01-views/04-formulas/4e-boolean-formulas#is-true).

PR #44, opened 2022-04-18
List data type specs

ppii775:


design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+1. is empty
+2. is not empty
+3. contains `<LIST ITEM DATA TYPE>`
+4. number of items greater than `<NUMBER>`
+5. number of items greater than or equal to `<NUMBER>`
+6. number of items equal to `<NUMBER>`
+7. number of items lesser than `<NUMBER>`
+8. number of items lesser than or equal to `<NUMBER>`
+
+Users are able to utilize multiple filters through the `Add Filter` feature. By default, the relationship between filters is set to `difference`. Users can enable the `Union` toggle to include data cells that appear in any of the filters.
```
kgodey, line 58: We already have an "and"/"or" option in our current filters, can we use that instead of introducing a new concept for union and difference?

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+
+1. Column header data type icon.
+2. Data type icon and description in the column header menu. `Allow List` option is required to be turned on for List data type.
+3. Tooltip information when hovering on the column header data type icon. 
+4. Items in a single cell is separated by comma delimiters and a `show all` option is present. Users can use `show all` function to expand the List and view all List items.
+
+---
+Wireframes
+
+[Viewing Existing List Data](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?node-id=7317%3A80859&viewport=306%2C48%2C0.59&scaling=min-zoom&starting-point-node-id=7379%3A82626&show-proto-sidebar=1)
```
kgodey, line 41: I think it would be good to show list using pills even before the user clicks on "show all"

kgodey: I also don't think we need delimiters at all if we're showing list items as pills.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+---
+Wireframes
+
+[Filtering Existing List Data](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?node-id=7317%3A80859&viewport=306%2C48%2C0.19&scaling=min-zoom&starting-point-node-id=7379%3A82944&show-proto-sidebar=1)
+
+---
+
+### 1.3 Editing Existing List Data
+
+Users can single-click a List data cell or click the `show all` button to expand the selected the data cell and enter editing mode. A background shade is added to each List item, not the delimiters, to highlight the nature of List data type. Under cell editing mode, users can again single-click the item that they wish to edit to finish the edit without compromising the data structure.
```
kgodey, line 69: It might be good to make the list item that's currently being edited more visually prominent.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+
+[Editing Existing List Data](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?node-id=7317%3A80859&viewport=306%2C48%2C0.24&scaling=min-zoom&starting-point-node-id=7809%3A84002&show-proto-sidebar=1)
+
+---
+
+### 1.4 Editing Existing List Data – Item Type Warning
+
+While editing List data items, if Mathesar detected that users were entering a potentially non-conforming item type (e.g., entering a number into a List of Text items), a warning modal would show up to inform users. 
+
+If Mathesar couldn't conform the new item to the existing List item data type, a serious warning would show up before converting all List data item in the column to Text.
```
kgodey, line 84: I don't think we should convert data to Text. It would be better to not allow the edit to be saved.

Combining editing an individual cell and changing the data structure of the entire column during the same action will confuse the user, I think, and often lead to unexpected outcomes. Data type conversions affects how the data is stored in the database and I think we should only allow conversions when the user explicitly requests it.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+---
+Wireframes
+
+[Editing Existing List Data](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?node-id=7317%3A80859&viewport=306%2C48%2C0.24&scaling=min-zoom&starting-point-node-id=7809%3A84002&show-proto-sidebar=1)
+
+---
+
+### 1.4 Editing Existing List Data – Item Type Warning
+
+While editing List data items, if Mathesar detected that users were entering a potentially non-conforming item type (e.g., entering a number into a List of Text items), a warning modal would show up to inform users.
```
kgodey, line 82: It would be better to have a different example here, numbers are text too. For example, a list of movies would have a list item type of Text but could include `9`, [which is a movie](https://www.imdb.com/title/tt0472033/).

Adding text into a list of numbers would be a better example. We definitely know when something is not a number.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+## 2. Converting Existing List Data to a Different Type
+
+Users can perform numerous data type operations to List data. There are two main categories of conversion:
+
+1. Converting List data to a basic data type as a whole.
+2. Converting List items to a different basic data type.
+
+### 2.1. Converting List Data to a Basic Type by Turning Off `Allow List`
+
+By default, the `Allow List` toggle in the column header menu would be turned on. While users are editing data cells, if Mathesar detects a comma delimiter and a data structure that resembles a List, Mathesar can make automatically inference and convert the column into a List data type. Users can still maintain the same basic data type for each List item, and the common basic data type settings, such as formatting and setting up Null values, are still available for List items.
```
kgodey, line 104: A couple of points here:

- We cannot turn on the "Allow List" toggle by default, since it changes how the data is stored in PostgreSQL. The user will need to turn this on.
- We will not be performing automatic data type conversions based on the data the user inputs, see my previous comment for more on that.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+
+Users can convert the entire List back to a basic type by simply turning off the `Allow List` toggle. While the toggle is turned off, the column will not automatically and cannot be converted to a List data type. In this case, users' comma delimiter formatted data can only be recognized as Text by Mathesar.
```
kgodey, line 106: It would be good to integrate the "Allow List" functionality into the "Database Options" sections of our current data type menus.

Also, I'm not sure we should have it be a toggle since it changes the whole data type of the column. Could we make it more consistent with how we're allowing users to change data types?

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+---
+Wireframes
+
+[Converting List Items](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?node-id=7317%3A80859&viewport=306%2C48%2C0.26&scaling=min-zoom&starting-point-node-id=8268%3A91011&show-proto-sidebar=1)
+
+---
+
+## 3. Creating a New Column in List Data Type
+
+Since Mathesar aims to support all basic data types in List format as well as detect List data structure automatically, List data capability comes with every new column added. Users can opt out this feature by turning off `Allow List`.
```
kgodey, line 128: Given the above comments, this needs to be reworked.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+
+---
+
+### 4.3 Converting Partially Correctly Formatted Non-List to List
+
+If Mathesar uses comma delimiter to detect List, Mathesar can still successfully parse Text data type as long as a comma is present. In this case, if the resulting List items are inferred as an incorrect basic data type due to the data error (e.g., Mathesar is not successful in inferring Number due to a data error), users can quickly realize something is wrong. Unfortunately, users still need to locate the errors by themselves.
+
+## 5. Other Considerations
+
+The List data type design exploration above has assumed a fair amount of background processing capability in Mathesar. For example, the `Allow List` option can simplify users' decision-making process by detecting the existence of a List in the background constantly, as users are actively editing data.
```
kgodey, line 187: Although we _can_ do this, I don't think we should be changing the structure of data without an active request from users to do so. Reasoning:
- Changing the structure of data may result in data loss, which is not something we should risk without explicit instructions from the user.
- Changing the structure of data also constrains what kind of data can be stored in a column.  

Unlike a spreadsheet, Mathesar requires that all data in a column have the same data type, and we cannot deal with inconsistencies well. This is a benefit of Mathesar in many use cases – we can guarantee data integrity, but that means we can't be as flexible.

kgodey (changes requested):
@ppii775 This is a good start, thanks for your work on this so far!

I've added some comments, mainly focusing on technical and product constraints. I'm sure @ghislaineguerin will also have a review soon. Please let me know if you have any questions or anything is unclear.

kgodey:
@ppii775 It also looks you added the file in two different places. I only read the one in `design/specs`

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+
+---
+
+### 1.3 Editing Existing List Data
+
+Users can single-click a List data cell or click the `show all` button to expand the selected the data cell and enter editing mode. A background shade is added to each List item, not the delimiters, to highlight the nature of List data type. Under cell editing mode, users can again single-click the item that they wish to edit to finish the edit without compromising the data structure.
+
+To add a new List item in the cell, users can simply single-click the white space within the cell when under editing mode. A new item placeholder and a new delimiter will show up. A preset Null value will be entered automatically if users have allowed and configured it.
+
+---
```
ghislaineguerin, line 73: How do you envision editing for longer lists. Do you think we need a way to retrieve/find the list item. Did you consider having a list like a grid rather than inline list? What would make the items easier to find?

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+Wireframes
+
+[Converting List Items](https://www.figma.com/proto/Uaf1ntcldzK2U41Jhw6vS2/Mathesar-MVP?node-id=7317%3A80859&viewport=306%2C48%2C0.26&scaling=min-zoom&starting-point-node-id=8268%3A91011&show-proto-sidebar=1)
+
+---
+
+## 3. Creating a New Column in List Data Type
+
+Since Mathesar aims to support all basic data types in List format as well as detect List data structure automatically, List data capability comes with every new column added. Users can opt out this feature by turning off `Allow List`.
+
```
ghislaineguerin, line 129: I believe that in the majority of cases, users will not want lists for all columns because it will make it difficult to identify relationships between data items. I think list will be an exemption for a few cases, so I'm not sure we want to enable the option by default.

ghislaineguerin (commented):
You are making some good suggestions, though I believe some of them require further clarity. I've added some questions.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+Read [Design for List data type](https://github.com/centerofci/mathesar/issues/978) for a more in-depth explanation of this feature.
+
+Read [Design for visual query builder](https://github.com/centerofci/mathesar/issues/1065) for mapping table details.
+
+# Scenarios
+
+## 1. Working with Existing List Data
+
+Users will be able to view, filter, and edit existing List data. This scenario assumes no List item data type change (e.g., converting a Number item to a Text item) during the edit.
```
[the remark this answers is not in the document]
ppii775: Sorry for the confusion – I was referring to the entire user action, not during the edit action specifically. I will revise the wording.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+
+---
+
+### 1.2 Filtering Existing List Data
+
+The following filters are available for List data:
+
+1. is empty
+2. is not empty
+3. contains `<LIST ITEM DATA TYPE>`
```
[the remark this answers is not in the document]
ppii775: These filter requirements was from the Discussion – I believe this one (no.3) will filter cells based on the data types of List items.

design/specs/list-data-type.md
ppii775: Got it. I can remove this description. I don't think these features are shown in the prototype I'm using but I could have missed them.

design/specs/list-data-type.md
ppii775: I would advocate for clearly communicate the expected List structure to users to facilitate their editing progress. However, this is just my expectation as a single user and I'm ok with changing it to what makes sense to you.

design/specs/list-data-type.md
ppii775: A great point – I can explore some options.

design/specs/list-data-type.md
ppii775: Agreed. The feature explored here is to allow Mathesar to inform users if it thinks users could have made a mistake, and it could happen to all basic data types. Mathesar would know whether 9 is a text or a number, but it could ask users to double-check when non-conforming items are detected.

design/specs/list-data-type.md
ppii775: Good point. However, how users change between data types are not shown clearly in the prototype I'm using. The prototype does do a great job demonstrating the initial data type setup. I found a Loom video from Pavish explaining the implementation of data type change for a couple types. Could you point me to the current considerations for the basic data type changes? I included more detail in Section 5 Other Considerations.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+editor: markdown
+dateCreated: 2022-04-06 18:02:58
+---
+
+# Context
+
+This document specifies the design of List data type within the Mathesar application. List data type can include one or more items that are in the same basic data type (e.g., Number), separated by a comma delimiter if more than one item is present. Mathesar aims to support all basic data types in List format. Users will be able to view data that is currently in Mathesar List data type, edit and filter List data, create new columns in List data type, and convert existing data into List data type. 
+
+The goal of the design in general is to help non-technical understand List data type and maintain List data integrity. Aggregating data into Lists via mapping tables is not beyond the scope of this spec and should be designed in visually different ways to communicate the affordance.
+
```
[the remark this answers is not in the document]
ppii775: To me, the mapping tables and aggregated Lists for Views feel like a different concept, and it might be beneficial to distinguish it in significant ways, possibly even with a different name. 

It's quite possible I haven't understood the connection between a "native" List in a single cell vs. an aggregated List. We can discuss.

ppii775:
> @ppii775 It also looks you added the file in two different places. I only read the one in `design/specs`

Sorry about that – my deletion was apparently not successful but it's cleaned up now.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+1. Converting List data to a basic data type as a whole.
+2. Converting List items to a different basic data type.
+
+### 2.1. Converting List Data to a Basic Type by Turning Off `Allow List`
+
+By default, the `Allow List` toggle in the column header menu would be turned on. While users are editing data cells, if Mathesar detects a comma delimiter and a data structure that resembles a List, Mathesar can make automatically inference and convert the column into a List data type. Users can still maintain the same basic data type for each List item, and the common basic data type settings, such as formatting and setting up Null values, are still available for List items.
+
+Users can convert the entire List back to a basic type by simply turning off the `Allow List` toggle. While the toggle is turned off, the column will not automatically and cannot be converted to a List data type. In this case, users' comma delimiter formatted data can only be recognized as Text by Mathesar.
+
```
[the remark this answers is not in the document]
ppii775: From a non-List to List, Mathesar would infer the List item data type and Text would be the default if inference was not successful.

From List to non-List, this scenario does not consider splitting Lists (like the Tableau example we looked at), so only Text can accommodate the change due to the existence of spaces and commas in the Lists.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+### 1.1 Viewing Existing List Data
+
+Existing Mathesar List data distinguishes itself from other data type through four unique ways: 
+
+1. Column header data type icon.
+2. Data type icon and description in the column header menu. `Allow List` option is required to be turned on for List data type.
+3. Tooltip information when hovering on the column header data type icon. 
+4. Items in a single cell is separated by comma delimiters and a `show all` option is present. Users can use `show all` function to expand the List and view all List items.
+
```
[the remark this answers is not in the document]
ppii775: That is a great point. I will try some new ways other than toggle.

design/specs/list-data-type.md
```diff
@@ -0,0 +1,194 @@
...
+2. is not empty
+3. contains `<LIST ITEM DATA TYPE>`
+4. number of items greater than `<NUMBER>`
+5. number of items greater than or equal to `<NUMBER>`
+6. number of items equal to `<NUMBER>`
+7. number of items lesser than `<NUMBER>`
+8. number of items lesser than or equal to `<NUMBER>`
+
+Users are able to utilize multiple filters through the `Add Filter` feature. By default, the relationship between filters is set to `difference`. Users can enable the `Union` toggle to include data cells that appear in any of the filters.
+
```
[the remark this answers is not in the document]
ppii775: Got it.

design/specs/list-data-type.md
ppii775: This is definitely something we should consider. How is the retrieving/finding feature handled currently? For example, does Mathesar allow users to quickly find a specific cell? It might be possible to use a similar design for the List item search because essentially it is also finding a specific cell.

design/specs/list-data-type.md
ppii775: This is a great point. Turning on List option does not convert data to List if List structure is not detected. However, Kriti mentioned Mathesar didn't have the capability to check data in realtime – I need to reconsider this design.

ppii775:
@kgodey @ghislaineguerin 
Thank you both so much for the comments! I apologize if I still misunderstood some of the basic technical limitations.
I'm scheduled to discuss with Ghislaine this Friday and I'm happy to continue our discussions here async in the meantime.

design/specs/list-data-type.md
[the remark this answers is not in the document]
ghislaineguerin: When it comes to the view, the data contained within an aggregated column will be treated as a list, much like a list in a table. However, the editing interactions for each must be distinct. A view, aside from having a query behind it, should look like a table.

design/specs/list-data-type.md
ghislaineguerin: In this case, I believe we might also want tag-like items. Assume we're entering dates or boolean list items; if we had a control to add a new item, it would then add a tag that, when interacted with, would display a checkbox or a date picker. Also, if you're entering an email, the pill-like design could help orient the user when there are validation errors (email entered in incorrect format, missing @, etc.).

kgodey:
You're welcome @ppii775! I'll defer to @ghislaineguerin to take the lead on following up and letting me know if/when I should review again. For explaining the technical limitations, @mathemancer may also be helpful.

ghislaineguerin (commented):
Thanks again Han. I left some additional comments. Please ping me when you submit another round of changes.

design/specs/list-data-type.md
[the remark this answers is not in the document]
ppii775: Thanks for the clarification! These bullet points are in line with what I'm thinking. The editing flow described in 1.3 can accomplish all these except deleting multiple list items at once – I treated editing a cell with List similar to a regular cell, just with pronounced List structure indication to help users – would love to know your thought.

design/specs/list-data-type.md
ppii775: Thanks for the clarification. I'm not confident that tags without delimiters is the best way to help users edit the List. For example, if they'd like to add two independent items by simply typing, what would be their expectation in telling Mathesar where the two items separate? Would it be a space, a comma, or something else? On the other hand, aggregated lists with an underlying mapping table wouldn't have this issue, because they are not created by typing in a cell. The attached prototype is also tag-like, just with delimiters to indicate how users should edit the List, and the interactions you mentioned should have ways to work without problems.

ghislaineguerin: I see what you mean. So in this case, when you enter a comma, does a new list item appear? how will the user insert a list item in between two existing items? How are other data types inserted into the list?

PR #47, opened 2022-06-21
Add formula column spec to design spec

Dhruvi16:
Spec for [formula column](https://github.com/centerofci/mathesar/issues/1252)

This is a work in progress. 
This spec explains hows and what of adding and editing a formula column.


kgodey:
@Dhruvi16 I marked as a draft since the description says it's a WIP. Please mark as ready for review when you're done.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+date: 2022-06-20T08:24:04Z
+tags: 
+editor: markdown
+dateCreated: 2022-06-20T08:24:04Z
+---
+
+## Context
+
+In addition to data columns, we have formula columns as well. These columns are basically derived from a formula. A formula may or may not use another data column to generate the formula column. The formula can also use parameters of different data types other than columns. 
+
```
ghislaineguerin, line 14: Let us also clarify that the use of formulas is currently restricted to queries created with Data Explorer. Let's also include a link to the wiki section that goes into greater detail about formulas.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+## Scenarios
+
+### Scenario 1 - Add a formula column without clicking outside of the inspector.
+
+On the left hand side of the explorer view, there is a  `Add Column`  button. When user clicks on it, there are two options visible to them. One is `Add direct column` and the other is `Add formula column`. To add a formula column user will need to click on `Add formula column`. 
+
+![](/assets/design/specs/add-formula-column/1.png)
+
+Once the user clicks on `Add formula column` option, the inspector shows formula settings that include `Name` input of the formula and a list of formulas to select from. The list of formulas is described in detail in the component section.
```
ghislaineguerin, line 25: We need to specify that the formula column must be named. Also please link to the **formula selector** section from here.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+
+On the left hand side of the explorer view, there is a  `Add Column`  button. When user clicks on it, there are two options visible to them. One is `Add direct column` and the other is `Add formula column`. To add a formula column user will need to click on `Add formula column`. 
+
+![](/assets/design/specs/add-formula-column/1.png)
+
+Once the user clicks on `Add formula column` option, the inspector shows formula settings that include `Name` input of the formula and a list of formulas to select from. The list of formulas is described in detail in the component section. 
+
+If the name is left empty by user for the next steps, there is a default name given to the column that is ColumnX. X is the number given to the column. For instance if its the first default named column it will be Column1.
+
+![](/assets/design/specs/add-formula-column/2.png)
```
ghislaineguerin, line 29: Can we move all wireframes to the top of this document and add them as balsamiq links instead of embedding them? It will make maintaining the document a lot easier.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+The list type formula works on columns with list data type. There can be different parameters depending on what the formula is.
+
+![](/assets/design/specs/add-formula-column/12.png)
+
+### Scenario 2 - Adding an erroneous parameter to the formula 
+
+Parameters of different formulas accept columns of different data types and values of different data types. The erroneous parameter here means the data type of the entered value or column is incompatible with the formula. 
+
+For instance, the length formula expects a list value or a list type column as an input to the parameter `List` but it gets a text type column which is movieNames and hence throws an error and the user is not able to see a filled column.
+
```
ghislaineguerin, line 100: Instead of allowing users to select an invalid column, I believe we should prevent the error. It's worth noting that we limit the number of columns and formulas that can be chosen based on existing parameters.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+#### Scenario 4.c - Clicks outside after successfully adding the formula
+
+If the formula and its parameters are added successfully and the preview column is filled. When the user clicks outside the preview is still visible. The user will get the inspector view of formula back if they click on the column. 
+
+![](/assets/design/specs/add-formula-column/16.png)
+
+#### Scenario 4.d - Clicks outside after adding erroneous parameters
+
+If the users have added erroneous parameters then they click outside the inspector, the parameter values will be lost and they will see a similar empty column preview with a warning.
```
ghislaineguerin, line 141: I think the process should go as follows:
- The user adds a formula column, and the inspector displays the column properties (name) and the formula selector.
- The table displays a preview of the column with no data.
- The user selects a formula, and a list of formula parameters is displayed.
- The user enters formula parameters.
- Once they select a column, we can begin displaying the output in the preview.
- The user can deselect a formula column at any time, and this should not result in the column being removed.

I believe we must ensure that validation in the formula form prevents invalid values from being entered.

ghislaineguerin (changes requested):

ghislaineguerin:
1. Types of error user might run into

### A column used in a formula was deleted

We might consider still displaying the name of the column that was selected but indicate to the user that it has been deleted. They can choose a replacement or delete the entire column.
image

### A column used in a formula has  been changed to a different type

In this case the formula should be run and the output should be invalid. We need to define what the content should be.

image

### A column used in a formula belongs to a table that was deleted

This is similar to the deleted column.

2. Do the users have to add column again after changing parameters? What if they add erroneous params?

The user should be able to change parameters for an existing formula column without having to add a new one. We should think about input validation to prevent wrong parameters. 

3. Do we really need add column to query action?

Can you please clarify what you mean by this?

design/specs/add-formula-column.md
Dhruvi16: So just confirming again, users will only be allowed to create formula columns while exploring the query in the data explorer, right?

Will surely add the link.

design/specs/add-formula-column.md
Dhruvi16: Yeah for the linking to different sections part I was thinking of doing it once I am done writing all the sections. Does that sound okay?

design/specs/add-formula-column.md
Dhruvi16: Sure.

design/specs/add-formula-column.md
Dhruvi16: I agree with you limiting column options but the users can also add values to the parameters. 

Plus I don't think we should limit the formula list based on the parameters since it might restrict user in going to a completely different formula if they are not happy with the result. 

LMK what you think

Dhruvi16:
> ## Types of error user might run into
> ### A column used in a formula was deleted
> We might consider still displaying the name of the column that was selected but indicate to the user that it has been deleted. They can choose a replacement or delete the entire column. image
> 
> ### A column used in a formula has been changed to a different type
> In this case the formula should be run and the output should be invalid. We need to define what the content should be.
> 
> image
> 
> ### A column used in a formula belongs to a table that was deleted
> This is similar to the deleted column.
> 
> ## Do the users have to add column again after changing parameters? What if the
y add erroneous params?
> The user should be able to change parameters for an existing formula column without having to add a new one. We should think about input validation to prevent wrong parameters.
> 
> ## Do we really need add column to query action?
> Can you please clarify what you mean by this?

By `Add column` in both my last question, I mean the little action on the right column to add column to query. 
Untitled

Dhruvi16:
@ghislaineguerin I want you to shed some light on how the error messages would look like obviously it will be a detailed discussion and where and how to show them but just putting it here so we do not miss it. Thanks 💛

Dhruvi16:
For the deletion of the column added to the input parameter of formula, we can either delete the formula column which is logical because since the user does not need the column itself they might also not need the formula column generated via that.

But if there are multiple column inputs and one of them is only deleted or some other case where the user might want to preserve the formula column. Can we preserve an empty column? As soon as the user interacts with the column they will need to update the parameters. 
Still not sure if we should preserve an invalid formula column but from a UX point of view we can preserve it in error state since deleting that might cause users pain.

Dhruvi16:
For the changed to invalid data type of the column, we can so something similar. Be transparent to the user and preserve the column on error state.

Dhruvi16:
I have a doubt @ghislaineguerin - Once the user has a successful preview, they need to go through two steps right? One is adding column to the query and one is saving the query. I am not sure why adding to query is an action. 

As the column is altered we can add it to query right without giving an action to the user. If not could you please shed light on why?

Less actions would make the journey more smoother I feel.

design/specs/add-formula-column.md
ghislaineguerin: For now we'll only allow formula columns while building queries (Data Explorer). In the future we might use formulas in other contexts.

design/specs/add-formula-column.md
ghislaineguerin: That sounds good. Thanks.

Dhruvi16:
I have another thing to discuss that is the recent section of the formula list in formula selection dropdown. Not sure how many recents can we show. Not sure what to consider when we decide on that.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Scenarios
+
+### Scenario 1 - Add a formula column without clicking outside of the inspector.
+
+![](https://share.balsamiq.com/c/dS6sAay4Re9f2R39zKDDj8.png)
+
+On the left hand side of the explorer view, there is a  `Add Column`  button. When user clicks on it, there are two options visible to them. One is `Add direct column` and the other is `Add formula column`. To add a formula column user will need to click on `Add formula column`. 
+
+Once the user clicks on `Add formula column` option, the inspector shows formula settings that include `Name` input of the formula and a list of formulas to select from. The list of formulas is described in detail in the [formula selection menu](#components) section.
```
ghislaineguerin, line 25: Formula settings in this case are shown after selection of the formula, the inspector will show details for the column. 
I'd go with something like:
The column details for a formula column are displayed in the inspector; from here, the user can name the column and select a formula from a list.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+### Scenario 1 - Add a formula column without clicking outside of the inspector.
+
+![](https://share.balsamiq.com/c/dS6sAay4Re9f2R39zKDDj8.png)
+
+On the left hand side of the explorer view, there is a  `Add Column`  button. When user clicks on it, there are two options visible to them. One is `Add direct column` and the other is `Add formula column`. To add a formula column user will need to click on `Add formula column`. 
+
+Once the user clicks on `Add formula column` option, the inspector shows formula settings that include `Name` input of the formula and a list of formulas to select from. The list of formulas is described in detail in the [formula selection menu](#components) section. 
+
+If the name is left empty by user for the next steps, there is a default name given to the column that is FormulaX. X is the number given to the column. For instance if its the first default named column it will be Formula1.
+
```
ghislaineguerin, line 28: I think the name will be set by default at the moment of column creation. The user can choose to change it.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+Case 2 - The user changes the selected formula.
+Once the user selects a different formula parameter values are not retained and they see empty parameters and an empty column again. To see the value they will have to add compatible params again.
+
+### Scenario 6 - Click outside the inspector
+
+#### Scenario 6.a - Clicks outside before adding name or selecting formula
+
+If the users click outside the inspector before even touching the formula settings that is adding name or selecting formula, there won't be any empty column added. And there would be absolutely no change in the explorer.
+
```
ghislaineguerin, line 113: I believe we discussed changing this so that all columns added remain until explicitly deleted by the user.

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,180 @@
...
+The column preview is necessary for the user to see the reflected changes when they interact with the formula settings. The different states of the column preview can be seen below. 
+
+1. Active empty column preview (Green highlight) - Users will see the active empty column preview when they are interacting with the formula settings in the inspector.
+
+2. Inactive empty column with a warning (Yellow highlight) - Users will see the inactive column preview with a warning when the user has click outside the inspector. 
+
+3. Erroneous empty column (Red highlight) - Users will see erroneous column when there are errors in formula settings inputs.
+
+![](https://share.balsamiq.com/c/bhzoPkNR4MtQnpb4mB2hws.png)
+
```
ghislaineguerin, line 152: I thought we'd moved past the need for the preview? Because we don't have a save action, it won't be obvious when the preview ends and the new column is added, which could result in users losing their progress inadvertently.

ghislaineguerin (changes requested):

design/specs/add-formula-column.md
```diff
@@ -0,0 +1,205 @@
...
+If the name is left empty by user for the next steps, there is a default name given to the column that is ColumnX. X is the number given to the column. For instance if its the first default named column it will be Column1.
+
+Once the user decides on what formula they want, the inspector shows relevant inputs for formula selected. The parameters can be of different data types depending on the selected formula. Parameters of the selected formula are described in scenarios below. 
+
+Before adding the parameters (if they are required), user would be able to see a preview of an empty column. 
+
+Once the user adds correct parameter the empty column gets filled with the calculated date and the user is able to preview the filled column. If they are not satisfied by the results they can change the formula selected as well as the parameters before adding the column to the query, this scenario is covered under edit column scenario.
+
+The `Add column` button in the inspector adds the column to the query. This button is only clickable if the column is filled.
```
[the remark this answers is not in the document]
Dhruvi16: Got it.

design/specs/add-formula-column.md
Dhruvi16: Got it

design/specs/add-formula-column.md
Dhruvi16: Yes.

design/specs/add-formula-column.md
Dhruvi16: You are right. Will change

design/specs/add-formula-column.md
Dhruvi16: yep.

Dhruvi16:
@ghislaineguerin I have updated the spec as per our recent discussions. I still do not have clarity on quit data explorer action and what happens if the data explorer is closed before saving.

Dhruvi16:
I have added two scenarios -
1. When user clicks outside inspector without saving
2. When user closes the explorer without saving.

For the second one, I wanted to know what are we doing when other changes are made in the query and not saved before closing. 
@ghislaineguerin

Dhruvi16:
Sure

ghislaineguerin (approved):

ghislaineguerin:
@kgodey I recommend we merge this as is and add a warning to the top since the current Data Explorer work will end up affecting this design. It does contain all required implementation details.

kgodey:
Okay, it's on the wiki here https://wiki.mathesar.org/en/design/specs/add-formula-column with a warning on top. @Dhruvi16 @ghislaineguerin feel free to edit as needed.

PR #48, opened 2022-07-27
Record Page design specs

seancolsen:
## The purpose of this PR

- This is my proposal for the Record Page design.

- Within our [features spreadsheet](https://docs.google.com/spreadsheets/d/1IqW3GiwYibvS3gleIKHhTa51WF_coOg2BqJA5ffPtVI/edit#gid=0), my intention is to set the Record Page UX Design Status to 100% complete once we merge this PR.

- Based on the spreadsheet, we need approval from @kgodey, @ghislaineguerin, and @mathemancer -- but I welcome critique from others too.

## Notes

- @ghislaineguerin had previously written an early draft of these specs based on our conversation. As the design owner, I chose to author this document from scratch, but to preserve Ghislaine's placement of this content within the wiki -- hence the large amount of "changes" in the diff here. I would recommend reviewing the [rendered content, as I wrote it](https://github.com/centerofci/mathesar-wiki/blob/record_page/design/specs/record-page.md) instead of the diff.


design/specs/record-page.md
```diff
@@ -1,71 +1,137 @@
...
+## Direct Fields
 
-### C.Record Fields 
+- This is a section of the page which comes after the title.
+- It contains a form-like area which has many labeled input fields. One field exists for each column in this table.
+- The fields are presented in the same order that they occur in the Table Page.
+- Field labels display the column name along with the icon used within the column header on the Table Page.
+- Edits to the field values are saved on blur, with a similar UX to editing cell values in a table. There is no save button.
+- Similar to the Table Page, a persistent but subtle status message appears at the top of all the fields which reads "Saving" or "All changes saved" when appropriate.
+- FK fields display the record summary for the linked record and allow the user to navigate to its Record Page.
```
ghislaineguerin, line 48: How do we handle navigating to a record vs updating or deleting a record?

design/specs/record-page.md
```diff
@@ -1,71 +1,137 @@
...
+
+- Above all widgets, a "Show Hidden Queries" dropdown button exists giving the user the capability to un-hide any auto-generated widgets widgets they have hidden.
+
+- Each widget is collapsible via a triangle icon to the left of its title. The collapsed/expanded state per-widget is persisted via local storage and will apply to all records within this table.
+    
+- The default ordering of widgets is not defined, but users *can* re-order all the widgets as as they so choose.
+
+    The mechanism for re-ordering the widgets is subject to further UX experimentation. Drag and drop would seem to be an obvious choice, though the height of each widget (when expanded) would make dragging cumbersome. It may be possible to collapse all widgets when the user initiates a drag, however the user's cursor might jump around on the page if we do this.
```
ghislaineguerin, line 135: As you point out, the height of the widgets will complicate drag-and-drop ordering. You may instead experiment with a menu that allows you to change the visibility and ordering of the widgets.
image

ghislaineguerin (commented):
Thanks Sean, looks pretty good. I've added a couple of comments.

design/specs/record-page.md
seancolsen: Yeah, that's a good question. By _"updating or deleting a record"_, I assuming you mean _"updating the FK cell value, or setting it to NULL"_. This question applies in many other contexts too. In #1467, I have begun building a `LinkedRecordInput` component which handles this behavior. It'll be used for FK cells in the Table Page, for FK fields on the Record Page, for FK inputs within the Record Selector, for setting the default value of an FK column, and for setting the filter condition value when filtering an FK column. It may need to have some slight differences in behavior across those various contexts. For example, maybe we don't need to provide the navigation capability within a filter condition. I'm not sure. 

I think the best path forward here is for me to continue improvising it within the Record Selector implementation and for you to review that work once I feel it's complete. At that point we can have a more productive discussion critiquing the `LinkedRecordInput` design because we'll have something more tangible. Is that approach okay with you, @ghislaineguerin? Or would you like to see more detail added to the specs within this PR?

design/specs/record-page.md
seancolsen: Cool idea! In 9a59ef10f964f3b04a1d9cd1f1d1335f5e16ab78 I've added the following text to the specs:

> Another idea is a drop down menu in which the user can re-order the widgets. This menu could potentially contain UI to show/hide the widgets too!

design/specs/record-page.md
kgodey: I was also wondering the same thing as @ghislaineguerin. I just want to make sure we handle navigation, updates, and deletes. Improvising the design during implementation sounds like a good plan to me.

design/specs/record-page.md
```diff
@@ -1,71 +1,137 @@
...
 
-The record page will also have a section with links to the selected record. This work is currently being defined, and it is likely to change. However, the aim is to add a default query for the tables with connected records and present them as tables or visualizations that the user can edit.
+One "widget" allows the user to see (and potentially modify) other data in the same schema *from the context of this record*.
 
-Custom embedded queries created by users with Data Explorer will be used to define the tables' fields and options.
+**Example:** On a Patrons record, we use a widget to see the Checkouts records associated with the patron.
 
-## Navigating to the record page
+*Note: The name "widget" may be a bit awkward, we're open to replacements. A widget is kind of like a query or a view -- but "query" and "view" already mean very specific things and a widget may not necessarily be either a Mathesar query or a Postgres view. Also, although these specs describe a limited set of initial functionality, we expect widgets to gradually evolve into a flexible and powerful feature (more details at the end of these specs), so we've coined a unique (and somewhat generic) name for it.*
```
kgodey, line 58: I think "widget" is a fine name, it's used for embedded versions of features quite widely.

For more precision I would call the auto-generated widgets "table widgets" (since they're automatically generated from a table and similar to tables) and the manually-generated widgets "query widgets" (since they're more freeform and are edited using our query builder).

design/specs/record-page.md
```diff
@@ -1,71 +1,137 @@
...
-![image](/assets/design/specs/record-page/179518507-ff971ee7-fc09-4c65-aaf0-df8a73743998.png)
+- Each widget has a **title** like:
+    
+    > Related `Checkouts` Records
 
-> This is contingent on the new design for the listed tables, which will contain buttons for searching and adding entries right from the list item.
-{.is-info}
+    If this table is referenced by multiple FKs from the same (other) table, or this table is referenced by an FK within this table itself, then the widget title will be:
+    
+    > Related `Checkouts` Records *(via `Patron Id`)*
```
kgodey, line 70: I like this.

design/specs/record-page.md
```diff
@@ -1,71 +1,137 @@
...
+    - **Manually generated widgets** (which are more complex and will hopefully be implemented later)
+    
+        Some example use-cases for manually-generated widgets are:
+
+        - Show a patron's currently-checked out books
+        - Show all publications of an author, along with the number of items that the library has for that publication.
+
+- The title is stylized differently for the two types of widgets (perhaps via a lock icon or similar) allowing the user to distinguish them.
+
+- Above all widgets, an "Add Query" button exists to add a new manually-generated widget.
```
kgodey, line 112: I'm not sure if we should call this "Query" (maybe "Widget" or "Related Records" or whatever) but we can figure out the name closer to implementation.

kgodey (approved):
Looks great @seancolsen! I added some inline comments, but they're mainly just thoughts.

design/specs/record-page.md
seancolsen: Yeah I struggled with this too. But that'll be easy to tweak later.

design/specs/record-page.md
seancolsen: Brilliant! This should be addressed in f13e2c75abfd6cba3c3b12e1517426c9bf11dc21

seancolsen:
I'm fairly certain I've addressed everyone's review critique, so I've gone ahead and merged this for the sake of moving forward quickly. I've also set the status to 100% in the [features spreadsheet](https://docs.google.com/spreadsheets/d/1IqW3GiwYibvS3gleIKHhTa51WF_coOg2BqJA5ffPtVI/edit#gid=0). Even though the status is 100%, I'm always open to revisiting these design choices!

mathemancer (commented):
I think this sounds great, and should be doable with our current API.

design/specs/record-page.md
seancolsen: Update: I'm proposing more detail in #50

PR #49, opened 2022-07-27
Navigation Specs

ghislaineguerin:
The goal of this PR is to propose specs for Navigation Design.
This spec is a revised version of the one currently in the wiki, it incorporates changes based on suggestions from @Sean.


design/specs/navigation.md
```diff
@@ -0,0 +1,133 @@
...
+- Through links in pages
+- Direct access through page URLs
+
+### Navigation via site header
+
+- The logo in the site header will lead visitors to the site's root, in this case the database homepage.
+- A breadcrumb-like navigation will be displayed at the top part of the site's header as the user navigates through the pages.
+- When users click on a breadcrumb element, a [page-selector](#page-selector) component is displayed with a list of schemas or entities based on the breadcrumb level.
+- Upon reaching the record level, the top navigation will show a 'Go to Record' action, which will open the [Record Selector](/engineering/specs/record-selector) component.
+- The current page will not be included in the breadcrumb path.
```
kgodey, line 56: I'm not sure what this means. Could you maybe include examples of breadcrumbs on each type of page?

design/specs/navigation.md
```diff
@@ -0,0 +1,133 @@
...
+  - Links to **Schema Pages** in a list of schemas.
+
+- **Schema Page**
+  - Links to **Table Pages** in a list of tables.
+  - Links to **Query Pages** in a list of queries.
+
+- **Table Page**
+  - Within the table data:
+    - Links to the **Record Page** for each row via the row's primary key cell value. _Note: Links within cell values will only be clickable when the cell is selected. This UX should be consistent with the behavior of the "URL" UI Type._
+    - Links to **Record Pages** for related records via the FK cell values
```
kgodey, line 75: Could we also link to the table page for the related table from the FK column header or something like that?

I think it would be nice for the user to be able to navigate to a related table without opening up the table inspector, but maybe that's over-optimizing.

design/specs/navigation.md
```diff
@@ -0,0 +1,133 @@
...
+  - Route: `/{database_name}/{schema_id}`
+  - Favicon: Schema
+  - `<title>`: `{schema_name} | Mathesar`
+
+- **Table Page**:
+  - Route:  `/{database_name}/{schema_id}/{table_id}/`
+  - Favicon: Table
+  - `<title>`: `{table_name} | {schema_name} | Mathesar`
+
+- **Query Editor Page**:
```
kgodey, line 106: What is the query editor page?

design/specs/navigation.md
```diff
@@ -0,0 +1,133 @@
...
+  - Route: `/{database_name}/{schema_id}/query/{table_id}/`
```
kgodey, line 107: what is the table id here?

design/specs/navigation.md
```diff
@@ -0,0 +1,133 @@
...
+  - Links to the **Table Page** for related tables via a section within the Table Inspector.
+
+- **Data Explorer**
+  - Features the same links as described by the "Within the table data" bullet point on Table Page above.
+
+- **Record Page**
+  - Links to other **Record Pages** via FK fields on directly on the record.
+  - Links to other **Record Pages** via tabular displays of _other records linking to this record_ via the same UI as the Table Page which has links in the PK cells.
+
+### Navigation via page url
```
kgodey, line 85: This is missing URLs for:
- Data Explorer
- Data Import
- Query Page

design/specs/navigation.md
```diff
@@ -0,0 +1,133 @@
...
+- When users click on a breadcrumb element, a [page-selector](#page-selector) component is displayed with a list of schemas or entities based on the breadcrumb level.
+- Upon reaching the record level, the top navigation will show a 'Go to Record' action, which will open the [Record Selector](/engineering/specs/record-selector) component.
+- The current page will not be included in the breadcrumb path.
+- The current page label will be placed underneath the top header component in a sub-header component.
+- An icon will be used to identify each type of page (table, schema, record etc.)
+- When we eventually add support for several databases, we'll need CRUD UI for databases, and we may design some additional UI that stays within the header at that point.
+
+### Navigation via links within pages
+
+Throughout Mathesar, the following links exist within pages which allow for more targeted navigation to other pages.
```
kgodey, line 63: Wouldn't there also be additional links to the Data Explorer from the Table Page? (e.g. to show a summarized view once the user groups something on the table page, or to build a query with filters applied on the Table Page)

kgodey (commented):
Looks pretty good to me overall, added some comments about things that confused me.

I would like to see more in-line wireframes to make some of the ideas clearer, but if @seancolsen has enough detail to implement, then that's fine. I'll just wait for the implementation to be done.

design/specs/navigation.md
seancolsen: Whoops that's my bad. I've changed that to `{query_id}` in 8383be0f1ad64fcd70d2bd5e657b9455770bb3c4

design/specs/navigation.md
seancolsen: I wrote that text. Maybe this should be "Data Explorer" instead? My understanding is that we are currently implementing a page where the user edits the query definition (while also previewing the results of that query) -- and that in the "Query Page" feature, we plan to implement a page where users will view the results of the query (without the UI to edit the query definition). I think I've just been subconsciously categorizing all our pages as nouns and "Data Explorer" didn't quite fit into the pattern of "Database", "Schema", "Table", "Record" etc so I was thinking we'd have a "Query Editor Page" and a "Query Results Page" or something, just to keep the naming consistent. But I have no qualms about changing this to "Data Explorer" (and I'm happy to add a commit which makes this change).

design/specs/navigation.md
seancolsen: I wrote this section and I didn't include that info because I don't yet understand how that functionality is supposed to work. I think we can safely omit those details from this spec and either: (a) include them within the Data Explorer spec (oh what that's at 100% now, so maybe not), or handle that work within smaller more-detailed tickets. I don't want to get too hung up on putting every detail into this "Navigation via links within pages", as I'm sure we're likely to discover more opportunities for inter-page navigation as the product evolves.

design/specs/navigation.md
seancolsen: > link to the table page for the related table from the FK column header

I'm hesitant about going this direction for the following reasons:

- In preliminary conversations about the Table Inspector design, I think Ghislaine and I have been coalescing on the idea that _clicking a column header should select all the cells within that column_. It's hard for me to imagine how to preserve that cell-selection design while also allowing the user to navigate to a Table Page via the column header.
- I'd like to stick to the design we [established in Matrix on Tuesday](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$JIGcTbMy-KCHeG5QAN0RkMa6fPT7npl32JdgVkbvIeI?via=matrix.mathesar.org&via=matrix.org), which keeps the column header as simple as possible. FK column get a link icon but nothing else special.

> or something like that

I'm open to hearing other ideas more specifically

seancolsen (commented):
I _think_ I'm good with everything here, although the "Navigation via site header" section is not super clear. As long as we're going with everything we discussed on the call, then I'm all set, but additional clarity here would help to ensure Kriti is on board with all of those specifics.

- In particular, this bullet point was a little confusing: 

    > When users click on a breadcrumb element, a page-selector component is displayed with a list of schemas or entities based on the breadcrumb level.

    What does "breadcrumb element" mean here?
    
    In [my design](https://gist.github.com/seancolsen/d3c42a14ba6c860af6e576205e5092d2) (depicted below) the breadcrumb UI is composed of: (a) "page buttons" which link directly to pages, and (b) "separator buttons" which open the page selector component you describe.

    mockup
    
    In your text, does "breadcrumb element" refer to (b), as in my design? Or does it refer to (a), as we briefly discussed in our call to be an alternative design employed by some other apps? I'd like to keep the "(b)" behavior in my design where the page-selector component opens in response to clicking the separator between page buttons.

- Perhaps it would be helpful to include the mockups that I put in my design, as they might help answer Kriti's question _"Could you maybe include examples of breadcrumbs on each type of page?"_

design/specs/navigation.md
kgodey: Let's not worry about this. If we run into use cases where navigating between tables needs to be made easier, we can figure it out then.

design/specs/navigation.md
kgodey: I would suggest three pages:

- **Query Page**: The is the "query results page" you're talking about. URL `/{database_name}/{schema_id}/query/{query_id}/`
- **Data Explorer**: This is where you start creating a new query. URL `/{database_name}/{schema_id}/data-explorer/`
- **Data Explorer with existing query**: This is where you open a query in the Data Explorer. URL `/{database_name}/{schema_id}/data-explorer/{query_id}/`

design/specs/navigation.md
kgodey: Thanks!

design/specs/navigation.md
kgodey: I think it's fine to defer to later specs. Perhaps add a line in this spec that this isn't meant to be comprehensive and other opportunities for inter-page navigation are expected to occur.

design/specs/navigation.md
seancolsen: After wrestling with this sub-thread a bit, I've realized that I _do_ have some qualms here, and that they're actually deeper than the URL choices.

But before I launch into my thinking, I'll begin by saying I don't feel very strongly about this and am fine with moving forward using your suggestion. These thoughts are out-of-scope with the Navigation specs anyway. But if my thinking resonates with you, then consider my proposal at the end of this message. If not, let's go with your proposal and I'll add a commit which modifies the spec to match it...

Fundamentally, I don't see the utility of using the terminology "query" _and_ the terminology "data explorer" to describe such similar aspects of the product. I understand that we use the data explorer to build queries. Are there _other_ things the data explorer will eventually build? If so, then distinct terms seems useful. If not, then I think the distinct terms will be confusing to users. These names have already been confusing _me_! I think users will have an easier time understanding Mathesar if we use _one_ of these terms consistently. While the term "data explorer" feels more inspiring and powerful, I prefer "query" because it feels simpler and more accurate.

The fact that I _"use the Data Explorer to create a Query"_ is not immediately evident from the terminology, whereas _"using the Query Editor to create a Query"_ is.

My inclination would be:

| Name | Url |
| -- | -- |
| New Query | /{database_name}/{schema_id}/query/ | 
| Query Editor | /{database_name}/{schema_id}/query/{query_id}/edit | 
| Query Viewer | /{database_name}/{schema_id}/query/{query_id}/ | 

With this terminology describing the features and functionality, we could retain the "explore" language in other contexts such as marketing copy like:

> Explore your data by creating custom queries and visualizations

design/specs/navigation.md
ghislaineguerin: I've included examples for each type of page.

design/specs/navigation.md
ghislaineguerin: > In preliminary conversations about the Table Inspector design, I think Ghislaine and I have been coalescing on the idea that clicking a column header should select all the cells within that column. 

This is consistent with the selection patterns I'm using for the table inspector UX. I think it should remain simple.

design/specs/navigation.md
ghislaineguerin: @seancolsen What you're proposing makes logic to me. When a query enters edit mode, users may be perplexed as to why the language changes to data explorer.
To maintain consistency with your example, we must also update the list with 'New Table,' 'New Record,' and 'New Schema.' Do you agree?

ghislaineguerin:
> I _think_ I'm good with everything here, although the "Navigation via site header" section is not super clear. As long as we're going with everything we discussed on the call, then I'm all set, but additional clarity here would help to ensure Kriti is on board with all of those specifics too.

> * In particular, this bullet point was a little confusing:
>   > When users click on a breadcrumb element, a page-selector component is displayed with a list of schemas or entities based on the breadcrumb level.

I agree, I've updated that section.
    
>   What does "breadcrumb element" mean here?

I was thinking it's each item representing a page

>   In [my design](https://gist.github.com/seancolsen/d3c42a14ba6c860af6e576205e5092d2) (depicted below) the breadcrumb UI is composed of: (a) "page buttons" which link directly to pages, and (b) "separator buttons" which open the page selector component you describe.
>   mockup
>   In your text, does "breadcrumb element" refer to (b), as in my design? Or does it refer to (a), as we briefly discussed in our call to be an alternative design employed by some other apps? I'd like to keep the "(b)" behavior in my design where the page-selector component opens in response to clicking the separator between page buttons.

It was the combination of both. I have now split them into the page link and toggle. 

> * Perhaps it would be helpful to include the mockups that I put in my design, as they might help answer Kriti's question _"Could you maybe include examples of breadcrumbs on each type of page?"_

Will do.

design/specs/navigation.md
seancolsen: > we must also update the list with 'New Table,' 'New Record,' and 'New Schema.' 

When you say "update the list", are you suggesting that we create additional pages for "New Table", "New Record", and "New Schema"? I'd be open to that, but I'm not sure it's necessary.

- When creating a schema, we only need a name. We acquire that via modal, which I thing provides for a more seamless workflow than a dedicated page would.
- We have two ways of creating a new table: (1) via import (which as @kgodey pointed out _does_ need an entry in the "Navigation via page url" section), or (2) via empty table (which I don't think warrants its own page).
- I think a dedicated page for "New Record" would be nice, but I'd rather not work on implementing it right now. I have some more ideas for this feature which I hope will aid with rapid data entry, but I think they're out of scope for this cycle, given that none of the storyboard screens seem to require a New Record page.

seancolsen:
@ghislaineguerin the additional detail you've added to the "Navigation via site header" section is great.

Design-wise, I do wonder whether we need to display the database name and database icon in the site header. After our [conversation last week in Matrix about multiple databases](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$uvABaY5sN5gz_tlIlgQFdJy7BqcKqgGSVHWTQ0B6jQA?via=matrix.mathesar.org&via=matrix.org) I was thinking it might be better to hide it, at least for the time being. 

Kriti said:

> I don't think it needs to be as easy (in terms of number of clicks/actions) to navigate to a different database as it should be to navigate between schemas, even when we do support multiple databases

That makes sense. I imagine that even moderately technical users might not understand the difference between "database" and "schema". For example, I used MySQL for several years where "database" and "schema" are interchangeable terms. A MySQL database contains _tables_, not _schemas_. So when I started learning PostgreSQL, the concept of having multiple schemas inside one database was very confusing at first.

I don't want to get too carried away with designing a UI for switching databases within Mathesar, but I have been thinking it would be nice to have that UI a bit more hidden, perhaps behind a dropdown, so that the user's attention is more on the schema and the concept of "the database" does not contribute to cognitive noise while learning the product. In [my mockup](https://user-images.githubusercontent.com/42411/180277150-e68cf483-9ce5-436f-b347-714937d22295.png), the header has a person icon at the far right. I've been imagining that a click to that icon would open a dropdown with some additional UI where we might put other settings, including a database switcher. A slightly more prominent approach would be to show a small gray dropdown database-switcher above the user icon.

I'm curious to hear thoughts from @kgodey here. Do you think the user needs to _see_ which database they're currently using at all times? If so, how prominently should it be displayed?

design/specs/navigation.md
kgodey: I feel pretty strongly about continuing to use the term "Data Explorer", I think it's more friendly to new users and conveys the concept of what users are meant to do with the feature better. Also, users don't _have_ to save their output from the Data Explorer, they should be able to use it to explore without saving. So tying the name to only one potential output (a saved query) doesn't make sense.

If we rename anything (which I'm not proposing we do right now), I think we should get rid of the word "Query" because it's a technical term. We can't rename schemas or tables because they represent database objects, but queries are Mathesar-only objects and we can call them whatever we want.

kgodey:
> Do you think the user needs to see which database they're currently using at all times? If so, how prominently should it be displayed?

No, I don't think that's necessary. If we find that users are confused about it after the product is released, we can solve that problem then.

design/specs/navigation.md
kgodey: Looks good, thanks.

kgodey (commented):
Most of this looks great. I'll approve once the discussion around navigation via URL and what pages need to be present is resolved, @seancolsen and @ghislaineguerin.

ghislaineguerin:
@seancolsen @kgodey Thanks for the feedback. I would leave the database reference because I'm thinking multiple Mathesar installs could exist for the same user. How else would they tell them apart? Do we add something referencing the database at the url level?

kgodey:
@ghislaineguerin that is a good point, if people have access to different mathesar installations, it would be good to have something to tell them apart.

ghislaineguerin:
@kgodey I have updated this to incorporate the terminology changes.

PR #51, opened 2022-08-02
Table Inspector Spec

ghislaineguerin:
The goal of this PR is to propose specs for Table Inspector.

Because the table inspector design has several interconnected aspects, I need specific input for:
- Table Inspector Behavior (should it be disabled or remain enabled always etc.)
- Table Inspector Content (sections that we need for each selection mode).
- Thoughts on having an 'Actions' panel instead of a disjointed set of UI controls for individual actions.


design/specs/table_inspector.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Directly click on the inspector toggle located in the table toolbar to hide the table inspector.
+
+## Using the Table Inspector
+
+Whatever is presently selected will determine the content of the table inspector. During non-selection mode, the inspector simply displays properties for the current table.
+
+### Table Inspector Modes
+
+The current selection is what triggers the various inspection modes. The inspector panel's header contains a brief description of what is included in the selection.
```
rajatvijay, line 44: > The inspector panel's header contains a brief description of what is included in the selection.

Is there a mock for this?

design/specs/table_inspector.md
```diff
@@ -0,0 +1,163 @@
...
+  
+![image](/assets/design/specs/table_inspector/182350542-a5af060c-16f7-40b5-9a00-b7b6cc32e2a2.png)
+
+### Multiple Columns
+
+Shown when multiple columns are selected.
+
+Multiple column properties:
+
+- Data Type
```
rajatvijay, line 95: This option won't always be possible right since the multiple columns selected can be of two different types.

design/specs/table_inspector.md
```diff
@@ -0,0 +1,163 @@
...
+
+![image](/assets/design/specs/table_inspector/182350659-10d1b6f7-2cbf-44cb-a728-8241ad9f7335.png)
+
+### Actions Panel
+
+Shown at all times, the content might change depending on the selection status.
```
rajatvijay, line 101: Detailed states of `Actions Panel` is still pending, right?

design/specs/table_inspector.md
```diff
@@ -0,0 +1,163 @@
...
+
+![image](/assets/design/specs/table_inspector/182121417-8eb51f90-8767-4a01-bc36-a9c55c1614b5.png)
+
+#### Row Selection
+
+![image](/assets/design/specs/table_inspector/182122068-c08fecea-7d18-48ad-a8de-1cb1992ca8aa.png)
+
+#### To Deselect
+
+Deselect all selected items by clicking on a currently selected item, on an empty space or pressing the escape key.
```
rajatvijay, line 127: 1. What is the initiation of the selection mode? Do you mean opening the inspector makes the table go into selection mode?
2. I think we should give a explicit action for `Deselect All`.

ghislaineguerin:
@kgodey @seancolsen @rajatvijay 
I have re-started review for this PR

seancolsen (changes requested):
## Overall impressions

- This is looking great! My remaining thoughts (though a bit verbose 😬) are quite minor.


## Clarifying questions

- How persistent is the visibility state of the Table Inspector? I see that it should be visible after navigating to a newly-imported table. That sounds good. If I hide it, does it remain hidden when I re-open that Table Page? If so, where is that state stored? In the browser's local storage, or in the Mathesar database? (I want to be clear that I view this decision as a design decision, not an implementation decision because it affects functionality.) How contextually specific is that visibility state as it's stored? Is visibility stored globally for all tables, all schemas, all databases? Or is visibility specific to each table?

- When editing fields like the table name and column name, how do I save the value? Does it save on blur? Or is there a save button?

- What happens when I click on the "Edit Selection" button?

- Most of our competitors¹, retain state for the users selection _and_ the **active cell** separately. The selection can contain multiple cells, whereas the active cell can only be one cell. This way you can edit the active cell even when you have multiple cells selected. I'm a bit torn on what I think would be best for our product at this stage. Google Tables, for example, does not allow editing cell contents when multiple cells are selected. That behavior is more simplistic than, say, Google Sheets which keeps an active cell within the selection at all times. I think I'd lean towards this behavior for us because it would probably be the easiest to implement.

## Critique

- For all modes

    - I'd like to re-order the "ACTIONS" section, placing it after "PROPERTIES" -- and then to make "Delete" an action instead of a button at the very bottom. Why? Because I want more vertical space in the inspector. Here's my thinking... We can probably all agree that the four tabs at the top of the Table Inspector should not scroll with the contents of the Table Inspector. We also have the following UI consuming vertical space in the viewport: site header, table actions bar, table status bar (at the bottom of the page for pagination). The delete button adds a little more pressure to that space. And we need all the space we can get because [the type configuration options can be a bit unwieldy](https://github.com/centerofci/mathesar/issues/1325#issuecomment-1139996794) -- a problem we've been planning to address with the Table Inspector.

- In "Cell Mode"

    - I'd like to remove "Delete Record", as that's a record-level operation, not cell-level.
    - I'd like to see "Set to NULL" moved to an "ACTIONS" section, stylized similarly to other actions.

- In "Column Mode"

    - I'd like the spec to indicate the behavior of "Allow NULL" and "Allow Duplicates" when multiple columns are selected. I propose that (at least for the time being) these actions are disabled when multiple columns are selected.
    
- The column header menu (which will be removed when implementing this spec) currently has UI to sort/group by the column, which I predict will be a commonly-used feature. With the column header menu gone, the specs as-written leave it to the user to manually add sorting/grouping conditions, which requires more clicking and thinking. It would be great to address this regression somehow. I'm not sure if the Table Inspector is best. Perhaps that UI gets moved into the ACTIONS section? Note however that the options in the column header menu currently function more as checkboxes than they do as actions. We could also leverage the cell context menu, by adding some UI there that doesn't necessarily get included within the Table Inspector.

- "Selection Modes"

    The mockups here don't quite capture my expected behavior. I would suggest deleting those mockups as they might be misleading. I'd replace them with the following text:

    > - The selected rows and columns are derived from the rows and columns that contain selected cells.
    > - Row headers and column headers are styled to visually indicate the selected rows and selected columns.
    > - Unselected cells do not receive any special styling, even if they fall within a selected row or selected column.

    I want to draw attention to that third bullet point; I don't think we should stylize deselected cells within selected rows any differently from deselected cells within deselected rows. For example, the first two mockups appear to contain a sheet with one selected cell but with a blue background on all the other cells in that selected row. I'd like for the _row header_ to be blue, but not the other cells in that row. The other present inconsistent behavior. For example, in the first "Column Selection" mockup, I'd expect all row headers to be stylized as selected because all rows contain selected cells. The corollary is true for the "Row Selection" mockups. All-things-considered, I'd say it's not worth the time to re-do the mockups.

- "To Select All"

    I'm confused by this text:

    > For records, the option to select all is available, which would include all columns and cells in the selection. The column header should include a control for this.

    - First of all, "is available" is a bit vague. Can you elaborate?

    - Also, putting a control within the column header seems to go against the design of using the column header to select all cells within the column. Maybe we're talking about the same thing but I'm just not understanding your phrasing.

    I propose rewriting the "To Select All" section as follows:

    > - Clicking on a row header selects all the cells within that row.
    > - Clicking on a column header selects all the cells within that column -- but only the cells within the current pagination page. There is no way (yet) to select cells across multiple pagination pages.

    In the future we could do additional design work to allow the user to select cells across multiple pagination pages. But I don't think we should rush into it. We could also allow the user to click the cell above all row headers and to the left of all column headers as a means to select all the cells within the sheet, but it's a little hard for me to imagine that being a common-enough operation to address any time soon.

- "To Deselect"

    > Deselect all selected items by... pressing the escape key

    For consistency with competitors¹, I'd like to retain the selection when pressing Escape. We also use Escape to cancel edit mode. If Escape does two things it might lead to users unintentionally clearing their selection.



## Comments for implementation

For these points, I'm not sure the spec needs any changes (because I imagine we're all picturing the same thing), but I want to mention them, especially for @rajatvijay.

- When planning this cycle, I was (perhaps naively) expecting that the Table Inspector implementation would require us only to _relocate_ existing UI -- not to build any new features. But now that we have a spec, I see that it contains new features too, like table descriptions, the display of table links, remove duplicates, and more. These are all great features, and I'm glad we're putting them into the Table Inspector. But for the purposes of planning and organizing our implementation, I think we (@kgodey?) might need to make some prioritization decisions. For example, I don't think we need table/column descriptions for the demo video, and I'd like to be able to mark the Table Inspector work as 100% complete without implementing it.

    I'd like to figure out a way to separate these additional work requirements from the Table Inspector feature, and I don't want to require @rajatvijay (the primary implementer) to create tickets for all of that work as part of his implementation.

- On a similar note... With the UI being capable of performing bulk operations on columns, records, and cells, I'm a little worried that we might not have sufficient API support to make those operations performant and easy to implement. For example, currently we can delete multiple rows at once. But that operation is [actually performed as multiple API requests in parallel](https://github.com/centerofci/mathesar/blob/a0708dcc7d715fd213c7aa6ebbbdeab686880fb8/mathesar_ui/src/stores/table-data/records.ts#L347). The code is ugly and slow. How do we want to approach this problem for other things like setting cells values to NULL, deleting columns, etc? I'm inclined to _disable_ these sorts of actions until we have proper API support for them, especially since we can't perform those actions with the current UI and (I'd argue) the main purpose of this Table Inspector feature is to lay groundwork for building more powerful features -- not to actually build all those features right immediately. @kgodey do you have thoughts on this?

Maybe we could discuss the above points at a team meeting to figure out how to coordinate/assign/manage that work?


## Footnotes

- ¹ I'm defining competitors as follows: AirTable, Google Sheets, Google Tables, LibreOffice (because these are products that I've been juxtaposing alongside our UX design process)

kgodey:
I'll do in-line comments on the spec after this, but here's my response to @seancolsen's comment:

> When editing fields like the table name and column name, how do I save the value? Does it save on blur? Or is there a save button?

We should have an explicit save action here, since editing the structure of tables and columns is a DDL operation. In general, we should be more cautious with DDL changes than with DML changes.

> When planning this cycle, I was (perhaps naively) expecting that the Table Inspector implementation would require us only to _relocate_ existing UI

> (I'd argue) the main purpose of this Table Inspector feature is to lay groundwork for building more powerful features -- not to actually build all those features right immediately.

This was exactly the intention, yes. We should not be implementing new functionality as part of this features.

> I see that it contains new features too, like table descriptions, the display of table links, remove duplicates, and more.

These should be separated out into a "possible future features" section or something like that. It's good that we're thinking ahead to see how the table inspector could be used, but we should not be implementing them right now.

> I'm inclined to _disable_ these sorts of actions until we have proper API support for them, especially since we can't perform those actions with the current UI

Agreed.

> Maybe we could discuss the above points at a team meeting to figure out how to coordinate/assign/manage that work?

I'm not sure we need a meeting for this. I think @ghislaineguerin should create a "future features" section and move anything that's not already implemented in there. There should be separate specs for anything new we plan to implement (e.g. column extraction, dynamic defaults).

---
When I talked with @rajatvijay earlier today, he said his plan was to implement this feature as follows:

> 1.  Remove row selection and implement Cell selection. Make the cell selection work with current row selection actions
> 2.  Implement the UI of table inspector with its opening and closing trigger
> 3.  Move the actions one by one to table inspector

I think this makes a lot of sense. The goal of the spec should be to support this implementation plan, and not add any new features

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Using the Table Inspector
+
+Selecting different inspector modes in the table inspector allows users to examine at the properties, options and settings of table components. Aside from that, various actions can also be initiated through the inspector.
+
+### Table Mode
+
+Shown when inspector mode is set to `Table`.
+
+Properties:
```
kgodey, line 62: Please divide this list of properties/settings/actions into features to implement for Cycle 3 (existing features only) and new features that should be implemented later.

kgodey: We also need room for a Save button somewhere.

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+- Table Name
+- Table Descriptions
+
+Settings:
```
kgodey, line 67: How are settings different from properties?

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+Shown when inspector mode is set to `Column`.
+
+Properties:
+
+- Column Name
+- If Column is Link:
+  - Link Source
+
+Options:
```
kgodey, line 91: I think it would be good to consolidate the nomenclature here. Properties, Settings, Options – these all seem like the same thing to me. I think we should be able to consolidate inspector to have two sections: Properties & Actions.

It would also be good to distinguish database-level properties (like name and description) from Mathesar-level properties (like record summary, column formatting options etc.) Maybe an icon or something can be used.

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+- Allow NULL
+- Allow Duplicates
+- Data Type
```
kgodey, line 95: I think it's a little confusing to have tabs for database and formatting options within the tabs for the inspector. Now that we have the full height inspector, can't we show both without tabs?

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+  - Database Options
+  - Format Options
+
+Actions:
+
+- Extract Columns to Table
```
kgodey, line 101: I think this should be a table level action

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+![image](/assets/design/specs/table_inspector/183432613-ead6315e-3802-4345-9427-820c66094797.png)
+  
+![image](/assets/design/specs/table_inspector/183433425-9a906a91-28a4-4045-9300-732af811ed8d.png)
+
+### Record Mode
```
kgodey, line 107: Can we just skip this tab for now since the only thing in there is a new feature?

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+Actions:
+
+- Duplicate Record
+
+![image](/assets/design/specs/table_inspector/183444464-fb268bfc-77e2-45cf-9180-373cf950ca63.png)
+
+### Cell Mode
+
+Shown when inspector mode is set to `Cell`.
```
kgodey, line 119: I would recommend that we have a Properties section here that shows the cell value and allows users to edit the cell, maybe. Set to `NULL` can be a button next to it.

design/specs/table_inspector.md
```diff
@@ -0,0 +1,205 @@
...
+
+![image](/assets/design/specs/table_inspector/183445209-a2d7bf2c-453b-4cae-84e8-94b645ce9271.png)
+
+### Multi-Select Mode
+
+Shown when multiple table components are selected. The table inspector will display the properties according to the current inspector mode and the shared properties between selected components.
```
kgodey, line 125: This is unclear, I'm not sure what it means.

kgodey (changes requested):

design/specs/table_inspector.md
```diff
@@ -0,0 +1,163 @@
...
+## Enabling the Table Inspector
+
+The table inspector can be enabled or disabled manually or automatically based on triggers.
+
+To enable the table inspector:
+
+- Directly click on the inspector toggle located in the table toolbar to show the table inspector.
+- Open a newly created table, as part of another flow like [data import](#data-import) or [column extraction](#column-extraction).
+
+When users enable the table inspector, a panel will appear on the right side of the screen, sharing the same container as the table and pushing its contents rather than being stacked on top of it.
```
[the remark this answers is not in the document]
ghislaineguerin: Yes, and whatever was visible at that moment should remain in the same position. Meaning the scroll is maintained.

design/specs/table_inspector.md
ghislaineguerin: Yes, the goal of this spec is to cover the sections in general but we'll define the various actions as we add them one by one.

rajatvijay:
@ghislaineguerin @kgodey @seancolsen  
In our previous call, we decided to remove the row selection and use the cell selection instead to take actions on row/cell/column. I was discussing this with Pavish and realized that we would miss the feature of selecting and deleting rows across pages. I was also checking out Apple Numbers (whose references we took during the discussion) and it also has a row selection pretty similar to google sheets (cmd + click). We have three options now

1. Keep row selection as it is. In this case, I will schedule a call with Pavish and Sean after our check-in call to clear some architectural roadblocks I hit.
2. The new way to select rows since the current checkboxes way will confuse the user once cell selection has landed. In this case, we should have a UX call (preferably after our check-in call) unless we are okay with getting rid of the row selection across page action and other such actions that might come up in the future.
3. Remove row selection altogether. In this case, too we should have a UX call (preferably after our check-in call)

Let me know if I miss something from our previous call that could help me here.
cc: @pavish

seancolsen:
Copied from [Matrix discussion](https://matrix.to/#/!xInuTkBwjXZXYatIlm:matrix.mathesar.org/$cfRqaK2PakCuNDXWTqYCasgt3gm99zBq4Y8Mi3kWIAE?via=matrix.mathesar.org&via=matrix.org)...

Reactions/thoughts:

- Let's refer to the process of "selecting and deleting rows across pages" as **multi-page record deletion**.
- Mathesar does not currently have multi-page record deletion, but I think it should eventually. For example, it would be great to apply a filter and then delete all the matching records even if they span multiple pages.
- As Kriti has recently [made clear](https://github.com/centerofci/mathesar-wiki/pull/51#issuecomment-1209746567), the scope of the Table Inspector feature should be limited to _relocating_ existing UI -- not building any new features. That means we don't need multi-page record deletion within this cycle.
- In redesigning the selection UX, I think multi-page record deletion is only relevant insofar as _we leave the door open to designing a UX for it eventually_. If our "cell-based selection" UX seems likely to inhibit multi-page record deletion later on, that's a problem. But I don't think it will...
- Here's a rough design for multi-page record deletion which builds on top of our current plans for cell-based selection:

    - The user's selection (if present) is _either:_ (a) many cells, or (b) a value indicating that all rows are selected. For example:

        ```ts
        interface SelectionOfCells {
          type: 'cells';
          /** Values are cell ids. */
          cells: Set<string>;
        }
        interface SelectionOfAllRecords {
          type: 'all-records';
        }
        type Selection = SelectionOfCells | SelectionOfAllRecords;

        let selection = writable<Selection | undefined>(undefined);
        ```

    - The "Records" tab of the table inspector displays the number of records selected _and_ the total number of records. E.g.

        > 100 of 12,876 Records

        Aside that UI, a button exists to "Select all records".

    - When all records are selected, no cells are selected. This means that no columns are selected.

- I want to reiterate: I don't think we should hash out the above design details now. Rather, we should only seek to answer the question: **"Is it _possible_ to design multi-page record deletion that builds on top of our current cell-based selection plans?"**. I present the above design to argue that the answer is "yes" -- meaning we ought to continue with that implementation and worry about multi-page record deletion later.

seancolsen:
We had a sync call about this spec today. Present were Sean (note-taking), Kriti, Pavish, Ghislaine, and Rajat.

Decisions we made during the call:

- **multi-page record deletion** -- we'll continue with our "cell-based selection" design. This will mean that the user has no way to delete records across multiple pages until we do more design work at some point in the future.

- **Visibility** -- Table inspector visibility does not persist. It's always open initially.

- **Save buttons** -- Fields that do DDL operations will have an "Edit" button. Then "Save" and "Cancel" buttons will appear. Anything that currently has a save button should continue to have a save button. Ghislaine will help with this design. Some of our current UI does actually perform DDL operations without save buttons (e.g. setting unique constraint). Those operations can remain as single-click operations.

- **"Edit Selection" button** -- not to be implemented this cycle.

- **"Active cell" vs "selected cells"** -- we'll retain state for _both_ (separately). We'll aim to mirror the functionality present in spreadsheets.

- **Delete button** -- will go inside the "actions" section.

- **Sections** -- we'll have "properties" and "actions". Everything in the inspector will be either a property or an action

- **Sorting and grouping** -- will be moved into the table inspector. Column headers should visually indicate whether the column is sorted or grouped, and we'll splinter this off into a small task

- **Selection modes** -- go with Sean's proposal

- **Scope** -- new features will be split off into a separate section of the spec

- **"Cell" tab** -- we won't implement this yet.

- **bulk operations/edits** -- we'll only support operations on multiple things at once if it's really easy to implement. Right now that might only be the "Column Extraction" feature.

- **Cell context menu**  -- we'll keep it as-is.

kgodey:
Next steps are:
- @rajatvijay updates this spec and asks for review from @ghislaineguerin @kgodey @seancolsen 
    - @rajatvijay will work with @ghislaineguerin as needed for designs for updating the Table Inspector to include explicit saving.
- @rajatvijay @pavish @seancolsen will have a call to work out some architectural issues

design/specs/table_inspector.md
rajatvijay: @kgodey @seancolsen @ghislaineguerin can you help me remember the conclusion on this?

design/specs/table_inspector.md
rajatvijay: @kgodey @seancolsen @ghislaineguerin can you help me remember the conclusion on this?

design/specs/table_inspector.md
rajatvijay: @kgodey @seancolsen @ghislaineguerin can you help me remember the conclusion on this?

design/specs/table_inspector.md
rajatvijay: @kgodey @seancolsen @ghislaineguerin can you help me remember the conclusion on this?

design/specs/table_inspector.md
seancolsen: For this cycle, the Table Inspector will only have tabs for "Table", "Column", and "Record". We may want to add a "Cell" tab later on, but for now I think it's safe to ignore the Cell tab entirely. We can pick up the design work later when/if needed.

design/specs/table_inspector.md
kgodey: I believe the conclusion was that we would not have tabs within tabs, we would show both in the inspector at the same time.

kgodey: @ghislaineguerin I assume you'll need to create new wireframes for the table inspector to address the editing issue, maybe you can put this in too?

design/specs/table_inspector.md
kgodey: We were going to remove this entirely, it will be addressed in the column extraction spec. https://github.com/centerofci/mathesar-wiki/pull/54

design/specs/table_inspector.md
kgodey: I think we were going to keep this and have a "delete record" action.

kgodey (approved):
I'm not going to re-review the content in detail, I trust @rajatvijay will make sure any decisions that are blocking him will be reflected in the spec. 

Since inline editing is being handled in a different spec (https://github.com/centerofci/mathesar-wiki/pull/55), the only changes I'd like to request here are:
- Please remove unused assets from the PR
- Please make sure the main specs page links to this one. I don't see any changes to the specs page in this PR.
- Please add a note saying that the editing portrayed in these wireframes is out of date and will be handled in a separate spec.

Once those are addressed, please go ahead and merge this @rajatvijay.

design/specs/table_inspector.md
ghislaineguerin: @kgodey Since this affected data import as well I think we need a separate issue to address only data type options.

design/specs/table_inspector.md
ghislaineguerin: @kgodey This no longer applies. It was relevant when the inspector states were dependent on current selection. I will remove it.

Issue #52, opened 2022-08-10
Add a spec for import data UX.


Dhruvi16:
We need a spec for for describing UX for importing data. 

This will also solve [redesigning the preview issue](https://github.com/centerofci/mathesar/issues/934).

PR #53, opened 2022-08-11
Update the import table spec

Dhruvi16:
This will address #52

### DIFF
--- a/design/specs/table-import.md
+++ b/design/specs/table-import.md
@@ -1,11 +1,11 @@
 ---
-title: Add Table from File Import Specs
+title: Add Table from Import/URI/Data paste Specs
 description: 
 published: true
 date: 2021-08-18T13:45:20.499Z
 tags: 
 editor: markdown
-dateCreated: 2021-07-01T21:03:22.324Z
+dateCreated: 2022-08-10T21:03:22.324Z
 ---
 
 # Context
@@ -14,153 +14,176 @@ Adding a new table from a file import allows users to input data into Mathesar w
 While data file import is a baseline feature for most applications in this category, additional functionality will be implemented in the future to enhance the data import process, such as automatic type detection, data cleansing, error detection, and others.
 
 # User Experience
-## Impact
-This feature will improve the user experience for users who need to input data into Mathesar by reducing the required effort when populating tables.
 
-This feature will also support the experience of first-time users, as it is expected by users of similar applications and would cause dissatisfaction if missing.
+## Navigating to the Data Import page
 
-## User Flow
+![](https://share.balsamiq.com/c/p3dqSXr66qa6EnPyykqpyZ.png)
 
-![](/assets/design/specs/table-import/Z3ngzR9.png)
+Data import navigation link will be found on the active schema page. Since we are only allowing users to import tables for now, positioning it with the existing tables/ active schema view made sense. 
 
+## Importing data into a new table
 
-### User Adds a Table
-The user starts the process of adding a table by clicking on any of the available options. The user can add a table from the sidebar menu(1) or the 'New' menu located on the top navigation bar in the main content area.
+![](https://share.balsamiq.com/c/w2AbsJLMY5Qwt4aGDuozLN.png)
 
-![](/assets/design/specs/table-import/DH3XRuj.png =400x)
+### Defining the New Table
 
-#### Notes
-- We can implement the 'New' menu at a later iteration.
+When the user jumps to the upload view. They see a name input which can be added or not added. If not added by user the table will be given a default name and that can also be changed in the confirmation view.
 
-### User Sets Table Options
-Once the add table process starts, the user is prompted to set a name, confirm, or select the schema that will hold the table.
-Using the lookup field, the user could also add a new schema from this dialog.
+### Uploading the Data File
 
-![](/assets/design/specs/table-import/ByfztC0tu.png =400x)
+#### File Upload Methods
 
-#### Valid and Invalid Table Names
-Based on [PostgreSQL Documentation](https://www.postgresql.org/docs/13/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), some naming conventions apply to objects, including tables, views, and columns. For most cases, we'll handle any necessary quoting automatically and transparently and generate unique names when required by adding a number as the last character.
+There are three main blocks that specifies three ways to upload the file/data for import. Each block has some important description with it. Each active way is highlighted when in use.
 
-- You can use any character for the valid characters except for a literal `null` character in a name.
-- For starting characters, anything goes (except a literal null) for our use case with the same caveats w.r.t. exotic characters above.
+##### Local File
 
-### User Selects Add Table Method
-After confirming the table name and schema, the user must select one of the methods available for adding a table. 
+The local file on the disk can be uploaded by dragging to the block or clicking on the block. Once the user drag the file to the block or select the file from their local machine, they can see the processing view.
 
-![](/assets/design/specs/table-import/chcPbaV.png =320x)
+##### URI
 
-Implementation Notes:
-- Copy will be revised later with the purpose of supporting best practices while guiding the user during the table creation process. We want to be more explicit in regards to goals for every step. 
+User can paste a URI in this block. If the URI is valid, the user would be able to click on the continue button to move to the processing import view.
 
-#### Implementation Note:
-- Empty table flow will be designed in a future iteration.
+##### Paste
 
-### User Selects Data Import Method
-Once the new table is created, The system will prompt the user to start the import data process. There are various options to import data, whether from a local file, clipboard, or remote location.
-By default, Upload File is the selected option.
-
-![](/assets/design/specs/table-import/zU9Csfe.png)
+User can paste data directly here in this block. If the data pasted here is valid, the user would be able to click on the continue button to move to the processing import view.
 
 
-### User Uploads a File
-The user then proceeds to upload a CSV file, and once it's completed, the system will render the file along with an option to remove it and upload another file.
+#### Import Options
 
-![](/assets/design/specs/table-import/1FZ5Efd.png)
+There are a few flags that user will find helpful for the import process. And they are present on the upload view itself. 
 
-In most cases, when the file is large, processing will take longer. For that reason, a loading UI state is needed.
+##### Data Type Inference
 
-![](/assets/design/specs/table-import/By4JP0CK_.png)
+This is a checkbox input. So when the checkbox is checked, the data type inference will not happen while the import process. This would make the import process faster and in some cases, the process possible too. The user would then have to select the data types in the confirmation stage. 
 
-If any errors prevent the file from being uploaded, an error message will be displayed.
+If not checked, the data type will be detected and this might take a longer time and sometimes the import might not even happen. The user can still change the data types in the confirmation stage.
 
-![](/assets/design/specs/table-import/HJxJY0AY_.png)
+##### Skip Warnings
 
-#### Implementation Notes
-- Paste and From URL methods will be designed in a future iteration. The current use case only uses the file import method. 
+This is again a checkbox, if checked he user would not be notified with any warnings related to the import process. This way when there are situations, the defaults will be considered and the user won't be informed about it.
 
-### User Confirms Data
-If the system successfully processed the file, it will prompt the user to confirm if a preview of their data looks correct. The user can also choose to ignore columns, rename them, and change data types from this view. 
+##### Automatic Delimiter Overwrite
 
-![](/assets/design/specs/table-import/Rpz9ZhV.png)
+This is an input field where user can enter a deliminator used in their file. The default is taken as a comma(,). 
 
-### User Opens the New Table
-![](/assets/design/specs/table-import/JMtTnDZ.png)
+### Processing Uploaded File
+
+This is the stage where the uploaded file is processed. The user sees the progress, warnings and errors related to the file here. 
+
+#### Status Indicators
+
+On the processing screen there is a progress bar which indicates where the upload and import process is. There will also be small messages below the progress bar to keep user informed about the process/status.
+
+#### File Processing Errors and Troubleshooting
+
+![](https://share.balsamiq.com/c/mo3WasYJpME3uCDc9dmuaW.png)
+
+##### Invalid file type 
+
+The user cannot upload files other than CSV/TSV. They will have to reupload it in such case.
+
+##### Invalid header
+
+When there are invalid headers. User will get an option of ignoring the headers.
+
+There might also be a case where they have to reupload the file so then we wont be showing ignore headers modal to them.
+
+##### Missing row data
+
+When the row data is missing, the message consists of row number. The user might get the option of skipping that row and move forward.
+
+In some cases the user might have to upload the file again.
+
+##### File too large
+
+The data inference of some files might take too long so there is an option for users to skip it.
+
+But sometimes the memory just runs out so the user would have to reupload a new file.
+
+##### File is corrupt
+
+No particular insightful messages. User will have to reupload new valid file.
+
+##### Date type error (Format errors)
+
+Q - Not very sure about this.
+
+##### Invalid data paste
+
+##### Invalid input in the confirmation state of a record
+Q - Is there an validation for data input in the table confirmation
+
+As much as I have seen it in staging. The iput validation will be similar as table view. And if the user confirms the table with erroneous records, the values will be set to null.
+
+#### Abandoning the File Processing Step
+
+![](https://share.balsamiq.com/c/m6Ui2naMGZXijJgjcaF15f.png)
+
+Since technically background/parallel imports are difficult, we wont be supporting in the user journey as well. When the user tries to move away there is a warning modal that warns them that they would lose the progress and they would land to the first screen if they land back to the import. 
+
+### Confirming Imported Data
+
+![](https://share.balsamiq.com/c/ds5qgkUM5bhGBWvkucdWnT.png)
+
+When the import is successful, the user lands on the active confirmation view.
+
+#### Go back button
+
+ This will take them to import screen. So before that there will be a warning modal which is listed in [Behavior section]. If the table gets confirmed it is added to the schema or else its discarded.
+
+#### Selected file section
+
+This consists of the file the user uploaded. They can access it from here when they click on it. If the reupload button is selected it has a similar behavior like go back button.
+
+#### Table name
+
+The input has a default name that can be changed by user there itself.
+
+#### Import Summary (Total Records, Columns)
+
+This has a small summary like number of rows, records etc. The missing data link will take them to the first row that has missing data(Not sure that is possible technically TBD). Similarly the duplicate column link takes the user to the first duplicate column(Not sure that is possible technically TBD).
+
+#### Options
+
+- Use headers - This will use the first row as headers if selected or else there will be default column names. This will remove duplicate column scenario.
+- Set missing data as null - This will set missing data as null and remove missing data scenario.
+
+#### Row limit for Preview
 
 
-## Feedback
-- Does the impact and user experience for this functionality help users achieve the goals established in the [Inventory Use Case](/design/exploration/use-cases/inventory-use-case)?
-- Is there an opportunity to allow a different user experience or workflow? Why or why not?
+#### Verifying and Updating Data Types
 
-# Interactions (WIP)
-Interactions for adding a table via file import focus on guiding the users through multiple steps, providing cues to help the user make decisions, and preventing unwanted outcomes by incorporating confirmation steps in error-prone steps.
+This is similar to the data options menu for the table view. But there are small differences to remove Cognitive load on user -
 
-## Wizards
-Wizards are series of screens that guide users through multi-step processes.
-In adding a table via file import, wizards are used to accomplishing some required tasks, such as selecting the data source to use (e.g., upload a file, paste, URL), mapping columns, and confirming field types.
+1. Removal of options like filtering, sorting etc since it might not be that useful to user at this point.
+2. Removal of options which are related to adding new data like set default value. Since we are not allowing user to add data at this stage keeping these options can be avoided. However all the options related to the data type and data format will be intact since this might be a deciding factor. 
 
-![](/assets/design/specs/table-import/ByYAYLmq_.png)
+Q - Are we allowing user to add column or a record at this stage? If not are we gonna do that in future?
 
-The wizard elements should provide users with information regarding the length of the process, i.e., the number of steps. It should also provide controls to cancel the operation.
+##### Allowed Data Type Changes
 
-![](/assets/design/specs/table-import/Bk1x58Q9O.png)
+The data type changes behavior will be similar to that of table data type changes. User will be able to switch to compatible data types so they do not lose the data.
 
+##### Revert Data Type Change
 
-## Pre-Determined Options
-When triggered from a particular context, actions like adding a new table will have pre-determined options, i.e., the target schema for a new table.
+User can change the data types to the other compatible ones at any moment in the confirmation stage. 
 
-For example, if calling the add new table function from the sidebar menu, the context will determine that the system must add the table to that particular schema. 
+#### Exclude Columns from Import
 
-However, a user might choose to add a new table from a different context, where there's no pre-determined option. In this case, a user would have to select the desired option. 
+Each column has a checkmark on the side. If selected the table will have the column or else it will be excluded from the table. 
 
-Making pre-determined options visible and editable in the interface can be helpful for users when confirming outcomes where contextual cues are missing.
+The duplicate columns will be unselected by default. To select them, the user will have to rename them. Until then the checkbox will be grayed out. 
 
-# User Interface (WIP)
-## Components
-### Input Controls
-Input controls allow the user to enter or select different values needed to perform tasks within Mathesar. In file import, they will enable the user to specify column types and parameters required to map columns.
+#### Abandoning the Confirmation Step
 
-#### Text input
-![](/assets/design/specs/table-import/rJoUmDX5_.png =140x)
+![](https://share.balsamiq.com/c/ajPCXuhshEjRMvJXMLb5Xu.png)
 
-#### Select input
-![](/assets/design/specs/table-import/rkavQwXc_.png =140x)
+Handling unconfirmed table in different scenarios might be difficult so the solution I propose for it is blocking the user to navigate away before confirming or dumping the import.
 
-#### Lookup input
-![](/assets/design/specs/table-import/BysumDQqu.png =140x)
-
-#### Radio buttons
-![](/assets/design/specs/table-import/H1P5XPQcu.png =320x)
-
-#### Checkbox input
-![](/assets/design/specs/table-import/HJnhQP7cd.png =200x)
-
-#### States
-- Default
-- On Focus
-- Disabled
-
-### Buttons
-Buttons allow users to perform actions and obtain a result from the system.
-
-#### Types
-There are different types of buttons, and their use depends on several factors, e.g., the type of task they are to perform and their context.
-
-![](/assets/design/specs/table-import/rkEwqImqd.png =200x)
-
-Primary buttons are used for actionable (e.g., submit content, apply a change) and some exceptional navigational purposes (e.g., continue to a next step). However, their use should be restricted to priority or practical actions, with the ideal number of primary buttons being one per view.
-
-Secondary buttons are used alone or in combination with primary controls to complement priority actions.
+## UI Components
 
 
-### Dialogs
-Dialogs are a type of modal window that appears in front of content and is used when the system requires users to make decisions.
-When active, dialogs will prevent other content from being interacted with. In most cases, they are driven by user action, and their unprompted use is discouraged.
-
-![](/assets/design/specs/table-import/Hk02qLQc_.png =280x)
-
-### System Notifications
-
-#### Error Messages
-![](/assets/design/specs/table-import/SJBTzvX9d.png)
+### File Upload Method Selector
 
 
+#### Toggle between methods
--- a/design/specs/table-import.md
+++ b/design/specs/table-import.md
@@ -104,16 +104,21 @@ But sometimes the memory just runs out so the user would have to reupload a new 
 
 No particular insightful messages. User will have to reupload new valid file.
 
-##### Date type error (Format errors)
+##### Format error
 
-Q - Not very sure about this.
+The column number might not match the number of header or there might be other formatting errors and in that case the user will have to reupload the file.
 
 ##### Invalid data paste
 
-##### Invalid input in the confirmation state of a record
-Q - Is there an validation for data input in the table confirmation
+The data format is not correct, hence the user will have to edit or paste the data again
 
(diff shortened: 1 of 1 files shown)


Dhruvi16:
Before reviewing I have some questions to discuss -

1. What will be the format errors specifically? We discussed about some date errors with Brent, I am not so sure about them @ghislaineguerin
2. What will be the errors related to the data paste and URI paste while inputting?
3. Will there be any difference in validation of change in data of the table at the confirmation stage?
4.  Are we allowing user to add column or a record at confirmation stage? If not are we gonna do that in future?
5. Will the navigation link of data import be visible when the user is inside the table view? 
6. For duplicate columns, the UX in my mind would be not allowing user to select them before they change the name of the header. Is that an appropriate take?

ghislaineguerin:
> 1. What will be the format errors specifically? We discussed about some date errors with Brent, I am not so sure about them @ghislaineguerin

- Improperly formatted could be due to mismatched number of headers and columns count

> 2. What will be the errors related to the data paste and URI paste while inputting?

- The URI points to a missing resource / broken link
- The URI text is in the correct form to potentially be a legitimate URI

> 3. Will there be any difference in validation of change in data of the table at the confirmation stage?

- Not at this point

> 4. Are we allowing user to add column or a record at confirmation stage? If not are we gonna do that in future?

- Not during the import step. We won't do this in the future.

> 5. Will the navigation link of data import be visible when the user is inside the table view?

- The user should always see their current location. We should ensure consistency with the latest navigation updates.

> 6. For duplicate columns, the UX in my mind would be not allowing user to select them before they change the name of the header. Is that an appropriate take?

- All columns are pre-selected by default. I'm assuming we'd show warnings in the duplicates and prevent user from completing the import process unless changed.

ghislaineguerin:
From our call: 

- Update file upload UI
- Update file upload UI with the warnings in context
- Check if abandoned import process can also delete the created table
- Add scenario where table confirmation was force-closed
- In confirmation step, ensure users are aware of selected columns and warnings that are hidden by the scroll
- Remove option to rename table from confirmation step
- Add note dropdown vs. table inspector for data type options

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
 While data file import is a baseline feature for most applications in this category, additional functionality will be implemented in the future to enhance the data import process, such as automatic type detection, data cleansing, error detection, and others.
```
ghislaineguerin, line 14: I think we can remove this part as we are adding type detection now and will handle other tasks separatedly.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
 
 # User Experience
```
ghislaineguerin, line 16: We can only have one h1 heading in the document. Please update.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
-This feature will improve the user experience for users who need to input data into Mathesar by reducing the required effort when populating tables.
 
-This feature will also support the experience of first-time users, as it is expected by users of similar applications and would cause dissatisfaction if missing.
+## Navigating to the Data Import page
 
-## User Flow
+![](https://share.balsamiq.com/c/p3dqSXr66qa6EnPyykqpyZ.png)
 
-![](/assets/design/specs/table-import/Z3ngzR9.png)
+Data import navigation link will be found on the active schema page. Since we are only allowing users to import tables for now, positioning it with the existing tables/ active schema view made sense.
```
ghislaineguerin, line 22: It would be enough to mention that user can access data import from the Schema page, since there will be a separate spec to address the design and layout for this page.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
-### User Adds a Table
-The user starts the process of adding a table by clicking on any of the available options. The user can add a table from the sidebar menu(1) or the 'New' menu located on the top navigation bar in the main content area.
+![](https://share.balsamiq.com/c/jyszUUBfSvoCqd6qWQcz3G.png)
 
-![](/assets/design/specs/table-import/DH3XRuj.png =400x)
+### Defining the New Table
 
-#### Notes
-- We can implement the 'New' menu at a later iteration.
+When the user jumps to the upload view. They see a name input which can be added or not added. If not added by user the table will be given a default name and that can also be changed after the import is confirmed.
```
ghislaineguerin, line 30: I have discussed this with Pavish and the current backend won't handle renaming the table before it is created. We will have to move table rename to the confirmation step instead.

ghislaineguerin (changes requested):
Thanks Dhruvi, looks good. I've added my feedback.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
+#### Import Summary (Total Records, Columns)
+
+This has a small summary like number of rows, records etc. The number of missing data and duplicated columns is present on the confirmation view. Although user can ignore those warnings and table can be confirmed as is.
+
+#### Options
+
+- Use headers - This will use the first row as headers if selected or else there will be default column names. This will remove duplicate column scenario.
+- Set missing data as null - This will set missing data as null and remove missing data scenario.
+
+#### Verifying and Updating Data Types
```
kgodey, line 144: This was the main problem that this spec needed to solve, we need more detail on the data type changing menu. 

Also, we're trying to have visual consistency with the table here, but I don't see anything about that in the spec.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
+
+*Note - The UX of changing data types depends on table inspector designs.*
```
kgodey, line 146: What does this mean? How does it depend on it?

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
+
+This is similar to the data options menu for the table view. But there are small differences to remove Cognitive load on user -
```
kgodey, line 148: Please provide more detailed specs and wireframes for this.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
+1. Removal of options like filtering, sorting etc since it might not be that useful to user at this point.
+2. Removal of options which are related to adding new data like set default value. Since we are not allowing user to add data at this stage keeping these options can be avoided. However all the options related to the data type and data format will be intact since this might be a deciding factor. 
+
+##### Allowed Data Type Changes
+
+The data type changes behavior will be similar to that of table data type changes. User will be able to switch to compatible data types so they do not lose the data.
+
+##### Revert Data Type Change
+
+User can change the data types to the other compatible ones at any moment in the confirmation stage.
```
kgodey, line 159: I'm not sure if I would call this reversion, it's just setting the data type.

design/specs/table-import.md
kgodey: What about display options?

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
+
+All the columns will be selected by default including the duplicated columns. The duplicate columns will be highlighted with a warning though.
+
+#### Abandoning the Confirmation Step
+
+![](https://share.balsamiq.com/c/ajPCXuhshEjRMvJXMLb5Xu.png)
+
+Handling unconfirmed table in different scenarios might be difficult so the user will no be allowed to navigate away without interacting with a modal. 
+
+In the case of force close, the table listed on the schema page will be highlighted with a warning. Any attempt to use an unconfirmed table will lead user to a modal to either confirm the table as is or go to the confirmation view.
```
kgodey, line 173: How will it be highlighted? Also, this is inconsistent with other parts of the spec about the table getting deleted if not confirmed (but this one is correct).

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
+
+## UI Components
+
+### File Upload Method Selector
+
+![](https://share.balsamiq.com/c/jyszUUBfSvoCqd6qWQcz3G.png)
+
+To select the method to upload the file there will be vertical tabs to select the method. There will be a rectangular space attached to the tabs. This space will be dedicated to the upload method selected.
```
kgodey, line 181: Please see note about vertical tabs above – they don't make sense for a mutually exclusive choice.

Dhruvi16:
> This spec needs a lot of work. There's a lot of things here that can't be implemented quickly and don't make any sense for the demo use case. There's very little detail on the actual problem that we need design work on – visual consistency with tables and the interface for changing data types and display options.

Alright. For the data types and the display options, I had proposed a similar data type selector as we have now in the staging since it is consistent but @ghislaineguerin pointed out that we were ripping it off so there is pending discussion yet to happen on it. So, @ghislaineguerin suggested to put a note on it and we can decide it when we have table inspector.

For visual consistency, I am not sure about it. Will have to look into it.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
-Once the add table process starts, the user is prompted to set a name, confirm, or select the schema that will hold the table.
-Using the lookup field, the user could also add a new schema from this dialog.
+### Uploading the Data File
 
-![](/assets/design/specs/table-import/ByfztC0tu.png =400x)
+#### File Upload Methods
 
-#### Valid and Invalid Table Names
-Based on [PostgreSQL Documentation](https://www.postgresql.org/docs/13/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), some naming conventions apply to objects, including tables, views, and columns. For most cases, we'll handle any necessary quoting automatically and transparently and generate unique names when required by adding a number as the last character.
+There are three main blocks that specifies three ways to upload the file/data for import. Each block has some important description with it. Each active way is highlighted when in use.
```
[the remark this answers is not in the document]
Dhruvi16: I think I disagree here. Tabs are also used for navigation. Plus here the user is accomplishing the same task but has different choices. And tabs are one of the most scannable components out there. So I am not sure how users are gonna miss them.

The current UI that is the radio buttons are bit more confusing since its a type of an input and not something user would perceive as a navigating tool.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
-![](/assets/design/specs/table-import/zU9Csfe.png)
 
+#### Import Options
 
-### User Uploads a File
-The user then proceeds to upload a CSV file, and once it's completed, the system will render the file along with an option to remove it and upload another file.
+There are a few flags that user will find helpful for the import process. And they are present on the upload view itself. 
 
-![](/assets/design/specs/table-import/1FZ5Efd.png)
+##### Data Type Inference
```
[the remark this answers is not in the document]
Dhruvi16: Okay I agree with this. This would put more cognitive load seeing our user persona. And we do not need this on the first screen since user is handling import here so we should let them focus on that rather than worrying about future errors caused by data inference.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
 
-In most cases, when the file is large, processing will take longer. For that reason, a loading UI state is needed.
+This is a checkbox input. So when the checkbox is checked, the data type inference will not happen while the import process. This would make the import process faster and in some cases, the process possible too. The user would then have to select the data types in the confirmation stage. 
 
-![](/assets/design/specs/table-import/By4JP0CK_.png)
+If not checked, the data type will be detected and this might take a longer time and sometimes the import might not even happen. The user can still change the data types in the confirmation stage.
```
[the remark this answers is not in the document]
Dhruvi16: In one of the discussions with @mathemancer and @ghislaineguerin we did discuss the probability of import getting stuck if the data inference is taking too much time. Simply opting out might let the import happen right?

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
 
-If any errors prevent the file from being uploaded, an error message will be displayed.
+##### Skip Warnings
 
-![](/assets/design/specs/table-import/HJxJY0AY_.png)
+This is again a checkbox, if checked he user would not be notified with any warnings related to the import process. This way when there are situations, the defaults will be considered and the user won't be informed about it.
```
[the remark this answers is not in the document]
Dhruvi16: Yeah I agree with this too. I was too skeptical about this. I believe we do not need this. Again the same comment as data inference.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
 
-#### Implementation Notes
-- Paste and From URL methods will be designed in a future iteration. The current use case only uses the file import method. 
+##### Automatic Delimiter Overwrite
 
-### User Confirms Data
-If the system successfully processed the file, it will prompt the user to confirm if a preview of their data looks correct. The user can also choose to ignore columns, rename them, and change data types from this view. 
+This is an input field where user can enter a deliminator used in their file. The default is taken as a comma(,).
```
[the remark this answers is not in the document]
Dhruvi16: As per a discussion with @ghislaineguerin, we discussed that sometimes there might be JSON data which might confuse the algorithm and hence the file would not be imported. So that is why I thought this might be a better way to handle it.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
-Interactions for adding a table via file import focus on guiding the users through multiple steps, providing cues to help the user make decisions, and preventing unwanted outcomes by incorporating confirmation steps in error-prone steps.
+#### File Processing Errors and Troubleshooting
 
-## Wizards
-Wizards are series of screens that guide users through multi-step processes.
-In adding a table via file import, wizards are used to accomplishing some required tasks, such as selecting the data source to use (e.g., upload a file, paste, URL), mapping columns, and confirming field types.
+![](https://share.balsamiq.com/c/mo3WasYJpME3uCDc9dmuaW.png)
 
-![](/assets/design/specs/table-import/ByYAYLmq_.png)
+##### Invalid file type
```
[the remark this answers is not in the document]
Dhruvi16: Oh okay. But I think it is good to have since we have everything here regarding the import.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
-### Buttons
-Buttons allow users to perform actions and obtain a result from the system.
+#### Abandoning the File Processing Step
 
-#### Types
-There are different types of buttons, and their use depends on several factors, e.g., the type of task they are to perform and their context.
+![](https://share.balsamiq.com/c/m6Ui2naMGZXijJgjcaF15f.png)
 
-![](/assets/design/specs/table-import/rkEwqImqd.png =200x)
+Since technically background/parallel imports are difficult, we won't be supporting that in the user journey as well. When the user tries to move away there is a warning modal that warns them that they would lose the progress and they would land to the first screen if they land back to the import.
```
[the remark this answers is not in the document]
Dhruvi16: Alright. Got it.

design/specs/table-import.md
```diff
@@ -14,153 +14,175 @@ Adding a new table from a file import allows users to input data into Mathesar w
...
 
-#### Types
-There are different types of buttons, and their use depends on several factors, e.g., the type of task they are to perform and their context.
+![](https://share.balsamiq.com/c/m6Ui2naMGZXijJgjcaF15f.png)
 
-![](/assets/design/specs/table-import/rkEwqImqd.png =200x)
+Since technically background/parallel imports are difficult, we won't be supporting that in the user journey as well. When the user tries to move away there is a warning modal that warns them that they would lose the progress and they would land to the first screen if they land back to the import. 
 
-Primary buttons are used for actionable (e.g., submit content, apply a change) and some exceptional navigational purposes (e.g., continue to a next step). However, their use should be restricted to priority or practical actions, with the ideal number of primary buttons being one per view.
+### Confirming Imported Data
```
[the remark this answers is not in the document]
Dhruvi16: Yep. I agree. Will make those changes.

design/specs/table-import.md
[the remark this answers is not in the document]
kgodey: > Plus here the user is accomplishing the same task but has different choices. 

This is my point exactly, the user has to make a single choice of import method here and we need to make it clear that they have three choices and they can pick one. I don't think we need to use the radio buttons (although they do fulfill that purpose) but we need something that shows all of the options in a visually prominent way. With vertical tabs, it's easy for people to focus on the center of the page and miss that there are other options on the side.

design/specs/table-import.md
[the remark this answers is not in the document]
kgodey: This is probably going to confuse our user persona (same as my comments above). It seems like showing an error for un-importable files would work fine.

kgodey:
@Dhruvi16 Please also take a look at https://github.com/centerofci/mathesar/issues/934, this was the original issue that described the problems the new design needed to solve. 

Also, the table inspector spec https://github.com/centerofci/mathesar-wiki/pull/51 and the inline editing spec https://github.com/centerofci/mathesar-wiki/pull/55 are pretty much done, even though they're not merged. You should be able to reference them enough to make the decisions you need.

design/specs/table-import.md
Dhruvi16: I did not get this, could we discuss this please?

design/specs/table-import.md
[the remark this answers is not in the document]
Dhruvi16: temp
I am not sure how vertical tabs are not visible enough. This is one example of tabs used for import file itself. We can I guess play around with the size of tabs and move the tabs inside the rectangle so they are visible enough?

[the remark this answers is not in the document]
kgodey: Those are more visible and a design like that would resolve my concerns about visibility. It still comes across as a navigation pattern, and not a pattern related to the the user needing to pick a single option out of a few available choices.

Dhruvi16:
So I think I need an approval on UI of the upload first so I can make other changes. I am figuring out data type change at the confirmation stage parallelly. But yeah need to discuss the UI here. 

Option 1 -


So this is again vertical tabs that I have seen widely across all the platforms that provides options for importing file. I have made tabs very prominent so they are visible and included them in the square. 

Option 2 -


Here I have made three options as action cards where the user will click on one of them and followed by it will be a modal where they upload the file or paste URI or Paste data. The progress and errors will be in the same modal. 

One very big advantage of this is that while progress the user wont be able to click anywhere. And as Kriti wanted the journeys will be mutually exclusive. 

@kgodey @ghislaineguerin @pavish please lemme know if you have any comments

kgodey:
@Dhruvi16 Before I make up my mind, I have a question. Why does Option 2 need a modal? It seems like you could replace the three action cards with the file upload interface since the cards are no longer relevant.

Dhruvi16:
> @Dhruvi16 Before I make up my mind, I have a question. Why does Option 2 need a modal? It seems like you could replace the three action cards with the file upload interface since the cards are no longer relevant.

You mean rather than cards, I use Drag & drop, URI paste and Data paste inputs right? That was my initial idea but I do feel that would be very cluttered since visually the inputs are not consistent plus for URI and data paste we need a continue button when the user pastes something. 
Modal gives a very clear and independent journey for each upload method I felt that is why I proposed modal rather that putting all the inputs on the screen.

kgodey:
@Dhruvi16 This is what I mean:

Dhruvi16:
> @Dhruvi16 This is what I mean: 
> 
> Once the user selects a method, then the action cards disappear entirely and is replaced by the input. You may want to provide a way of going back to selecting a different action method if needed.

Oh okay! I think this is a better idea. I think then we can create three steps rather than 2. Like Select upload method -> Upload data -> Confirm data.

So the user can go back forth the first two steps.

kgodey:
> Oh okay! I think this is a better idea. I think then we can create three steps rather than 2. Like Select upload method -> Upload data -> Confirm data.

Yeah, that sounds good to me. I like this option the best because all the options are visually prominent and the other options go away once they've made their choice, which reduces cognitive load.

Dhruvi16:
> > Oh okay! I think this is a better idea. I think then we can create three steps rather than 2. Like Select upload method -> Upload data -> Confirm data.
> 
> Yeah, that sounds good to me. I like this option the best because all the options are visually prominent and the other options go away once they've made their choice, which reduces cognitive load.

Alright I will update the spec with this UI then.

kgodey:
> Alright I will update the spec with this UI then.

@ghislaineguerin probably also has an opinion. I think it's fine to update the spec if it unblocks you for now, as long as you're still expecting feedback.

Dhruvi16:
@ghislaineguerin I have updated the spec. Please take a look. Also Could you please tell me where can I see UI for schema page? I need to update the spec with unconfirmed table part.

ghislaineguerin:
I think we should consider how the user will interact with the data import screen.
After their first interaction with Data Import, it is likely that the user will be familiar with the different upload methods, or even have a preference for one of them. I worry about making all the methods accesible exclusively in the second step.

### Proposed Flow

In the same view the user should be able to drag a file, or click on the browse button to select a file. They should also be able to use a URL or paste data.

The file type information should be global, and not specific to the upload method. At the top of the upload controls we should have the following:

- A title for the section "Upload the data file"
- The detailed list of supported file types (e.g. CSV, TSV, etc.)

The upload controls favor the drag & drop method over the others. I'm assuming that the user will prefer the drag & drop method most of the time. It's the only one that allows for multiple file upload once available, so it's a good bet for future improvements.

For the paste and URL options we can open a modal to provide an input. I know that usually we want to have all options happen in the same window, but even for file selection, the user will still deal with the systems file selection dialog. It is also a pattern we have in other parts of the application, for example when naming a new schema.

Here's the proposed UI for the upload controls:

image

Dhruvi16:
@ghislaineguerin I do think the UI is a bit cluttered here. And there are inconsistent user journeys for each upload method which I think can be confusing. Lemme know what you think

ghislaineguerin:
@Dhruvi16 I get what you are saying.
Our present design includes radio buttons for each upload method. Changing the radio button would bring up the relevant input control, which is a common way to display these. This is similar to your first proposed idea with the tabs.
Our thinking with having the different cards was to give access to all controls so that the user wouldn't have to navigate to each, but having an additional step seems harder to implement and doesn't solve the original problem.

I'm ok with going back to radio buttons as the original design if we find that the updates add unnecessary complexity.

ghislaineguerin:
If we were to consider a solution that doesn't involve radio buttons I think https://uppy.io/ does it well.
- It combines the drop area with the other upload options
- It provides a way to replace all content with the desired upload control

Dhruvi16:
If we are discussing to retain radio buttons, I would like to point some thoughts -
1. Radio buttons are preferably not used for changing views. [Radio buttons](https://material.io/components/radio-buttons)
2. Radio buttons are majorly used as form components. 
3. Tabs have always meant switching between different views. You can see the [design guidelines](https://material.io/components/tabs)

Our requirements -
1. Parallel navigation between views.
2. Only one view is visible at a time so choices are mutually exclusive.
3. The choices should be clear and prominent.

Dhruvi16:
But if we feel we should stick to radio buttons I will change the spec UI to radio buttons. But I personally feel that we should once consider Tabs. @ghislaineguerin would like to know your thoughts. Attaching platforms that use it -
temp
image

ghislaineguerin:
@kgodey @Dhruvi16 
I agree with keeping the radio buttons, just wanted to clarify where the requirement came from.
Originally, we discussed allowing users to create their first table directly from an empty schema page (no tables), thus we needed a way to present the different upload methods and discussed the cards layout (a card for each upload method with a description). However, since we opted to have a dedicated Data Import view, this requirement seems unnecessary.

We can consider upgrading the radio buttons once we have more flows that require uploading of files.

pavish:
@Dhruvi16 

> Radio buttons are majorly used as form components.

The import method section is a form input. We are asking for the user's input on which method they would like to import data into Mathesar.

Based on the chosen input, we are displaying more form content. We are not switching views.

> Tabs have always meant switching between different views

Yes, but tabs are not meant for making a choice, which is what we require. They are only meant for navigating across independent entities.

I do not see the problem that tabs solve over the radio buttons. If we are talking about making the options more prominent, we may want to focus on that angle instead. 

@ghislaineguerin 

> Our thinking with having the different cards was to give access to all controls so that the user wouldn't have to navigate to each, but having an additional step seems harder to implement and doesn't solve the original problem.

The user will still have to perform an additional step even if we display all of the options together. For eg., in your [option 1](https://github.com/centerofci/mathesar-wiki/pull/53#issuecomment-1229842127), the user will still have to click on their preferred method before we show the related content. I believe option 2 would result in a cognitive overload for the user.

I still do not understand why the original problem that's mentioned here is a problem.

I believe the only issue we have in this stage of import is providing more context to the user helping them choose an option.

Dhruvi16:
I agree with the facts @ghislaineguerin & @pavish stated. So should I go ahead and change the UI again to radio buttons?

ghislaineguerin:
@Dhruvi16 yes, please. Also suggest anything that might help provide more context to users as @pavish mentioned.

design/specs/table-import.md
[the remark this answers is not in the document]
mathemancer: Sorry, I'm pretty late to this convo; just catching up. It's possible to choose the incorrect delimiter and still have a valid import set up.

Imagine (if you will) a headerless TSV whose correct delimiter is "tab". Now, further suppose there is a JSON column in this TSV, with a fixed schema. In particular, it has the same number of commas on each row. Our algo will not be able to choose properly which delimiter to use, and in fact, will chose comma in that case. We could change that specific behavior (i.e., put tab above comma in the choice hierarchy), but I don't think that solves the problem in general.

I think we _should_ display and give the option to modify the delimiter which has been chosen by the algorithm, but we shouldn't assume it's a comma unless that's what's returned by the algo. I think people who use spreadsheets regularly should be okay with that, since it's a common step for importing data into spreadsheet software.

@kgodey Are you arguing that they shouldn't have a chance to change the delimiter, or am I misunderstanding?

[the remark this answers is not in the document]
mathemancer: Okay, now I've taken a moment to read through the document and see this in context. I agree with @kgodey that the user shouldn't have to make this choice at the upload/initial stage. I think it's more appropriate to give the chance to modify at the confirmation stage if something has gone wrong.

mathemancer:
> The file type information should be global, and not specific to the upload method. At the top of the upload controls we should have the following:
> 
> * A title for the section "Upload the data file"
> * The detailed list of supported file types (e.g. CSV, TSV, etc.)
> 
> The upload controls favor the drag & drop method over the others. I'm assuming that the user will prefer the drag & drop method most of the time. It's the only one that allows for multiple file upload once available, so it's a good bet for future improvements.

@ghislaineguerin 

A couple of technical points: I don't think we should filter files on suffix. While we can show some info about what we support, it might confuse users. In the wild, there are many (many) TSVs with a `.txt` suffix or no suffix at all, and we should make sure the user understands that this isn't relevant for how the file is processed. All import processing is based on the actual data in the file, not its filetype suffix or other metadata.

Maybe some users prefer drag and drop, but I definitely prefer the file browser. Also, I'm pretty used to being able to open (or import in our case) multiple files via a file browser (with shift-click or ctrl-click for a range or individual selections, respectively).

mathemancer:
@Dhruvi16 Have the comments about the delimiter overriding been addressed? It seems like the spec is still implying that the user needs to deal with this before importing.

design/specs/table-import.md
```diff
@@ -1,166 +1,107 @@
...
-### System Notifications
-
-#### Error Messages
-![](/assets/design/specs/table-import/SJBTzvX9d.png)
+If the table is unconfirmed, there will be a warning icon on the side of table listing which when hovered on will have a confirm warning. When clicked on the table listing the user will be taken to confirmation stage.
 
+### Usage of an unconfirmed table
+![](https://share.balsamiq.com/c/kLkhvgR9Uk1njGbmgh9msB.png)
 
+If the user tries to use an unconfirmed table, there will be a modal for confirming the table where the user can directly confirm the table or go to the confirmation view. They won't be able to use it without confirming it.
```
ghislaineguerin, line 107: It is not clear to me what 'View Table' will do. Will this force confirmation? If so, can we update the action to reflect that?

design/specs/table-import.md
```diff
@@ -1,166 +1,107 @@
...
+Since multiple scrolls on the screen is a bad UX. I propose limiting table's height to a certain percentage of the screen height. The percentage might be more clearer when we include styling part here. But as far as I think 25-35% of screen height for the table would suffice. 
 
-#### Implementation Notes
-- Paste and From URL methods will be designed in a future iteration. The current use case only uses the file import method. 
+#### Changing data type
+![](https://share.balsamiq.com/c/agFL3gk3gc7pkgjc2GUcw3.png)
 
-### User Confirms Data
-If the system successfully processed the file, it will prompt the user to confirm if a preview of their data looks correct. The user can also choose to ignore columns, rename them, and change data types from this view. 
+When the settings beside the data types is clicked, the user is presented with a choice of compatible data types. The compatible data types will be similar to the table view. The data options will not be available at this point.
```
ghislaineguerin, line 84: What do you mean by no data options available?

design/specs/table-import.md
```diff
@@ -1,166 +1,107 @@
...
+![](https://share.balsamiq.com/c/r7fmqfwG2hmjTTbwvMRehr.png)
 
-### User Sets Table Options
-Once the add table process starts, the user is prompted to set a name, confirm, or select the schema that will hold the table.
-Using the lookup field, the user could also add a new schema from this dialog.
+Once the user selects this upload method, the user gets an input text block for URI where they can paste valid URI and click on continue to upload the file.
 
-![](/assets/design/specs/table-import/ByfztC0tu.png =400x)
+#### Copy & paste data
+![](https://share.balsamiq.com/c/uC1C6dxPCze6oiHHTAkAvt.png)
```
ghislaineguerin, line 41: The drop area shouldn't be visible at this point right? (the dashed border)

ghislaineguerin (commented):
Thanks @Dhruvi16 the updates look good. I've added some notes/questions.

kgodey:
I will leave the next pass of review to @ghislaineguerin and @pavish. Please let me know once it looks good to both of you and I'll do a final pass from a product perspective.

pavish (approved):
I think the spec is good to merge. We will not be fully implementing this and we would probably need further updates when we are ready to implement it.

I've added a note in https://github.com/centerofci/mathesar-wiki/pull/53/commits/2e6ccb2a2250b3d947afb35f8e427eac0bdbe2b7 to archive the spec for now.

For the demo video & the live demo, the design deliverable would directly be the frontend PR which can be reviewed by @ghislaineguerin and @kgodey .

kgodey (approved):

kgodey:
Approved on my end, since it still says "changes requested" by @ghislaineguerin, I'll leave it up to her to merge this spec.

ghislaineguerin (approved):

PR #54, opened 2022-08-17
Column Extraction

ghislaineguerin:
This PR adds further information to the Column Extraction spec on the following:
- Table Inspector modifications
- Column Extraction Dialog Design


design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
+- Help users understand how the tables are linked after the split operation
 
-1. **The user imports a table from a file**
+## Splitting a Table by Extracting Columns
 
-    This user flow assumes that the starting point is an imported table containing columns that can be extracted and turned into entities. A column that may be transformed into a table or entity will usually include other columns that are attributes for that entity. If the database is not normalized, the entity to be extracted might include duplicated values. For example, consider a table of publications with multiple entries for the same publisher in the publisher column.
+### Selecting columns as a starting point
 
-2. **The user selects the columns (attributes) for the new entity**
+The user will select one or multiple columns to start with the column extraction process. All columns can be extracted, except for primary key columns. Once selected, the inspector panel will list 'Extract Columns to Table' as an action. Selecting this option will open the 'Extract Columns to Table' dialog.
```
kgodey, line 27: I would recommend calling this "Move Columns to Another Table" instead of "Extract Columns to Table", it's more clear.

Also, I assume we want the singular "Column" ("Move Column to Another Table") if there's only one column selected, I think it's worth specifying this.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
 
-    The user chooses all the columns that will be attributes for the new object. For example, in the instance of the publisher, there could be columns for 'Publisher Name' and 'Publisher Id.' Including all the attributes is very important since columns cannot be moved after the fact.
+![image](/assets/design/specs/column-extraction/185139636-9a8a048c-9f87-4e26-aa2c-88f4e8d1648d.png)
 
-     <img width="1512" alt="image" src="https://user-images.githubusercontent.com/845767/177544077-9f074830-ce64-4ed9-80bc-53de53977f9c.png">
+## Extract Columns Dialog
 
+The dialog will list the columns that will be extracted and the table that will be created. The user can change the name of the new table. The dialog will also list the foreign key that will be created between the new table and the original table. The user can change the name of the foreign key.
```
kgodey, line 33: This presumes that we will always move columns to a new table. We also support moving columns to an existing table (and we need this for the demo, please see Screen 15 in the [Storyboard](https://www.figma.com/file/F0FmNaNz8hvrgxPax3Bix1/Cycle-3-reference?node-id=0%3A1))

design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
 
-3. **The user clicks on the 'Extract Columns' option**
+![image](/assets/design/specs/column-extraction/185140900-ea1514c3-2423-42ad-ab98-a4f87848f463.png)
```
kgodey, line 35: We're using the term "foreign key" in the wireframe here, but we're referring to them as "links" elsewhere in the product. I think we should standardize on the term "link".

design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
-    The user can choose to keep copies of the columns in the table if needed. By default the extracted columns are deleted.
+### Impact of the Extract Columns Operation
 
-4. **The user inspects the new foreign key column (link)**
+Since the split operation will affect the original table, the user should be made aware of the operation's impact. The dialog will make the following information clear:
 
-    The new foreign key column will contain links to the new records that have been created. Additionally, when selected, the inspector panel will display the link properties.
+- The original table will be updated to remove the extracted columns
+- A new table will be created with the extracted columns
+- A foreign key will be created between the new table and the original table
```
kgodey, line 43: Please see note above about FK terminology.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
 
-    <img width="1469" alt="image" src="https://user-images.githubusercontent.com/845767/177546415-23b11ded-8987-4e69-ab1a-1decb9e773d3.png">
+### Link Column Name Convention
 
+To help users understand the role of the link column, the name of the link column will follow a convention that will be applied when users name the new table. The convention will be:
 
-5. **The user navigates to the new table**
+`table_name` `+` `Id`
```
kgodey, line 49: Can we please capitalize both letters? `ID`?

Also, what do we do if the name is already taken?

design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
-# Unresolved Design Problems
+Under the `Columns to Extract` section, the user can extend the original column selection by selecting additional columns. The user can also remove columns from the selection.
 
-## Extracting a foreign key column to create a mapping table
-If a user had a foreign key referencing a publisher in a publications table and wanted to link to more than one publisher, could they select both the table's primary key and that foreign key and create a new table from both? Should there be a special action that does this without adding a foreign key in place of the columns?
+This component could be reused for all actions involving column subsets.
 
-## Suggesting columns to be extracted
-Offering suggestions for columns that can be extracted into new tables would improve the feature's discoverability and provide users with a clear path to a better data model.
+Additionally, the user can rename the columns that will be extracted in cases where the column name is not descriptive enough.
```
kgodey, line 59: I'm not sure if the backend supports renaming the column. @silentninja would know. If we don't, then I suggest implementing this renaming feature in the future, we can make a ticket to track the work.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,58 @@ editor: markdown
...
-### Visual Cues
-WIP
-## Reordering Columns
-WIP
-## Cell-Only Selection
-WIP
\ No newline at end of file
+Once the user is happy with the changes, they can click on the `Extract Columns` button. The inspector panel will be toggled and the new link column will be selected in the original table. Under the `Columns` section, the user will see the new table listed as a link.
+
+![image](/assets/design/specs/column-extraction/185143860-f2995fc4-c767-4cce-aaa6-319ef27630ff.png)
```
kgodey, line 65: I don't think there's enough information under "Link Properties" in the table inspector to implement it. Can you please either add details or make a note that that will be covered in a different spec?

todo
```diff
@@ -0,0 +1,9 @@
+- Positioning and Scroll
```
kgodey, line 1: I assume this file was not meant to be committed, please remove it.

kgodey (changes requested):

seancolsen (commented):
Looks good overall! A couple minor points...

- I don't understand the "Impact of the Extract Columns Operation" section. It says:

    > The dialog will make the following information clear

    What, specifically, should the dialog do in order to make that information clear?

- For adding additional columns once the modal is open, I think the plus button should be relocated to its own row, below all rows for columns. As it stands in the specs, it looks like the plus button is associated with one column. It should be associated with the entire list of columns.

design/specs/column-extraction.md
seancolsen: In this [email thread](https://groups.google.com/a/mathesar.org/g/mathesar-developers/c/yu1dOjV7EC8) we decided to use the table name only. @ghislaineguerin can you update the specs to reflect that?

design/specs/column-extraction.md
ghislaineguerin: @kgodey My concern with moving vs. extraction is that moving suggests just changing the columns location, but we also create new columns and change the total rows of the moved data (deduplicate).

design/specs/column-extraction.md
ghislaineguerin: I think in this case 'moving columns' would apply, and the available tables would have to be linked already.

design/specs/column-extraction.md
ghislaineguerin: updated

design/specs/column-extraction.md
ghislaineguerin: I have updated the naming convention section.

ghislaineguerin:
@seancolsen you asked:
> what should the dialog do in order to make that information clear?
I was referring to the informational text that is available under each section.

I agree with your suggestion for button placement and have updated accordingly.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,66 @@ editor: markdown
...
+- Help users understand how the tables are linked after the split operation
 
-1. **The user imports a table from a file**
+## Splitting a Table by Extracting Columns
 
-    This user flow assumes that the starting point is an imported table containing columns that can be extracted and turned into entities. A column that may be transformed into a table or entity will usually include other columns that are attributes for that entity. If the database is not normalized, the entity to be extracted might include duplicated values. For example, consider a table of publications with multiple entries for the same publisher in the publisher column.
+### Selecting columns as a starting point
 
-2. **The user selects the columns (attributes) for the new entity**
+The user will select one or multiple columns to start with the column extraction process. All columns can be extracted, except for primary key columns. Once selected, the inspector panel will list 'Extract Columns to Table' as an action. Selecting this option will open the 'Extract Columns to Table' dialog.
```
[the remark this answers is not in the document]
silentninja: How do we make it clear that only certain columns can be extracted? The user might think it is a bug when certain columns don't show for extraction

design/specs/column-extraction.md
```diff
@@ -8,87 +8,66 @@ editor: markdown
...
+- A new table will be created with the extracted columns
+- A link will be created between the new table and the original table
 
-    <img width="1469" alt="image" src="https://user-images.githubusercontent.com/845767/177546415-23b11ded-8987-4e69-ab1a-1decb9e773d3.png">
+### Link Column Naming Convention
 
+It was decided that the link column naming convention should be as follows:
 
-5. **The user navigates to the new table**
+>In the application, where singularization is not straightforward, we don’t suggest column names. Instead we require the user to manually name the FK column.
```
silentninja, line 49: The below section `Using Rules to Generate Column Names` talks about using rules for generating column names to be a future feature. But here it is mentioned 
> where singularization is not straightforward, we don’t suggest column names

So Should we suggest the link column name if singularization is possible now?

design/specs/column-extraction.md
```diff
@@ -8,87 +8,66 @@ editor: markdown
...
 
-    The new table will be listed in the list of tables for the current schema. The user can modify the table, add new columns or rename it if needed.
+In this case, since we cannot infer the new table name from the selected columns, then the user must manually enter both, new table name and link column name.
 
-    <img width="874" alt="image" src="https://user-images.githubusercontent.com/845767/177550009-30fcae81-2ae9-4539-ad85-5a0d1289790b.png">
+#### Using Rules to Generate Column Names
```
silentninja, line 53: We would be generating a single foreign key column, so shouldn't it be `Column Name` instead of `Column Names`?

silentninja (changes requested):

design/specs/column-extraction.md
kgodey: I get your point but there's still something about the current phrasing that bugs me. Part of it is that you can either create a new table or move them to an existing table, so you're not always extracting the columns to a table. 

Maybe we can instead have two separate actions?
- Create new linked table using columns
- Move column to existing linked table

design/specs/column-extraction.md
kgodey: Yes. The spec needs to cover this case, though, since it's required for the storyboard. Please see my note above about two separate actions.

design/specs/column-extraction.md
kgodey: @ghislaineguerin This is not what we decided in the email thread, this was @seancolsen's original proposal. We decided that the link column's name would be based on the table name. So if the user enters `Author` as a table name, then the link column would also be called `Author` (unless the name is taken already, then we can leave it blank).

Also it would be helpful to link to the email thread (on the mailing list) from the spec so people can follow the reference back to the source.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,66 @@ editor: markdown
...
+
+Additionally, the user can rename the columns that will be extracted in cases where the column name is not descriptive enough.
+
+## Finalizing the Extract Column Operation
+
+Once the user is happy with the changes, they can click on the `Extract Columns` button. The inspector panel will be toggled and the new link column will be selected in the original table. Under the `Columns` section, the user will see the new table listed as a link.
+
+![image](/assets/design/specs/column-extraction/185143860-f2995fc4-c767-4cce-aaa6-319ef27630ff.png)
+
+Additional details regarding 'Link Properties' will be added in a future spec.
```
kgodey, line 73: This doesn't make it clear that `Link Properties` is not meant to be implemented in this spec. Please make that clear.

kgodey (changes requested):
Please see individual comments. Also before this is merged, @ghislaineguerin please make sure that the main specs page links to this page and any unused images are removed.

design/specs/column-extraction.md
ghislaineguerin: I've updated the actions

ghislaineguerin:
@kgodey @silentninja 
Since we'll no longer do renaming, I feel that the column selection with dropdown list won't make sense. Could we do something like this instead:

https://user-images.githubusercontent.com/845767/186213149-d3ed84f5-36d8-43d4-8b68-cbbe8c0fcb9c.png

- It would allow for multi-selection
- It can include changing order of the selection in the future

seancolsen:
@ghislaineguerin A few points regarding the UI for the column selection...

- I like the design in your mockup. I'll call it a side-by-side-multi-select.

- @kgodey [said](https://github.com/centerofci/mathesar-wiki/pull/54#discussion_r948496519):

    > I suggest implementing this renaming feature in the future

- Because we plan to eventually change this UI, I'd be inclined to do the least amount of work necessary to select columns in the short term.

- We don't yet have a side-by-side-multi-select component, and building one will require some work. I'd rather not do that work now if we don't really need it.

- Here are a couple options which would require less work than building the side-by-side-multi-select.

    - (A) We could implement the column selection UI originally in this PR, just without the renaming. That would position us well to implement the renaming feature in the future.

    - (B) Back in April, I did some work to build a UI for selecting columns to be used in our old record selector design. I started creating this multi-select component for columns:

        Peek 2022-08-23 13-27

        We could potentially finish going this direction without too much work.

kgodey:
I think @seancolsen's reasoning is sound, I like option (B) in his proposal. I'm not sure if/when we'll prioritize renaming or reordering columns, we can rethink this UX when we do that.

ghislaineguerin:
@seancolsen @kgodey I like Sean's proposal as well. I will update.

design/specs/column-extraction.md
ghislaineguerin: We won't be renaming for now. We can create a ticket in the future when needed.

design/specs/column-extraction.md
ghislaineguerin: @kgodey I think instead of blank, a number suffix can be added to the link column name if the name is already taken. The number will be incremented by 1 for each new link column.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,62 @@ editor: markdown
...
 ---
 
-# Context
-Tables must be split into smaller tables during the normalization process to ensure data integrity. To make this process easier, the product should include capabilities that allow users to restructure data by splitting, merging, and combining existing tables.
-Initially, just table splitting will be implemented, so that a user can select columns from an existing table and create a new one.
+## Context
 
-# Splitting a Table by Extracting Columns
-To create a new table from an existing one, select the columns to include in the new table. These columns must not be primary keys for the parent table (although they can be foreign keys to another table). Once selected, the inspector panel should display a list of possible 'Selection Actions,' with the 'New Table from Columns' option allowing the user to extract the columns into a new table.
+In data modeling, it's usual practice to split data up into related tables. In the context of Mathesar, one way of doing this is by extracting columns from a table into a new table. This process is known as column extraction.
```
kgodey, line 13: Nitpick, but it might be useful to explain why data is split up into related tables (one table per type of object, data deduplication, etc.)

design/specs/column-extraction.md
```diff
@@ -8,87 +8,62 @@ editor: markdown
...
+- Help users understand how the tables are linked after the split operation
 
-1. **The user imports a table from a file**
+## Splitting a Table by Extracting Columns
 
-    This user flow assumes that the starting point is an imported table containing columns that can be extracted and turned into entities. A column that may be transformed into a table or entity will usually include other columns that are attributes for that entity. If the database is not normalized, the entity to be extracted might include duplicated values. For example, consider a table of publications with multiple entries for the same publisher in the publisher column.
+### Selecting columns as a starting point
 
-2. **The user selects the columns (attributes) for the new entity**
+The user will select one or multiple columns to start with the column move process. All columns can be moved, except for primary key columns, unorderable columns like JSON-type columns and referent columns (Columns referenced by other columns).
```
kgodey, line 27: Also nitpick, but `columns` should be lowercase inside the params.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,62 @@ editor: markdown
...
 
-    The user chooses all the columns that will be attributes for the new object. For example, in the instance of the publisher, there could be columns for 'Publisher Name' and 'Publisher Id.' Including all the attributes is very important since columns cannot be moved after the fact.
+Once selected, the inspector panel will list 'New linked table from columns' as an action. Selecting this option will open the 'New linked table from columns' dialog.
 
-     <img width="1512" alt="image" src="https://user-images.githubusercontent.com/845767/177544077-9f074830-ce64-4ed9-80bc-53de53977f9c.png">
+When a user selects a column that cannot be moved, the move action will not be available in the inspector panel.
```
kgodey, line 31: Can we instead disable the action and have a help icon that explains why it's disabled?

kgodey (changes requested):
@ghislaineguerin This spec has no details on the "Move columns to existing linked table" action, which we also need for the demo. Please also see inline comments.

Also, one of the images isn't working so I can't evaluate the spec entirely. Please also make sure you delete all unused images (they're in `assets/.unused`)

design/specs/column-extraction.md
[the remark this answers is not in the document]
ghislaineguerin: @silentninja We will not list them and instead have a help icon in the column selection menu.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,62 @@ editor: markdown
...
+This spec describes the steps a user would take to create a new table from a subset of columns from an existing table.
 
-## Selection Actions
-Depending on a particular selection, the inspector panel will display actions that can be performed on the selection, as described above with columns. These selection actions will be determined by the type and number of items chosen.
+## Design Goals
 
-# User Flow
+- Allow users to split a table into multiple new tables by selecting columns
+- Help users understand the impact of splitting a table
+- Help users understand how the tables are linked after the split operation
```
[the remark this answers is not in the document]
ghislaineguerin: I have removed it. We can add this to the new issue.

design/specs/column-extraction.md
```diff
@@ -8,87 +8,62 @@ editor: markdown
...
+## Extract Columns Dialog
 
-    The user can choose to keep copies of the columns in the table if needed. By default the extracted columns are deleted.
+The dialog will list the columns that will be moved and the table that will be created. The user can change the name of the new table. The dialog will also list the links that will be created between the new table and the original table. The user can change the name of the link column.
 
-4. **The user inspects the new foreign key column (link)**
+![image](/assets/design/specs/column-extraction/7prBiuRUXhPYi6wZxwRcyV.png)
 
-    The new foreign key column will contain links to the new records that have been created. Additionally, when selected, the inspector panel will display the link properties.
+### Impact of the Move Columns Operation
```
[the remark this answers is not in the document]
ghislaineguerin: Yes, I've updated

design/specs/column-extraction.md
ghislaineguerin: Yes, I think that will work. I have updated

ghislaineguerin:
@kgodey I've added the details for "Move columns to linked table". Please review.

kgodey (approved):
Looks good to me

silentninja (changes requested):

design/specs/column-extraction.md
```diff
@@ -8,87 +8,87 @@ editor: markdown
...
+
+This spec does not include the implementation of 'Link Properties' which should be handled in a separate spec.
+
+## Scenario 2: Moving Columns to a Linked Table
+
+In some cases, a user may want to move columns from one table to another table. This can be done by selecting the columns and using the 'Move to existing linked table' action in the inspector panel.
+
+### Selecting a the link column
+
+A table might be linked to through more than one column. The user will be able to select the link column to use for the move operation under the 'Link Column' section. If only one link column exists, it will be selected by default.
```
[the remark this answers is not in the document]
ghislaineguerin: @silentninja I think the case you are presenting makes sense. However the user should have moved both 'First Name' and 'Last Name' to begin with right?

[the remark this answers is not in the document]
kgodey: @silentninja I think your comments make sense but they're out of scope for the current spec. The scope of the spec (and the cycle) is to only implement the "happy path" features that are in the library demo. Please keep your requested changes to this scope.

In the next couple of cycles, we'll be polishing our features and testing them with different data sets, and we can address any problems that come up then.

silentninja (approved):

PR #55, opened 2022-08-19
Inline Edit and Saving

ghislaineguerin:
This PR addressed an unresolved design issue found in #51 . It introduces a new pattern to the application that allows users to edit values inline and save them.


design/specs/inline-save-edit.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Inline Saving and Editing
```
kgodey, line 2: @ghislaineguerin This spec is not linked to from the Specs page in this PR.

PR #56, opened 2022-08-28
Add front end coding standard to prefer function over const

seancolsen:


seancolsen:
Closing in favor of #57

PR #58, opened 2022-08-29
Users & Permissions product spec

kgodey:
This is a product spec for how the Users & Permissions feature should work in Mathesar.

Anyone is welcome to review it, although we will assign individual team members responsible for implementing this in our Cycle 4 planning meeting this week.

[Here's a link to the most readable format for the spec](https://github.com/centerofci/mathesar-wiki/blob/users_product_spec/product/specs/users-permissions.md).


seancolsen (commented):
- For the sake of focus and task prioritization, I suggest relegating "Table permissions" and "Exploration permissions" to "Bonus goals". I think the complexity within those more granular features will reach it tentacles pretty deep within the product, and it would be better to defer that work until our higher priority needs are addressed first.

- > Allow library patrons to log in and see only their own checkouts with the associated due dates.

    This is a really demonstration of the value of allowing users to be represented in Mathesar data!

- > Administrators can reset anyone's password.

    For clarity, I suggest changing this to:

    > Administrators can generate a temporary password-reset URL for any user, copy it, and manually send it to the user (outside of Mathesar), allowing the user to enter a new password into a web form.

- I think it's worth giving some thought up-front to the ways in which an authentication system might slow down local development. Sometimes it can be a pain logging into a local site just to work on it, and it can make E2E testing trickier. I wonder if if would be possible to have an `.env` setting to disable authentication altogether.

kgodey:
> Administrators can generate a temporary password-reset URL for any user, copy it, and manually send it to the user (outside of Mathesar), allowing the user to enter a new password into a web form.

We could also implement it without password reset URLs entirely – the administrator can enter the user's new password into a web form and send this new password to them outside of Mathesar. I think that level of implementation detail is outside the scope of this spec.

seancolsen:
> the administrator can enter the user's new password

This is the situation I'm trying to avoid by suggesting the change. An admin should never see a user's password.

kgodey:
> This is the situation I'm trying to avoid by suggesting the change. An admin should never see a user's password.

I was modeling it on creating new accounts in other tools (e.g. G Suite, Wiki.js), you can set a temporary password when you create a user and then they are required to change it on their first login.

kgodey:
@seancolsen your concern should hopefully be addressed in https://github.com/centerofci/mathesar-wiki/pull/58/commits/88851fc583d856a8a2279ed2fa942fe3dd2b91cd

dmos62:
It's important that we get a few review passes by backend developers. There is some relevant technical debt on the backend that would be perilous to ignore when planning this. @mathemancer  has the best grasp of this, imo. I'd be happy to provide concrete input as well, but I'd need to familiarize myself with Postgres permission system first.

kgodey:
@dmos62 I had a conversation with @mathemancer yesterday about the technical debt and potential implementations for users and permissions. We'll also be discussing it at the weekly meeting tomorrow with the entire backend team.

ghislaineguerin (approved):
I've read through the content and it all makes sense. I don't have any specific feedback at this point.

pavish:
The initial set of goals & related plans look good to me.

I have a few questions:
* The spec needs to mention how the public schema should be handled.
  - Do we intend to make it accessible to all users?
  - Since Mathesar users are different from Postgres users, do we have to portray public schema differently? For eg., in schema page, we decided special behaviour for public schema. Do we need that?
* Public sharing (sharing to everyone via url) needs to be planned out before starting implementation
  - Do we need public sharing for the initial release?
  - If public sharing is required, we might need to add table & explorations sharing to our current goals.
* Thoughts for later:
  * The bonus permissions need more thought. What if the user has Manager access to an exploration, but has no access to the underlying tables used in the exploration?
  * The plan is to have default permissions like `DB Editor`, `DB Viewer`. I assume we'd scope all permissions within the Database level. What would be the plan when we have multiple database support?
  * In the long term, it'll be great if we can create custom roles by choosing individual permissions, and then assigning those roles to users.

I'm sure we'd have more questions during implementation/planning, but I can't think of anything major at this point.

kgodey:
> * The spec needs to mention how the public schema should be handled.
>   * Do we intend to make it accessible to all users?
>   * Since Mathesar users are different from Postgres users, do we have to portray public schema differently? For eg., in schema page, we decided special behaviour for public schema. Do we need that?

Good point. I think we should allow everyone Manager access to the public schema by default and make that clear in the UI when we implement.

Added to the spec in https://github.com/centerofci/mathesar-wiki/pull/58/commits/4f6244d141dc5e8f591fe71ad89e458ba16b5016

>  * Public sharing (sharing to everyone via url) needs to be planned out before starting implementation       
>    * Do we need public sharing for the initial release?
>    * If public sharing is required, we might need to add table & explorations sharing to our current goals.

I consider public sharing a separate feature. It's a nice-to-have for the first release. If we do have time to do it, I'll write up a separate spec for it.

> The bonus permissions need more thought. What if the user has Manager access to an exploration, but has no access to the underlying tables used in the exploration?

Users can only have Viewer access to explorations if they don't have access to the underlying tables.  Added to the spec in https://github.com/centerofci/mathesar-wiki/pull/58/commits/4f6244d141dc5e8f591fe71ad89e458ba16b5016

> The plan is to have default permissions like `DB Editor`, `DB Viewer`. I assume we'd scope all permissions within the Database level. What would be the plan when we have multiple database support?

I'm not sure, options are:
- we allow more Administrator accounts to be created o
- we have a new level of permissions called Mathesar Editor / Viewer
- we do nothing and insist that permissions have to be scoped by DB

It depends on how people want to use multiple databases.

>  In the long term, it'll be great if we can create custom roles by choosing individual permissions, and then assigning those roles to users.

Agreed

>  I'm sure we'd have more questions during implementation/planning, but I can't think of anything major at this point.

We'll have a design spec (or multiple specs) before implementation too.

mathemancer (commented):
Overall, I think this makes sense for now.

I do think it would be nice to have a setup where the database manager didn't actually have access to schemas or their subobjects except to delete them. You don't necessarily want the person who sets up the schema to then have access to all data in the schema forever.

pavish (approved):

kgodey:
> I do think it would be nice to have a setup where the database manager didn't actually have access to schemas or their subobjects except to delete them. You don't necessarily want the person who sets up the schema to then have access to all data in the schema forever.

I think this would make sense in a "set up custom roles" scenario. I'm going for a simple-to-understand model here, and "there's a hierarchy and you have the same permissions all the way down the hierarchy" is easy to grasp.

kgodey:
@mathemancer confirmed on Matrix that he's done reviewing this, so I'm merging it in.

PR #59, opened 2022-09-02
Data Type Options

ghislaineguerin:
Fixes [#1578 ](https://github.com/centerofci/mathesar/issues/1578)


design/specs/data-type-options.md
```diff
@@ -0,0 +1,50 @@
...
+
+![image](/assets/design/specs/data-type-options/6VPzH8Cd7USbhf6jZpzQvc.png)
+
+## Alternatives
+
+### Remove the sub-sections and instead use an icon next to database-level options
+
+A warning message can be displayed inline when the user tries to change a database-level option. The icon can then be used for other instances where the user needs to be aware of database-level options.
+
+Removing sub-sections will make the UI cleaner.
```
pavish, line 48: 1. Could you elaborate on this? I don't understand what is meant by removing sub-sections for database-level options.
2. The image does not show any warning or icon. What would the warning be and how would it look like?

design/specs/data-type-options.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Use a dropdown menu for the data type selection instead of the scrollable list
+
+![image](/assets/design/specs/data-type-options/gL37Da5Tdp4iTZScWJPGC.png)
+
+### Align some of the label and field groups inline
+
+Inline alignment can be used for groups of fields in order to save vertical space.
+
+![image](/assets/design/specs/data-type-options/6VPzH8Cd7USbhf6jZpzQvc.png)
```
pavish, line 40: I can see that the gain obtained in making the input labels inline and using a select menu instead of scrollable list would make up for the vertical space lost in getting rid of tabs.

I think we can proceed with these set of changes for now.

design/specs/data-type-options.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Alternatives
+
+### Remove the sub-sections and instead use an icon next to database-level options
+
+A warning message can be displayed inline when the user tries to change a database-level option. The icon can then be used for other instances where the user needs to be aware of database-level options.
+
+Removing sub-sections will make the UI cleaner.
+
+![image](/assets/design/specs/data-type-options/fpnLR7JtkJJZXGybZoYL8d.png)
```
pavish, line 50: For both the proposed solution and the alternative:

We always need an explicit Save button. This save button _needs to be common_ for all the database options combined and _**not** per input_ because the determination of column type depends on the fully filled database options form. The behaviour of the save button and it's placement needs to be mentioned clearly on the spec.

pavish (changes requested):

design/specs/data-type-options.md
ghislaineguerin: @pavish we have a separate spec for inline saving. I will include a reference here and have a wireframe for reference.

pavish: I understand we have a separate spec for inline saving, my question is whether it will cover the entire form?

> we have a separate spec for inline saving

The current inline saving seems to be per input, each input having it's own save button. But for the data type menu, we cannot do that, because the type is calculated based on the result of all the related inputs. So, we need a common save button.

design/specs/data-type-options.md
ghislaineguerin: 1. I meant we don't need to hace sub-headings for "Database" and "Format" options.
2. I will include an example.

design/specs/data-type-options.md
ghislaineguerin: @pavish yes, the entire section should be saved at once. I will include the example in the spec.

ghislaineguerin: @pavish I'm assuming that save is only required for the database options and that we won't be asking the user to save format options. Is that the right assumption?

pavish: Yes, we can make format options save automatically. However, **I'd prefer the save button to save both database and format options together**. We show the same options in both data import menu and the table inspector, and it would be a lot simpler implementation-wise and maybe more user friendly, if they were as similar to each other as possible.

ghislaineguerin: It makes sense. I was concerned about the inconsistency between this and inline saving but I think you make a good point. I will update.

ghislaineguerin:
@pavish I have updated the wireframe to show the save and cancel options for the entire section.

kgodey (commented):
This looks fine to me, but default value should not be folded into the Data Type "save" action, it should be a separate action.

Also @ghislaineguerin please remove unused images from the PR.

I think we generally need to do some explanatory copy for the table inspector, but we can handle that separately as part of demo video readiness.

pavish (changes requested):
I find the UX a bit confusing because Database options includes the "Data type" Select menu, and it contains a border. The Formatting options are placed out of it, even though they depend on the selected data type.

I think the "Data Type" menu should be placed out of the Database options border, or we should find a way where it is clear that the formatting options depend on the selected data type.

I also agree with Kriti that the default value option should not be folded with the "save" action.

I'd prefer it if 'Default value' is not part of the Data type section, but placed within a constraint section, but this is something we can come back to later.

design/specs/data-type-options.md
pavish: I have further comments placed here: https://github.com/centerofci/mathesar-wiki/pull/59#pullrequestreview-1098147017

ghislaineguerin:
@pavish The default options are a separate section. I agree with your concern about the border, I wasn't sure if grouping the data type selection was a good idea, but I did since it was technically a database option as well. But it might be better as a separate option that determines all the others. I will update.

pavish (commented):
@ghislaineguerin @kgodey

The spec looks good to me.

I do have one concern, which didn't occur to me earlier.

* Whenever we change the type of a column, the default value has to change/get cleared.
  * Since the default value is outside the database form, it may not be immediately known to the user that it has been cleared.
  * I personally find it okay, and I think we can show a confirmation alert modal before changing the type of a column (same as the one we would have for deleting), since column type change would also have data loss implications. We can show those warnings in that modal. 
* Or, we can retain the default value and let the user modify it manually. The column type change would mostly throw an error if the user tries to update the column without removing the default value first.

Which option should we go with? It's not immediately blocking, I've already started implementation on this.

kgodey:
>  I think we can show a confirmation alert modal before changing the type of a column

I think this is fine, can we change the wording of the modal if there's a default value set? We can indicate that the default value will be cleared.

pavish:
> can we change the wording of the modal if there's a default value set

Yes, we can do that.

pavish (approved):
@ghislaineguerin I've added a commit https://github.com/centerofci/mathesar-wiki/pull/59/commits/ffea0d1fb0fc8b1b5db798b8dec6e0bd4f54f2d8 addressing https://github.com/centerofci/mathesar-wiki/pull/59#pullrequestreview-1104535767, https://github.com/centerofci/mathesar-wiki/pull/59#issuecomment-1244085603.

I'm approving the PR from my end. I've leave it to you to take a look before merging.

PR #61, opened 2022-09-16
Propose route changes to keep tables and explorations on the same level

pavish:
This PR proposes some navigation route changes:
* Keeps tables and explorations on the same hierarchical level
* Uses `explorations` as the url path for opening existing explorations
* Since the new designs include tabs for table list and exploration list in the schemas page, the urls `/{schema_id}/tables/` and `/{schema_id}/explorations/` would select the respective tabs
* Adds different routes for import upload and confirmation pages. This is already implemented in current master.

### DIFF
--- a/design/specs/navigation.md
+++ b/design/specs/navigation.md
@@ -103,14 +103,30 @@ Throughout Mathesar, the following links exist within pages which allow for more
   - `<title>`: `{database_name} | Mathesar`
 
 - **Schema Page**:
-  - Route: `/{database_name}/{schema_id}`
+  - Route: `/{database_name}/{schema_id}/`
   - Favicon: Schema
   - `<title>`: `{schema_name} | Mathesar`
+  - By default, redirects to "Schema Page - Tables tab".
+
+- **Schema Page - Tables tab**:
+  - Route: `/{database_name}/{schema_id}/tables/`
+  - Favicon: Schema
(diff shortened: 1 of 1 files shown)


design/specs/navigation.md
```diff
@@ -103,32 +103,48 @@ Throughout Mathesar, the following links exist within pages which allow for more
...
 
 - **Table Page**:
-  - Route:  `/{database_name}/{schema_id}/{table_id}/`
+  - Route:  `/{database_name}/{schema_id}/tables/{table_id}/`
   - Favicon: Table
   - `<title>`: `{table_name} | {schema_name} | Mathesar`
 
 - **Record Page**:
-  - Route: `/{database_name}/{schema_id}/{table_id}/{record_id}`
+  - Route: `/{database_name}/{schema_id}/tables/{table_id}/{record_id}`
```
kgodey, line 147: This is not related to your changes, but I think it would be clearer for this URL to be 

`/{database_name}/{schema_id}/tables/{table_id}/record/{record_id}`

kgodey (approved):
Approved, suggested an additional change. @seancolsen @ghislaineguerin please review my comment too.

design/specs/navigation.md
seancolsen: I'm ambivalent on this.

seancolsen (approved):

PR #65, opened 2022-10-05
Propose front end standard that prohibits setting z-index values on components

seancolsen:
@pavish @rajatvijay We have a number of small bugs related to z-index that I'd like to fix en masse. I'd like to adopt a new pattern (as documented in this PR), and employ this pattern to fix all our z-index bugs. How does this sound?


engineering/standards/frontend.md
```diff
@@ -151,11 +151,15 @@ Note: some of our older code still does not conform to this standard.
...
+    - It's fine for child elements _within the component_ to set a z-index, but in such cases the component's root element must establish its own [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). It's best to use `isolation: isolate;` to establish the stacking context because it clearly communicates intent and works without setting z-index. Once your component has its own stacking context, then you're free to set the z-index values within the component without thinking about anything outside the component. You can use simple values like `1` and `2` within the component because everything is encapsulated.
+    - If you want to render a child component with a specific z-index, then prefer to nest the child component inside a DOM element, setting the z-index on the DOM element (not the component). Alteratively if you absolutely must pass a z-index value _into_ a component, then do so by setting something like `z-index: var(--component-name-z-index, auto);` within the component, and then setting something like `--component-name-z-index: 2;` in the parent.
```
pavish, line 161: I assume you're not suggesting using style props here. I'd like to completely avoid style props in our code base since it adds a dom element automatically.

engineering/standards/frontend.md
```diff
@@ -151,11 +151,15 @@ Note: some of our older code still does not conform to this standard.
...
 
 - **margin**: The component's root element should not set any margin.
-
 - **padding**: If the component's root element has border, it's fine to set padding because the border will serve as the outer-most visual edge. But if there's no border, then there should be no padding.
+- **z-index**:
+    - The component's root element should not set any z-index.
+    - It's fine for child elements _within the component_ to set a z-index, but in such cases the component's root element must establish its own [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). It's best to use `isolation: isolate;` to establish the stacking context because it clearly communicates intent and works without setting z-index. Once your component has its own stacking context, then you're free to set the z-index values within the component without thinking about anything outside the component. You can use simple values like `1` and `2` within the component because everything is encapsulated.
```
pavish, line 160: We may want to establish some css variables with default z-index values for use with components within the component library. I think it's reasonable for components like modal and dropdown content to have z-index values.

engineering/standards/frontend.md
seancolsen: This PR does not suggest or prohibit style props. I'm suggesting that components can only accept z-index values via CSS variables. Once a component accepts a CSS value through a CSS variable we can pass a value to it either (a) through style props, or (b) through CSS defined within an enclosing scope. If you want to completely avoid style props in our code base, I think it would be good to open a separate PR proposing a standard for it. We can have a separate discussion there. If we decide to ban style props, we could still make use of the CSS variable approach here. Notice that in this PR I'm suggesting the CSS variable be named `--component-name-z-index` and not `--z-index`. Namespacing the CSS variable within the component name makes it easier for consuming components to define that CSS variable within their CSS instead of through style props. I do think it's worth having a separate conversation about style props as a pattern (because I don't think we're entirely in agreement), but I think the outcome of that conversation needs to affect the standard I'm proposing in this PR.

engineering/standards/frontend.md
seancolsen: Let me see if I follow you...

1. The Modal system would set:

    ```css
    .modal {
      z-index: var(--modal-z-index, auto);
    }
    ```

1. The Toast system would set:

    ```css
    ..toast-presenter {
      z-index: var(--toast-presenter-z-index, auto);
    }
    ```

1. In `src/component-library/common/styles/variables.scss`, we'd set something like:

    ```css
    :root {
      --modal-z-index: 100;
      --toast-presenter-z-index: 101;
    }
    ```

    This way, toast messages appear above the modal overlay.

Is that what you had in mind? If so, I think that's good. I would just want to confirm that any CSS we define within `App.css` will have higher precedence than that defined within that `variables.scss` file -- because component library consumers will need a mechanism to override those z-index values.

We'll need to play with Dropdown a bit to make sure everything works -- i.e. a dropdowns opened outside of modals should display under modals and and dropdowns opened within modals should display above modals. My intuition tells me that modals and dropdowns will need to have the same z-index value. I'm not positive that'll cover all the use-cases though.

pavish: > Is that what you had in mind

Yes, this is what I had in mind.

engineering/standards/frontend.md
pavish: Sounds good

pavish (approved):

PR #68, opened 2022-10-12
Rewrite record summary specs

seancolsen:
@kgodey and @ghislaineguerin this PR delivers my simplified design for the UX to configure the record summary template. The design as spec'ed in this PR will be much easier to implement than the design as previously spec'ed. I also simplified some of the surrounding content within the spec and went into greater detail in some areas too.

Rendered: https://github.com/centerofci/mathesar-wiki/blob/record_summary_template_config_specs/design/specs/record-summary.md


kgodey:
I recommend further simplifying the configuration by:
- removing the default / custom toggle and always showing the record summary field with whatever is currently configured
- removing the option to toggle through different previews – show a preview of one record

I also recommend renaming "Insert Field" to "Insert Column", since we do not use the "Field" nomenclature anywhere else. I'm not sure if it's best to place this above or below the summary field, but I assume that can be easily adjusted later.

seancolsen:
Thanks for the quick review, @kgodey. Good suggestions.

- I've added 138798b which removes the toggle buttons and uses "column" instead of "field".

- I think it's still important to keep the "default" vs "custom" radio options for the following reason:

    1. Say we have the following record

        ```json
        {
            "101": 1,
            "102": "Foo"
        }
        ```

    1. The default template would be `{102}`.

    1. But there is a difference between _storing_ a template of `{102}` and storing no template at all. When no template is stored, then it gets generated on-the-fly. That way, if the first non-pk column is deleted, then the next non-pk column will get used.

kgodey:
Makes sense. I will discuss with @ghislaineguerin at our call today and approve it afterwards.

When do you think we should implement this – before or after @rajatvijay's table inspector design work?

seancolsen:
> When do you think we should implement this

Good question. I'd like to wait until after the other table inspector changes are implemented. That'll be most efficient I think.

ghislaineguerin:
@kgodey and I reviewed this. It looks good. I will updated Table Inspector design.

seancolsen:
@ghislaineguerin one thing to note... I will be implementing the UI for this (not Rajat). I don't need you to make any updates to figma. I have everything I need to move forward with the implementation based on your approval here.

kgodey:
@seancolsen I asked Ghislaine to make a quick update in Figma so that I could see the record summary in the context of the rest of the Table Inspector and see if any additional changes to the UX would be needed.

Now that I see it in [Figma](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=769%3A11268), we need some sort of way to explain what the record summary is and where it will be used. We could do a help tooltip (the red circle with a question mark inside it), although I'm not sure about the placement of it. I also think the Record Summary should go below the Links section, but above Actions.

Do you want to add the help text in this spec or should we just figure that out in Figma?

kgodey:
On second thought, I think we should merge this and figure out the styling and help text work in Figma since it should be consistent with the rest of the inspector. I'm merging this.

seancolsen:
Cool cool. I'll follow what's in figma when I implement this.

PR #75, opened 2023-01-19
Users and permissions RFC

pavish:
Contains RFC document for Users and Permissions.

[Rendered](https://github.com/centerofci/mathesar-wiki/blob/fe-rfc-users-permissions/engineering/specs/usersandpermissions.md)

### DIFF
--- a/engineering/specs.md
+++ b/engineering/specs.md
@@ -8,6 +8,9 @@ editor: markdown
 dateCreated: 2022-03-18T22:25:33.876Z
 ---
 
+## 2023-01
+- [Users and Permissions frontend RFC](/en/engineering/specs/usersandpermissions)
+
 ## 2022-10
 - [Dependents graph](/en/engineering/specs/dependents-graph)
 
--- a/engineering/specs/usersandpermissions.md
+++ b/engineering/specs/usersandpermissions.md
@@ -1,0 +1,182 @@
+---
+title: Users and Permissions frontend RFC
+description:
+published: true
+date: 2023-01-19
+tags: 
+editor: markdown
+dateCreated: 2023-01-19
+---
+
+**Feature**: Users & Permissions frontend
+**Meta ticket**: [Users & Permissions Frontend - Meta ticket](https://github.com/centerofci/mathesar/issues/2205)
+**Design**: [Figma design link](https://www.figma.com/file/xHb5oIqye3fnXtb2heRH34/Styling?node-id=3666%3A25085&t=b2Pw2pUJh42nRKQ3-1)
+**Related**: N/A
+
+This document is a RFC for high level technical specification and a task list for the initial version of the Users & Permissions feature. Note that this only focuses only on the frontend.
+
+# Task list
+* API utility for users
+* API utility for permissions
+* Stores needed for user profile
+* Utilities for user permissions
+* Account profile route & page
+* Users route & page
+* Password reset page UI & rerouting logic
+* Permission based restriction in DB route
+* Permission based restriction in Schema route
+* Permission based restriction in Table route
+* Permission based restriction in Exploration routes
+* Permission based restriction in Record route
+* Modal for editing access level in DB page
+* Modal for editing access level in schema page
+
+# High level implementation details
+## Dedicated API utils
+### Issue
+* We have been utilizing stores directly and encapsulated all API calls within them.
+* This is problematic for a number of reasons including:
+  - Not being able to make API calls that do not affect the store.
+  - Certain objects do not require a store (in this case - Users).
+* We have been side-stepping these issues by:
+  - Having utility functions within store files to make API calls & not update the store.
+  - Making API calls directly within components
+
+### Proposal
+* A long-term solution is to separate out API calls from the stores, and have dedicated API utils for each of our endpoints.
+* We have a directory `/api/` that contains `/api/types` and `/api/utils`. The new files will be placed within `/api/`.
+* We will no longer place types within `/api/types`. Each utility file will export it's own related types.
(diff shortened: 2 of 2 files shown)


seancolsen:
@pavish Here are some thoughts I have. I feel like a sync call would be a good next step here.

## Dedicated API utils

- I really like all the plans in this section! I'm excited about eventually moving the rest of our codebase to follow this pattern

## Routes

- Instead of `/profile/`, I'd rather use `/profile/{user_id}`. We can have Django redirect `/profile/` to `/profile/{user_id}` if we like.

## Permission access level checks in App routes

- You have:

    ```ts
    const currentUser = getUserProfileContextStore();

    $: schemaAC = getSchemaAccessController({ databaseId });
    $: hasManagerAccess = currentUser.hasPermission(schemaAC.manage);

    $: tableAC = getTableAccessController({ databaseId, schemaId });
    $: hasTableEditAccess = currentUser.hasPermission(tableAC.edit);
    ```

- I want to critique the code above, but first let me adjust the example code without changing the overall design.

    Let's say I'm somewhere inside a schema and I want to produce a `canEditData` variable which is `true` when the current user can do DML operations.

    With your design we'd have:

    ```ts
    import { getUserProfileContextStore } from 'path01';
    import { getSchemaAccessController, getTableAccessController } from 'path02';

    const currentUser = getUserProfileContextStore();

    export let database: Database;
    export let schema: SchemaEntry;

    $: schemaAC = getSchemaAccessController({ databaseId: database.id });
    $: hasManagerAccess = currentUser.hasPermission(schemaAC.manage);
    $: hasEditorAccess = currentUser.hasPermission(schemaAC.edit);
    $: canEditData = hasManagerAccess || hasEditorAccess;
    ```

    That's a lot of code to do something pretty common. (Note that I included the imports to demonstrate the how weighty this approach feels.) Can we make this easier on ourselves?

- What if instead of those _nine_ lines of code we had only these _two_?

    ```ts
    export let schema: Schema

    $: canEditData = schema.acl.allowsCurrentUserTo('editData');
    ```

- Here's what I have in-mind for supporting code that would allow us achieve the more concise version above...

    - Somewhere in `/src/api/...` we'd have types like these to define response values we get from the API. (This is not my design, but I'm including it just for clarity's sake.)

        ```ts
        type Role = 'manager' | 'viewer' | 'editor';

        interface DatabaseRole {
          id: number,
          database: number,
          role: Role,
        }

        interface SchemaRole {
          id: number,
          schema: number,
          role: Role,
        }

        interface User {
          // ...
          database_roles: DatabaseRole[];
          schema_roles: SchemaRole[];
        }
        ```

    - Then, in several other front-end files we'd have...

        ```ts
        type RoleSet = Pick<User, 'database_roles' | 'schema_roles'>;

        type AccessOperation = 'editData' | 'editStructure' | 'editAccess';

        abstract class Acl {
          abstract allows(
            roleSet: RoleSet | undefined,
            operation: AccessOperation
          ): boolean;

          allowsCurrentUserTo(operation: AccessOperation): boolean {
            return this.allows(getUserFromContext());
          }
        }

        class SchemaAcl extends Acl {
          constructor({
            databaseId,
            schemaId,
          }: {
            databaseId: number,
            schemaId: number,
          }) {
            this.databaseId = databaseId;
            this.schemaId = schemaId;
          }

          allows(roleSet: RoleSet | undefined, operation: AccessOperation): boolean {
            // ...
          }
        }

        /**
         * This would supplant the current `SchemaEntry` type that we're using across
         * the front end. We'd still keep `SchemaEntry` as an API type, but we'd
         * construct a `Schema` from a `Database` and a `SchemaEntry`.
         */ 
        interface Schema {
          // These fields also exist 
          id: number;
          name: string;
          description: string | null;
          hasDependencies: boolean;
          numTables: number;
          numQueries: number;

          // These fields are new compared to the existing `SchemaEntry` type
          database: Database;
          acl: SchemaAcl;
        }
        ```

        Similar to `SchemaAcl`, we'd also have a `DatabaseAcl` class, and eventually a `TableAcl` class.

pavish:
@seancolsen Sounds good, we can have a sync call over permissions and finalize it.

> Instead of /profile/, I'd rather use /profile/{user_id}. We can have Django redirect /profile/ to /profile/{user_id} if we like.

I'm not seeing the value of having user_id here. Does it serve any purpose?

pavish:
@seancolsen 

This is a clarification with the design proposed, as dicussed on the call:

You mentioned,
> With your design we'd have:
> 
> ...
> 
> $: schemaAC = getSchemaAccessController({ databaseId: database.id });
> $: hasManagerAccess = currentUser.hasPermission(schemaAC.manage);
> $: hasEditorAccess = currentUser.hasPermission(schemaAC.edit);
> $: canEditData = hasManagerAccess || hasEditorAccess;

This is not how it works. The permission `schemaAC.edit` checks if the user has manager or editor roles within the schema access controller. So, we'd only have:

> $: schemaAC = getSchemaAccessController({ databaseId: database.id });
> $: canEditData = currentUser.hasPermission(schemaAC.edit);

`Permission` here is the same as `AccessOperation` mentioned in your proposal. So, the design is essentially similar.

What you seem to be suggesting is creating an abstraction over the low level utils to reduce the number of lines, and I think that's a good idea. I assumed it would be an implementation detail not required for the RFC, but I'll document it as well.

With this, I believe we have no more changes, I'll update the RFC and merge it in.

seancolsen:
I'm closing this because I don't think we need it anymore. It served its purpose. I don't think merging it would do any good since the code in this PR is already quite out-of-date compared to what we ended up going with. Happy to re-open if anyone actually wants this committed for some reason.

PR #77, opened 2023-01-31
Add code standard for $$Props and $$restProps

seancolsen:
## What

This PR proposes a new front end code standard mandating that whenever we use `{...$$restProps}` we also define a `$$Props`  type.

## Why

- Without `$$Props`, I see errors in my editor. Example:

    image

    I don't understand why I don't see a type error in my editor but I _don't_ see a type error when running `svelte-check`. I suspect it has something to do with different versions of TS or Svelte Language Tools on my host machine vs our Docker container.

### DIFF
--- a/engineering/standards/frontend.md
+++ b/engineering/standards/frontend.md
@@ -361,61 +361,56 @@ Additional context:
 
 
 
-### Usage of `{...$$restProps}`
+### When using `{...$$restProps}`, define a `$$Props` type
 
-We use `{...$$restProps}` occasionally in our code despite some drawbacks.
+- Example:
 
-Example:
+    - `Child.svelte`
 
-- `Child.svelte`
+        ```svelte
+        <script lang="ts>
+          export let word: string;
+        </script>
 
-    ```svelte
-    <script>
-      export let word;
-    </script>
+        <span class="child">{word}</span>
+        ```
 
-    <span class="child">{word}</span>
+        `Child` explicitly accepts a `word` prop.
+
+    - `Parent.svelte`
+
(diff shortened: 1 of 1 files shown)


pavish (approved):
Surprisingly, I don't see this issue on my editor, I use the same versions within docker and my local setup.

I agree with the reasoning however, and I think it's an improvement.

rajatvijay (approved):
Thanks, sm for this. I definitely have struggled with this before.

pavish:
@seancolsen @rajatvijay 

I tried to look into this further and I think I understand the cause. Seems to be a bug or maybe an intended restriction with the language server.

`WarningBox` does not have any props of it's own. `$$restProps` is meant to pass down all props that isn't defined as a prop of the component itself. In order to passdown all props, we should be using `$$props`.

If WarningBox is updated to `<MessageBox {...$$props} ../>` instead of `<MessageBox {...$$restProps} ../>`, notice that the errors disappear. The errors also disappear when WarningBox has a prop of it's own.

seancolsen:
@pavish

I put a screenshot in this PR description showing a TS error. Let's define that situation as a "false positive" TS error. (I see an error when I should not see an error).

If I switch to `{...$$props}` then the false positive is eliminated. Great!

But now I have this false _negative_:

image

I don't see an error when I _should_ see an error.

pavish:
@seancolsen Makes sense. This is a case made for stronger typing in our codebase, and I agree with it.

I was trying to describe the issue behind this anomaly in my comment and why `svelte-check` didn't throw any errors.

PR #80, opened 2023-02-07
Added `status: ` label practice in contributing guide for external contributors

rajatvijay:


kgodey (changes requested):
@rajatvijay Unfortunately, this won't work. Community members don't have the permissions necessary to request reviewers or change labels. They need to be added to the repo to have the permissions.

rajatvijay:
@kgodey thanks for pointing it out, didn't know that. Closing it in that case.

PR #87, opened 2023-03-21
Update default milestone in issue-triage doc

rajatvijay:


kgodey:
We also have the `Icebox` milestone for low-priority items, but I think this is fine for now.

PR #90, opened 2023-04-03
Propose Clipboard Interaction specs

seancolsen:
[Rendered](https://github.com/centerofci/mathesar-wiki/blob/clipboard-specs/design/specs/clipboard-interactions.md)

### DIFF
--- a/design/specs/clipboard-interactions.md
+++ b/design/specs/clipboard-interactions.md
@@ -1,0 +1,315 @@
+---
+title: Clipboard Interaction specs
+description: 
+published: true
+date: 2022-03-27T00:00:00.000Z
+tags: 
+editor: markdown
+dateCreated: 2022-03-27T00:00:00.000Z
+---
+
+## Design goals
+
+- Allow users to copy/paste one (and sometimes multiple) cell values within Mathesar -- and also between Mathesar and another applications.
+
+## Additional Context
+
+- [GitHub Issue](https://github.com/centerofci/mathesar/issues/2377)
+- [Usability Improvements project](https://wiki.mathesar.org/en/projects/2023-04-usability-improvements.md) which contains this work
+
+## Terminology and abbreviations in this document
+
+- ⏩ represents a tab character within strings
+- MIME refers to a [MIME type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)
+- TSV refers to "Tab Separated Value"
+- Similar products
+    - **Calc** means LibreOffice Calc
+    - **Sheets** means Google Sheets
+    - **Airtable** means Airtable
+    - **GT** means Google Tables
+
+## Research
+
+### How clipboards work
+
+- When you copy data into your system clipboard, that data is stored along with an associated MIME type. In the simplest scenario of copying plain text, the text you copy is stored along with the MIME type `text/plain`.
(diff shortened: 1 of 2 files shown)


seancolsen:
@pavish and @ghislaineguerin I have assigned you to review these proposed UX specs for the Clipboard Interaction design as motivated by https://github.com/centerofci/mathesar/issues/2377.

I've assigned you two since you're designated helpers on the [Usability Improvements project](https://wiki.mathesar.org/en/projects/2023-04-usability-improvements.md) which contains this work.

I also have a [draft PR](https://github.com/centerofci/mathesar/pull/2773) which begins to implement the "copy" functionality.

@pavish and @Anish9901 I'll highlight that this design work seems to have some overlap with the [UI for Importing data into existing tables](https://wiki.mathesar.org/en/community/mentoring/project-ideas/ui-import-data-existing-table) GSoC project, which you are mentoring.

design/specs/clipboard-interactions.md
```diff
@@ -0,0 +1,315 @@
...
+
+- In my [initial "copy" implementation](https://github.com/centerofci/mathesar/pull/2773), I notice that copying a lot of cells takes some time and **locks the main thread.** This is not great, but it's because we're using the sync API. It's hard to get around this though since the async API isn't fully implemented in Firefox yet. We could potentially use the async API if it's available, while falling back to the sync API for Firefox. If we want to do that, I'd rather do it later when/if users ask for improved performance here.
+
+- It would be nice to add **context menu entries for copy/paste**, but there are some challenges here.
+
+    - When copying: If we use the synchronous clipboard API it seems[^1] that we can't trigger copy events programmatically. Somehow Google Sheets is doing it, but I haven't dug deeper to figure out how. With the async clipboard API Firefox lacks support for MIME types other than `text/plain`.
+
+    - When pasting: The synchronous clipboard API doesn't allow programmatic pasting at all. The async clipboard API doesn't work allow pasting in Firefox and requires a special permission in other browsers.
```
pavish, line 295: Doesn't the clipboard API support this?

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

We could use `navigator.clipboard.writeText()` method to write to the clipboard, the only caveat being Firefox only supports this event from a user action, which will be the case for context menus.

For paste, we'd be reading from the clipboard using the `navigator.clipboard.readText()` method, and performing "paste" manually i.e. updating the cells ourselves.

I remember doing this a long time ago and there were some challenges like the elements need to be in focus. 

I'm adding this comment here, perhaps we could update the document with this as well, so that when we want to implement this, this comment can be a starting point.

pavish: It does. Perhaps for Firefox we can limit support to `text/plain` and try working with that, and/or we could write items to localstorage as a way to improve copy/paste support within Mathesar.

Either way, I agree this isn't a high priority at the moment, and we can consider picking it up later.

ghislaineguerin:
@seancolsen The spec looks good. Regarding your design notes, I think having no contextual menu option is fine for now since content of our cells is very consistent with no styling, so we can do without paste options. I'm assuming that's what you meant by "How to Paste." Is that correct? What other paste options did you had in mind?

One feature I find very helpful is the ability to paste a list of text with line breaks, creating new rows for each entry. I'm not sure if that's a consideration, but it's something we should explore in the future.

seancolsen:
I'm going to merge this now since it's seems like we have sufficient approval. We can continue to tweak smaller details as necessary.

PR #91, opened 2023-04-04
Localization Project Proposal

rajatvijay:
[Rendered](https://github.com/centerofci/mathesar-wiki/blob/localization-project-proposal/projects/localization.md)

### DIFF
--- a/projects/localization.md
+++ b/projects/localization.md
@@ -23,15 +23,15 @@ dateCreated: 2023-03-15T00:00:00.000Z
 
 ## Problem
 
-Currently, Mathesar has hardcoded English strings and only works for the English language. It should be enabled to support any language.
+Currently, Mathesar's codebase has hardcoded English strings and only works for the English language. It should be enabled to support any language.
 
 ## Solution
 
-### Language Detection
+### Language Detection & Language Switcher
 
-_TODO_
+The `selected_language` will be stored inside the DB and sent to the front end inside the `common_data`. For a particular user when there is no `selected_language` stored, the backend will detect the language from `Accept-Language` HTTP header with a fallback to `en`.
 
-**Language Switcher**
+The logged-in user will have a language switcher with a list of all of the currently supported languages by Mathesar. This option will be provided on the user profile page available at `/profile` path. Selecting a new language from here will make an API call to the backend to update the `selected_language` in the DB.
 
 ### Translations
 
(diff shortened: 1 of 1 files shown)


kgodey:
@rajatvijay Please merge this and send an email for review, that workflow is easier for me.

seancolsen:
Thanks for this document, @rajatvijay. I have been spending some time reading it and familiarizing myself with `typesafe-i18n`. I have a lot of questions, critique, ideas, opinions, and thoughts. But before I launch into them, I'd like to piggy-back on @kgodey's comment by suggesting something _slightly_ different (though still compatible with Kriti's request, I think)...

I see this document as a mix of two things that warrant two separate discussions, so **I'd like you to divide this document up** as follows:.

1. Simplify the document in this PR to be higher-level project plan which summarizes _what_ we will do, _who_ will do it, and _when_ we will do it. Merge it yourself, and solicit feedback on it via email so that Kriti can approve it.

1. Take the content on _how_ we will do the work, and extract it into a separate specs document at a wiki path like `engineering/specs/localization.md`. Then make a PR with that document where interested parties can have a more technical discussion about the details. There are some significant architectural question involved in this localization work, and I'd like for @pavish to give it a look as well, before we decide on our approach.

I'll get a little more specific about the sort of delineation I'm requesting, and I'll also offer some critique in a effort to move us towards the sort of "project plan" that I think we need for this project...

- I'd imagine taking the entire "Solution" section and moving that into a specs document.

- In the "Solution" section, I'd want to see something like:

    1. Write architectural specs for our translation system.
    1. Review and approve specs.
    1. Research translation workflow processes compatible with our architecture, and decide out how we will outsource the actual translation work.
    1. Implement the translation architecture, placing _a single UI string_ into the translation system, and providing a Japanese translation for that string.
    1. Implement a linting rule to identify untranslated UI strings.
    1. Wrap all UI strings in translation functions, leaving all but one string untranslated. At this point, the Japanese language is enabled, with English as a fallback.
    1. Test, validate, and hone our workflow for translators to become suitable for less-technical people to perform the translation work.
    1. Publish documentation for developers, providing clear guidelines and standards for authoring components in a manner compatible with our translation system. E.g. [CiviCRM's "Translation for Developers" docs page](https://docs.civicrm.org/dev/en/latest/translation/)
    1. Publish documentation for volunteer translators, making it possible for community members to contribute to translations without very much technical knowledge.
    1. At this point, this project is considered complete, even though almost all our UI strings will still remain untranslated in English. The work of translating the UI strings will be handled later.

- Note that the "Solution" content above is higher-level than what you wrote (because it makes not mention of details like typesafe-i18n). But, it also provides more information about what happens at that high level. One of the important pieces of critique that I have on your project plan is that, from your document, it's not very clear what the criteria is for completion of the project. I'd like to see some sort of plan like the one above which makes it clearer what the state of our i18n work will be when this project is done.

I'm writing up a bunch more thoughts on the technical specs, and I'll share those when we have a thread that's more focused on that discussion.

kgodey:
@seancolsen's plan for splitting up the project and spec makes sense to me.

PR #99, opened 2023-07-20
1 hour quick stab at migration to mkdocs

seancolsen:
After the most recent wiki editing conflict [fiasco](https://matrix.to/#/!kpLRqiyQwsndKhBIvd:matrix.mathesar.org/$lT4QsEX3twddkveSx6F0zNwstVoD8BjHnDvw5WtTQvw?via=matrix.mathesar.org&via=matrix.org), I thought it would be interesting to see how far I could get in **1 hour** of trying to move the wiki to mkdocs.

I got decently far!

image

image

This is just to demonstrate that _**if**_ we wanted to move the wiki to mkdocs, I don't think it would be too painful. I think we'd be in decent shape after 1 day of work, maybe 2 (with some older content remaining in slight disarray).


kgodey:
Thanks for exploring this @seancolsen, this seems like a good idea. Maybe we migrate during the next break between cycles.

Could you  create an issue to track moving the wiki to MkDocs?

seancolsen:
Sure @kgodey I created [Migrate wiki to MkDocs](https://github.com/centerofci/mathesar/issues/3079) and assigned it to myself.

seancolsen:
I'll leave this PR open and assess how many conflicts it has when we do decide to take up the work. I may choose to begin from scratch or to pick up where I left off with this PR.

seancolsen:
Closing since I'll be re-doing this

PR #110, opened 2024-04-01
Permissions spec for Beta

pavish:
This adds the revamped permissions spec for beta to utilize the PostgreSQL permission system.

[Rendered document](https://github.com/mathesar-foundation/mathesar-wiki/blob/permissions-spec/docs/product/specs/permissions-revamp.md).

### DIFF
--- a/docs/product/specs/permissions-revamp.md
+++ b/docs/product/specs/permissions-revamp.md
@@ -14,6 +14,7 @@ This spec describes the revamped product considerations for managing Permissions
 * We will remove the ability to share credentials across databases.
   - We currently have this while creating/configuring a database.
   - We will revisit this feature later down the road when we have requirements from users.
+* Mathesar admins will no longer be able to create a new database on external database servers (servers that do not contain the Django DB).
 
 ## Terminologies
 * User refers to Mathesar user.
@@ -45,9 +46,7 @@ This spec describes the revamped product considerations for managing Permissions
   - This will use the same role utilized by the Django database.
     - The user should be able to create new roles on the internal db and use them if they wanted to.
   - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
-* Mathesar admins should be able to create a new database on external servers, by specifying the same details as the 'pre-existing' database configuration.
-  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
-  - These privileges should be tested and proper errors and rectification steps should be provided to the admin.
+  - (Internal database server refers to the database server that contains the Django DB. It could be on a different host)
 * Pre-defined roles should be automatically created in this setup, similar to pre-existing databases.
 
 ### Role configuration
--- a/docs/product/specs/permissions-revamp.md
+++ b/docs/product/specs/permissions-revamp.md
@@ -3,12 +3,12 @@
 This spec describes the revamped product considerations for managing Permissions within Mathesar. This is targetted towards the Beta release.
 
 ## Goals
-- A. Improve our permissions architecture and user flows to utilize PostgreSQL's permission system for DB objects.
-  * Our current permission handling is custom written and managed entirely within the Mathesar service layer.
-  * Several users have expressed interest towards using PostgreSQL roles, a few even mentioned that it's a blocker for them to start using Mathesar.
-  * Building our permission layer on top of PostgreSQL's permission system is future-proof and robust.
-- B. We should try to maintain feature parity with the flows we currently have.
-- C. Provide a rudimentary UX for ownerships and get users aquainted with the concept.
+1. Improve our permissions architecture and user flows to utilize PostgreSQL's permission system for DB objects.
+	1. Our current permission handling is custom written and managed entirely within the Mathesar service layer.
+	1. Several users have expressed interest towards using PostgreSQL roles, a few even mentioned that it's a blocker for them to start using Mathesar.
+	1. Building our permission layer on top of PostgreSQL's permission system is future-proof and robust.
+1. We should try to maintain feature parity with the flows we currently have.
+1. Provide a rudimentary UX for ownerships and get users aquainted with the concept.
 
 ## Features we intend to remove
 * We will remove the ability to share credentials across databases.
--- a/docs/product/specs/permissions-revamp.md
+++ b/docs/product/specs/permissions-revamp.md
@@ -1,0 +1,115 @@
+# Permissions - Revamp
+
+This spec describes the revamped product considerations for managing Permissions within Mathesar. This is targetted towards the Beta release.
+
+## Goals
+A. Improve our permissions architecture and user flows to utilize PostgreSQL's permission system for DB objects.
+  * Our current permission handling is custom written and managed entirely within the Mathesar service layer.
+  * Several users have expressed interest towards using PostgreSQL roles, a few even mentioned that it's a blocker for them to start using Mathesar.
+  * Building our permission layer on top of PostgreSQL's permission system is future-proof and robust.
+B. We should try to maintain feature parity with the flows we currently have.
+
+## Features we intend to remove
+* We will remove the ability to share credentials across databases.
+  - We currently have this while creating/configuring a database.
+  - We will revisit this feature later down the road when we have requirements from users.
+
+## Terminologies
+* User refers to Mathesar user.
+* Admin refers to Mathesar admins.
+* Role refers to PostgreSQL role/user.
+
+## What we want the user to be able to do
+### Adding/configuring a database
+* A database should be unique. The same database cannot be added multiple times.
+  - This is a change from the existing setup where a database can be added multiple times with different connection credentials.
+#### Pre-existing database
+* Mathesar admins should be able to configure a pre-existing database, using the following details:
+  - DB server Host
+  - DB server Port
+  - Database name
+  - PostgreSQL role -> username and password
+* Mathesar admins should be able to optionally generate pre-defined roles for Manager, Editor, and Viewer, to satisfy Goal (B).
+  - These roles will have similar hierarchial permissions to what we currently have at the database level.
+  - This requires the PostgreSQL role used in the previous step to contain CREATEROLE attribute on the database server.
+  - This step is optional and Mathesar should function normally without these roles.
+#### New database
+* Mathesar admins should be able to create a new database on the internal database server by only entering the database name.
+  - This will use the same role utilized by the Django database.
+    - The user should be able to change this, if they wanted to.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+* Mathesar admins should be able to create a new database on external servers, by specifying the same details as the 'pre-existing' database configuration.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+  - These privileges should be tested and proper errors and rectification steps should be provided to the admin.
+* Pre-defined roles should be automatically created in this setup.
+
+### Role configuration
+* Admins should be able to configure existing roles in a database server, within a database page.
+  - Wherever possible, we should reflect roles that already exist and let the admin add them by specifying the password.
+* Admins should be able to create new roles on the database server and associate them with the database.
+  - This will require the role used by the admin to have CREATEROLE privilege.
+* Future considerations:
+  - Roles exist on the database server and not under a database.
+  - However, inorder to simplify the user flow, we are not introducing a UX dedicated for database servers.
+  - We can explore this option in the future, based on user feedback or product strategy discussions.
+
+### Access control options
+* Admins should be able to provide Mathesar users access to each database configured in Mathesar.
+* Each Mathesar user, including admins, should only be associated with a single role per database.
+  - Admins should be able to update this user-to-role configuration.
+* All Mathesar admins should be able to access all databases configured in Mathesar. Admins cannot be removed from a database.
+  - When a Mathesar admin is newly added into the product, they will automatically be provided the 'default role' for all databases.
+  - The concept of a 'default role' is explained further below in this document.
+* The role for admins can be changed.
+  - This can be restricted to reduce scope and we could mention that admins will always assume the default role.
+  - The challenges are mentioned in further below in the 'Dealing with Ownership and Collaboration issues' section of this document.
+* Standard Mathesar users can be added/removed from a database.
+* Standard Mathesar users will not be able to access a database without being provided explicit access by a Mathesar admin.
+* The permissions for DB objects will be determined by the priviliges of the role the user is associated with.
+* For Mathesar specific objects, we will have a setting per user per database that allows the user to 'read' or 'read-write'.
+  - This applies for explorations and metadata such as table and column settings.
+  - Future consideration:
+    - We can allow an additional option 'Infer from DB privileges' that will automatically identify access to each Mathesar-object based on the privileges on the underlying DB objects.
+
+### Schema level access control options
+* If a user has ownership access (or appropriate GRANT priviliges) to a schema based on their underlying role, they should be able to grant/revoke access to that schema to other roles configured in Mathesar.
+* While granting access to a role, the user can choose access control at three different levels.
+  - manage: USAGE, GRANT, SELECT, INSERT, UPDATE, DELETE, SET and ALTER
+  - edit: USAGE, SELECT, INSERT, UPDATE, DELETE, SET and ALTER
+  - view: USAGE, SELECT
+* This setting will iteratively apply to all objects within the schema.
+* For schemas created using the Mathesar UI, we will automatically grant view access to all configured roles, manage access to the 'default role', and we will provide access accordingly for auto-generated roles.
+* This will not be present for the 'public schema'.
+  - It should be made clear to the user that all users can access the public schema.
+* The UX should make it clear that the roles are the ones being granted access, and should display a list of users per role to make it easier for the user to understand which Mathesar users get access.
+* In the future, this option will be extended to all DB objects such as tables and views.
+
+### Default role
+* There should be one default role per database.
+* When a database is first configured, the role used to connect to the database will be set as the default role.
+* When new admins are added to an existing Mathesar instance which contains a number of databases, they will have access to all those databases by utilizing default roles (which are configured per database).
+* It is also used to deal with ownership issues, mentioned later in this document.
+* The default role cannot be changed. 
+  - We can provide this option in the future.
+  - This restriction is to reduce scope since changing the default role has more considerations such as changing ownerships.
+* The other way to think of this role is as 'Mathesar server's role to connect to the database for operations that do not directly involve a user'.
+
+### Dealing with Ownership and Collaboration issues
+* PostgreSQL has ownership for all DB objects.
+* Tables created by one role cannot be viewed/accessed by another role, unless the access is granted explicitly.
+* When tables/schemas are created via Mathesar UI, it will cause collaboration issues among Mathesar users if they utilize different roles.
+* We will be handling this by following the steps below:
+  * When a DB object is created via the UI, the ownership is transferred to the 'default role'.
+    - This is to maintain a cleaner ownership distinction between objects created using Mathesar and objects created externally.
+    - It would be best for us to prevent changing the 'default role'.
+      - If we allow updating it, we should also automatically transfer ownerships.
+  * The user and their associated role that originally created the object via the UI will be granted all privileges including 'GRANT', but it will not be the owner.
+  * All other priviliges are granted automatically to each role configured within Mathesar.
+  * For auto-generated roles i.e. Manager, Editor, and Viewer, this will be granted in a granular restricted manner based on the role.
+* We will still face collaboration issues when objects are created outside of Mathesar, however, we could handle that by mentioning in our docs to grant access to the 'default role' for any objects users create outside of Mathesar.
+  - This way Mathesar admins will have access to those objects and can grant privilges within Mathesar accordingly.
+
+## Related
+- [Brent's architecture notes](https://github.com/mathesar-foundation/mathesar-wiki/pull/108)
+- [Permissions for Beta - approaches & comparison](https://hackmd.io/@mathesar/Hkads67nT)
+- [Rough figma design to explain permissions approach for Beta](https://www.figma.com/file/S97Mma0hAy5Syh1w85RWoB/Permissions-UX?type=design&node-id=109-98&mode=design)
--- a/docs/product/specs/permissions-revamp.md
+++ b/docs/product/specs/permissions-revamp.md
@@ -8,6 +8,7 @@ This spec describes the revamped product considerations for managing Permissions
   * Several users have expressed interest towards using PostgreSQL roles, a few even mentioned that it's a blocker for them to start using Mathesar.
   * Building our permission layer on top of PostgreSQL's permission system is future-proof and robust.
 - B. We should try to maintain feature parity with the flows we currently have.
+- C. Provide a rudimentary UX for ownerships and get users aquainted with the concept.
 
 ## Features we intend to remove
 * We will remove the ability to share credentials across databases.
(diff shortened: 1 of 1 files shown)


pavish:
@mathemancer I'd particularly like you to go through the [Dealing with Ownership and Collaboration issues](https://github.com/mathesar-foundation/mathesar-wiki/blob/permissions-spec/docs/product/specs/permissions-revamp.md#dealing-with-ownership-and-collaboration-issues) section and the [Default role](https://github.com/mathesar-foundation/mathesar-wiki/blob/permissions-spec/docs/product/specs/permissions-revamp.md#default-role) section, since they add more importance to the default role and require architectural sign-off.

This approach isn't something I'm strongly tied to, my concern is to solve the collaboration issues that would come up when working with multiple roles. If you have a better recommendation, I'm open to it.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+* The permissions for DB objects will be determined by the priviliges of the role the user is associated with.
+* For Mathesar specific objects, we will have a setting per user per database that allows the user to 'read' or 'read-write'.
+  - This applies for explorations and metadata such as table and column settings.
+  - Future consideration:
+    - We can allow an additional option 'Infer from DB privileges' that will automatically identify access to each Mathesar-object based on the privileges on the underlying DB objects.
+
+### Schema level access control options
+* If a user has ownership access (or appropriate GRANT priviliges) to a schema based on their underlying role, they should be able to grant/revoke access to that schema to other roles configured in Mathesar.
+* While granting access to a role, the user can choose access control at three different levels.
+  - manage: USAGE, GRANT, SELECT, INSERT, UPDATE, DELETE, SET and ALTER
```
mathemancer, line 77: To be clear, the table from [relevant docs](https://www.postgresql.org/docs/current/ddl-priv.html) is:

| Privilege    | Applicable Object Types                                                        |
|--------------|--------------------------------------------------------------------------------|
| SELECT       | LARGE OBJECT, SEQUENCE, TABLE (and table-like objects), table column           |
| INSERT       | TABLE, table column                                                            |
| UPDATE       | LARGE OBJECT, SEQUENCE, TABLE, table column                                    |
| DELETE       | TABLE                                                                          |
| TRUNCATE     | TABLE                                                                          |
| REFERENCES   | TABLE, table column                                                            |
| TRIGGER      | TABLE                                                                          |
| CREATE       | DATABASE, SCHEMA, TABLESPACE                                                   |
| CONNECT      | DATABASE                                                                       |
| TEMPORARY    | DATABASE                                                                       |
| EXECUTE      | FUNCTION, PROCEDURE                                                            |
| USAGE        | DOMAIN, FOREIGN DATA WRAPPER, FOREIGN SERVER, LANGUAGE, SCHEMA, SEQUENCE, TYPE |
| SET          | PARAMETER                                                                      |
| ALTER SYSTEM | PARAMETER                                                                      |

The spec you have would imply a manager gets:

| Privilege    | Applicable Object Types                                                        |
|--------------|--------------------------------------------------------------------------------|
| SELECT       | LARGE OBJECT, SEQUENCE, TABLE (and table-like objects), table column           |
| INSERT       | TABLE, table column                                                            |
| UPDATE       | LARGE OBJECT, SEQUENCE, TABLE, table column                                    |
| DELETE       | TABLE                                                                          |
| USAGE        | DOMAIN, FOREIGN DATA WRAPPER, FOREIGN SERVER, LANGUAGE, SCHEMA, SEQUENCE, TYPE |
| SET          | PARAMETER                                                                      |

Minor issues:

- There is no `GRANT` privilege. It's an option on each privilege granted.
- There's no need to grant `SELECT` on table columns specifically.
- There's no need to grant anything on table columns specifically at this time.
- There's no need to grant `USAGE` on `FOREIGN DATA WRAPPER`, `FOREIGN SERVER`, or `LANGUAGE`
- I don't think it makes sense to grant `SET` from Mathesar, since we don't use it.
- _There is no ALTER privilege_ !!!

Major problem:

Dropping and altering database objects are privileges exclusive to the owner of the object, and can't be granted except by granting membership in the owning role. This is the main thing that makes me hesitant to try to get schema-level privileges handled by the beta release. Getting that right would mean (at a minimum):

- More emphasis on creating roles exclusively for use as groups, and
- More UI/UX work around this ownership concept.

I don't think we can do it quickly enough.

The same follows for the "Edit" role on schema level. "Viewer" should be possible on a schema level.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  - edit: USAGE, SELECT, INSERT, UPDATE, DELETE, SET and ALTER
+  - view: USAGE, SELECT
+* This setting will iteratively apply to all objects within the schema.
+* For schemas created using the Mathesar UI, we will automatically grant view access to all configured roles, manage access to the 'default role', and we will provide access accordingly for auto-generated roles.
+* This will not be present for the 'public schema'.
```
mathemancer, line 82: We should discuss in a call the changes around permissions granted to the `PUBLIC` group in PG 16.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  - We can provide this option in the future.
+  - This restriction is to reduce scope since changing the default role has more considerations such as changing ownerships.
+* The other way to think of this role is as 'Mathesar server's role to connect to the database for operations that do not directly involve a user'.
+
+### Dealing with Ownership and Collaboration issues
+* PostgreSQL has ownership for all DB objects.
+* Tables created by one role cannot be viewed/accessed by another role, unless the access is granted explicitly.
+* When tables/schemas are created via Mathesar UI, it will cause collaboration issues among Mathesar users if they utilize different roles.
+* We will be handling this by following the steps below:
+  * When a DB object is created via the UI, the ownership is transferred to the 'default role'.
```
mathemancer, line 102: What about simply granting any role created or managed via Mathesar to the 'default role'? If you transfer ownership to the 'default role', you won't be able to `ALTER` or `DROP` a table you yourself created, regardless of your other privileges.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+    - This is to maintain a cleaner ownership distinction between objects created using Mathesar and objects created externally.
+    - It would be best for us to prevent changing the 'default role'.
```
mathemancer, line 104: Agreed, at least for beta.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+      - If we allow updating it, we should also automatically transfer ownerships.
```
mathemancer, line 105: I don't want to do this for beta. It's super complicated.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  * The user and their associated role that originally created the object via the UI will be granted all privileges including 'GRANT', but it will not be the owner.
```
mathemancer, line 106: They still wouldn't be able to `ALTER` or `DROP`.

mathemancer (changes requested):
I left many specific comments.

My big overarching comment is that `ALTER` and `DROP` are controlled by ownership, and can't be granted except by granting ownership, or granting access to act as the role that owns the object in question. I think we should consider having a call to go through the details of that. I also think we should avoid trying to squeeze schema-level access in before beta.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Default role
+* There should be one default role per database.
+* When a database is first configured, the role used to connect to the database will be set as the default role.
+* When new admins are added to an existing Mathesar instance which contains a number of databases, they will have access to all those databases by utilizing default roles (which are configured per database).
+* It is also used to deal with ownership issues, mentioned later in this document.
+* The default role cannot be changed. 
+  - We can provide this option in the future.
+  - This restriction is to reduce scope since changing the default role has more considerations such as changing ownerships.
+* The other way to think of this role is as 'Mathesar server's role to connect to the database for operations that do not directly involve a user'.
```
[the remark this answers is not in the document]
seancolsen: It would be nice to list these examples in the doc.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+* PostgreSQL has ownership for all DB objects.
+* Tables created by one role cannot be viewed/accessed by another role, unless the access is granted explicitly.
+* When tables/schemas are created via Mathesar UI, it will cause collaboration issues among Mathesar users if they utilize different roles.
+* We will be handling this by following the steps below:
+  * When a DB object is created via the UI, the ownership is transferred to the 'default role'.
+    - This is to maintain a cleaner ownership distinction between objects created using Mathesar and objects created externally.
+    - It would be best for us to prevent changing the 'default role'.
+      - If we allow updating it, we should also automatically transfer ownerships.
+  * The user and their associated role that originally created the object via the UI will be granted all privileges including 'GRANT', but it will not be the owner.
+  * All other priviliges are granted automatically to each role configured within Mathesar.
```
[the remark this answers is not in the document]
seancolsen: > All other priviliges are granted automatically to each role configured within Mathesar.

I would also like clarity on this. Can you give some examples?

docs/product/specs/permissions-revamp.md
seancolsen: > wouldn't be able to `ALTER` or `DROP`

I think this might be okay. 

There is potential for users to end up in a situation where they can create things but not delete them. That would be confusing. We want to avoid that situation.

As I see it, we basically want to steer people towards using the default role if they're trusted with the privilege of creating or deleting things.

I would see us doing this steering by providing clear in-app documentation that makes this behavior apparent to users as early in their flows as possible. Like, we'd need to make Mathesar admins aware of this behavior when they are configuring the roles for a DB.

docs/product/specs/permissions-revamp.md
seancolsen: > hesitant to try to get schema-level privileges handled by the beta

Me too

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+* Mathesar admins should be able to create a new database on external servers, by specifying the same details as the 'pre-existing' database configuration.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+  - These privileges should be tested and proper errors and rectification steps should be provided to the admin.
+* Pre-defined roles should be automatically created in this setup.
+
+### Role configuration
+* Admins should be able to configure existing roles in a database server, within a database page.
+  - Wherever possible, we should reflect roles that already exist and let the admin add them by specifying the password.
+* Admins should be able to create new roles on the database server and associate them with the database.
```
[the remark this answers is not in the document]
seancolsen: > With our newly-merged models, the "associate them with the database" part is a bit wonky.

This is part of the point I was trying to make with my [critique](https://github.com/mathesar-foundation/mathesar-wiki/pull/108#discussion_r1544731849) of the models.

(Not trying to beat a dead horse — just trying to connect some dots.)

seancolsen (commented):
This document is immensely helpful. Thank you @pavish.

Generally, I think this all looks great.

My one small hesitation at the product level is this:

- This spec has a limitation which I don't love: it's hard to create a new database on a remote server because I need to enter a password.

    For the "admin interface" kind of product, this limitation is perfectly acceptable.
    
    However, during our [strategy meeting](https://hackmd.io/I8PrhOcWTg2NLtcPp34rGQ), several people expressed interest in continuing to design for (and market towards) the "from scratch" use case where people use Mathesar as a relational spreadsheet that stands on its own, like Airtable. For that use case I think Mathesar needs a product-level abstraction akin to an Airtable "Base", i.e. a self-contained project that can be easily created, managed, and destroyed. Without schema-level permissions (unlikely by beta, it seems), then the "project" abstraction might tend to become a "database" in people's minds. If they use the internal DB server, then they can easily create new databases, but I think that those databases will basically end up becoming "global" because there's not a password-free way to create them using more granular ownership and permission. For example, it would be quite cumbersome to create a new DB that only I can see. I might want to do this just to experiment without cluttering up other people's experience in Mathesar. And if for some reason they're not using the internal DB (e.g. SQLite internal DB, or they want managed backups for their user data), then they can't even create new databases in a password-free way.
    
    These limitations make Mathesar less compelling for the stand-alone "relational spreadsheet" use case. That's okay with me because my inclination is to focus on the "admin interface" type of product for now anyway. But I just want to flag this as a concern since other people didn't seem as keen to focus exclusively on the admin interface approach in the short term.

Also I have a side note on the sequence of planning:

- Throughout our permissions conversations, Pavish and I have had a hard time reconciling what has appeared to be different approaches to product design: I've been pushing for understanding the models sooner, and Pavish has been pushing for understanding the user experience sooner. Now that we have this product spec, I'll say that _this_ is exactly the kind of design that I need to understand _before_ understanding the models. My personal approach is (1) product spec, (2) data models, (3) UX. We can debate semantics on what sort of design actually counts as "UX", but point is that (for me) the document in this PR takes precedent over Brent's [models](https://wiki.mathesar.org/engineering/architecture/models/) doc.

    What I'm suggesting here is that we work towards agreeing first on this product spec and then (subsequently) ensure our models are appropriately designed to implement it. I make this suggestion because my opinion is that we still have significant architectural incompatibilities between this product design and our planned models, but I don't want to raise that critique here. I'd rather agree on this product design on its own — with zero dependencies — and then base my critique of the models on this product design.

docs/product/specs/permissions-revamp.md
mathemancer: I'm thinking the other way around in this case. If we configure the role `brent` to be used from Mathesar, then we'd `GRANT brent TO default_role`, not `GRANT default_role TO brent`.

With that said, I had a chat with Pavish today, and my (our) thinking has evolved a bit.

seancolsen: > I'm thinking the other way around in this case

Ah right. Sorry — I misread your comment.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  - This requires the PostgreSQL role used in the previous step to contain CREATEROLE attribute on the database server.
+  - This step is optional and Mathesar should function normally without these roles.
+#### New database
+* Mathesar admins should be able to create a new database on the internal database server by only entering the database name.
+  - This will use the same role utilized by the Django database.
+    - The user should be able to change this, if they wanted to.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+* Mathesar admins should be able to create a new database on external servers, by specifying the same details as the 'pre-existing' database configuration.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+  - These privileges should be tested and proper errors and rectification steps should be provided to the admin.
```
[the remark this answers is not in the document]
pavish: Yes, the privileges can be tested lazily. My intention is to specify that the errors are handled properly and the user knows whats happening.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  - Database name
+  - PostgreSQL role -> username and password
+* Mathesar admins should be able to optionally generate pre-defined roles for Manager, Editor, and Viewer, to satisfy Goal (B).
+  - These roles will have similar hierarchial permissions to what we currently have at the database level.
+  - This requires the PostgreSQL role used in the previous step to contain CREATEROLE attribute on the database server.
+  - This step is optional and Mathesar should function normally without these roles.
+#### New database
+* Mathesar admins should be able to create a new database on the internal database server by only entering the database name.
+  - This will use the same role utilized by the Django database.
+    - The user should be able to change this, if they wanted to.
```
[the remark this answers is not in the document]
pavish: > Should they just be able to create other roles on the internal DB server in general?

Yes, that's correct.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+* Mathesar admins should be able to create a new database on external servers, by specifying the same details as the 'pre-existing' database configuration.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+  - These privileges should be tested and proper errors and rectification steps should be provided to the admin.
+* Pre-defined roles should be automatically created in this setup.
+
+### Role configuration
+* Admins should be able to configure existing roles in a database server, within a database page.
```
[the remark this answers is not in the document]
pavish: > "configure" in this case just means add them to the relevant Mathesar models, correct?

Yes, that's correct.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: > What, precisely, do you mean by that?

I meant that the newly created roles should have CONNECT, and CREATE to the database. They should also have the ability to SELECT the existing schemas and tables within the database.

I will make this clearer in the spec.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Access control options
+* Admins should be able to provide Mathesar users access to each database configured in Mathesar.
+* Each Mathesar user, including admins, should only be associated with a single role per database.
+  - Admins should be able to update this user-to-role configuration.
+* All Mathesar admins should be able to access all databases configured in Mathesar. Admins cannot be removed from a database.
+  - When a Mathesar admin is newly added into the product, they will automatically be provided the 'default role' for all databases.
+  - The concept of a 'default role' is explained further below in this document.
+* The role for admins can be changed.
+  - This can be restricted to reduce scope and we could mention that admins will always assume the default role.
```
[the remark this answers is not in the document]
pavish: If the role for admins can be changed to something other than 'default role', we would have to deal with ownership and collaboration issues. For Beta atleast, I don't think we should allow that.

A related point is here: https://github.com/mathesar-foundation/mathesar-wiki/pull/110#discussion_r1546929840

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+* Standard Mathesar users can be added/removed from a database.
+* Standard Mathesar users will not be able to access a database without being provided explicit access by a Mathesar admin.
+* The permissions for DB objects will be determined by the priviliges of the role the user is associated with.
+* For Mathesar specific objects, we will have a setting per user per database that allows the user to 'read' or 'read-write'.
+  - This applies for explorations and metadata such as table and column settings.
+  - Future consideration:
+    - We can allow an additional option 'Infer from DB privileges' that will automatically identify access to each Mathesar-object based on the privileges on the underlying DB objects.
+
+### Schema level access control options
+* If a user has ownership access (or appropriate GRANT priviliges) to a schema based on their underlying role, they should be able to grant/revoke access to that schema to other roles configured in Mathesar.
```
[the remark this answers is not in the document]
pavish: The role should just be visible in Mathesar i.e. it should have been saved in the relevant Mathesar models.

docs/product/specs/permissions-revamp.md
pavish: @mathemancer and I had a call to discuss this spec in general and we agreed that manager would directly mean the 'default role', which is also used by the admins.

For any custom role creation, we'd only consider viewer and editor.

Schema-level privileges, however, do require a concept of ownership to make it easier for the user to understand and to accurately depict what's happening in the DB.

We are having another call where we will discuss this more.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: This was under the assumption that we'd allow admins to have different roles. From the last call I had with @mathemancer, our thoughts are to always use only the 'default role' for the admins.

I will update the spec after we discuss this more.

ghislaineguerin (approved):
Thanks @pavish . I don't have any question or comments for now. This seems clear enough for me to get started.

pavish:
@ghislaineguerin There were further discussions that took place regarding ownerships which aren't updated in the spec yet. I'll update it and request review again.

pavish:
@mathemancer @kgodey @seancolsen I've updated the spec based on the recent DB ownership discussion. Here's the commit: https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

Please take a look.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: I believe this is resolved.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: This seems resolved.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: This is mentioned more clearly in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: We've introduced displaying the ownership for objects, and this is no longer relevant.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: This seems resolved.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+* While granting access to a role, the user can choose access control at three different levels.
+  - manage: USAGE, GRANT, SELECT, INSERT, UPDATE, DELETE, SET and ALTER
+  - edit: USAGE, SELECT, INSERT, UPDATE, DELETE, SET and ALTER
+  - view: USAGE, SELECT
+* This setting will iteratively apply to all objects within the schema.
```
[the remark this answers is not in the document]
pavish: This is no longer relevant.

We've introduced displaying the ownership for objects and there are changes revolving granting of privileges in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

docs/product/specs/permissions-revamp.md
pavish: We've introduced displaying the ownership for objects and there are changes revolving granting of privileges in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

docs/product/specs/permissions-revamp.md
pavish: Mentioned in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: Mentioned in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

docs/product/specs/permissions-revamp.md
pavish: This is no longer relevant.

We've introduced displaying the ownership for objects and there are changes revolving granting of privileges in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

docs/product/specs/permissions-revamp.md
pavish: No longer relevant.

docs/product/specs/permissions-revamp.md
pavish: We've introduced displaying the ownership for objects and there are changes revolving granting of privileges in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

docs/product/specs/permissions-revamp.md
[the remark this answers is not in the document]
pavish: No longer relevant.

We've introduced displaying the ownership for objects and there are changes revolving granting of privileges in https://github.com/mathesar-foundation/mathesar-wiki/commit/2ad0d7262212be9b5897deda8b7962e99d2f7191.

seancolsen (approved):
Looks good @pavish

seancolsen:
Sorry... one more thought/question/suggestion that's been nagging at me...

A couple weeks ago we (Sean/Brent/Pavish) had an impromptu call to discuss my [concern](https://github.com/mathesar-foundation/mathesar-wiki/pull/108#discussion_r1544547713) about the data models. Near the end of the call I asked Pavish a couple questions which helped clear up a misunderstanding I had about Pavish's intended product-level functionality.

Our dialog went something like this:

- **Sean:** So lets say I use Mathesar to connect to 10 different Postgres DBs on the same server with the same Postgres role. How many entries should Mathesar have for that Postgres role?

- **Pavish:** Ten.

- **Sean:** Ok. Now let's say I've changed the password for that role in Postgres. How many times do I need to update the password in Mathesar?

- **Pavish:** Ten.

- **Sean:** Ok.

To be clear: this product design is perfectly fine with me. But it was definitely _not_ how I thought the permissions system would work until we had that call.

I'd like to make sure everyone is on the same page about this behavior, and I think it would help to put something in this PR. I'm proposing that we add the following section to the end of the `permissions-revamp.md` document:

## Encapsulation

Within Mathesar, configuration of a Postgres server and its roles should be performed _within the context of a database_. All such configuration should be siloed to that database. For example:

- Within a database, Mathesar should only present Postgres roles which have explicitly configured _for that specific database_. Mathesar should hide roles which have been configured for other databases, even if those other databases exist on the same server.

- Changing the password stored in Mathesar for a Postgres role should only affect the Postgres role configured _within one database_. If the same role is used for multiple databases, then the password will need to be changed multiple times.

- Changing the port number configured for one Mathesar user database should not affect the port number configured for other Mathesar user databases. Like passwords, updates to port numbers will need to be performed multiple times.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,146 @@
...
+* For Mathesar specific objects, we will have a setting per user per database that allows the user to 'read' or 'read-write'.
+  - This applies for explorations and metadata such as table and column settings.
+  - Future consideration:
+    - We can allow an additional option 'Infer from DB privileges' that will automatically identify access to each Mathesar-object based on the privileges on the underlying DB objects.
+
+#### Role for admins
+* An admin is not associated with a single role.
+* Admins should be able to utilize any role that's configured for a database.
+* When an admin opens a DB object, lets say a table,
+  - If the owning role of that table is configured, then any operation performed by the admin will utilize that role.
```
mathemancer, line 87: I don't think we should do this by default. It makes configuring any owning role in Mathesar a dangerous proposition. It also ruins any chance of auditing user actions on the underlying database.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,146 @@
...
+  - If the owing role isn't configured, we'll try to figure out the best role for the admin to use. In cases where it's not possible, we will use the `default` role.
+* More information is provided in the 'Default role' and 'Dealing with Ownership and Collaboration issues' sections of this document.
+
+### Default role
+* There should be one default role per database.
```
mathemancer, line 92: To double-check: This _doesn't_ imply creating a new default role for each database, correct? I.e., we could have a DB server with only one role, used as default for each DB?

mathemancer: Or do you mean the credential model instance here? In that case, we really need to consider how to present this in a non-confusing way to users. Are they supposed to understand that multiple roles in the Mathesar UI all map to the same actual role on the underlying database (server)?

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,146 @@
...
+* When a database is first configured, the role used to connect to the database will be set as the default role.
+  - This role is used for installing/upgrading the functions on the DB.
+* The default role cannot be removed.
```
mathemancer, line 95: This implies you mean the credential instance. If correct, we should make that obvious in the document.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,146 @@
...
+  - For existing objects, we'll provide an option to iteratively provide privileges to these roles.
+    - This will however depend on ownerships of each object and can fail.
+  - We will no longer have the manager role as it would be the same as having admin access.
+    - Future consideration: We could provide a concept of database admins as a Mathesar level access control option per database. This will help maintain feature parity with the 'DB level admins' feature we currently have.
+#### New database
+* Mathesar admins should be able to create a new database on the internal database server by only entering the database name.
+  - This will use the same role utilized by the Django database.
+    - The user should be able to create new roles on the internal db and use them if they wanted to.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
+* Mathesar admins should be able to create a new database on external servers, by specifying the same details as the 'pre-existing' database configuration.
```
mathemancer, line 48: I know we've already discussed this, but this walkthrough makes the namespacing of a role under a database more obviously overcomplicated. Why can't the admin who has a connection to the external server already established (with sufficient privileges) just create a new DB by entering a database name?

It's also confusing for the user. They'll have to enter a previous database, as well as credentials that have privileges to connect to _that_ database (along with appropriate `CREATEDATABASE` privileges. If I were a user, I'd be trying to figure out:
- Why I had to enter an arbitrary preexisting database (You could enter any that the role in question has `CONNECT` privileges on)
- What, exactly, am I creating here?
  - the database (this one is obvious)
  - A new role?
  - A copy of the role?
Even more questions arise for a user who _does_ understand how roles, etc. work in PostgreSQL, since they'll be naturally confused as to why we're simulating a bunch of roles when there's actually only one.

After reading through this in its current form, I'm even more convinced that trying to namespace roles under databases in the users' minds is a complicating rather than simplifying abstraction.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,146 @@
...
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
```
mathemancer, line 49: What does `CREATEDATABASE` even mean as a privilege if roles are subordinate to databases (in the UI, and in users' minds)? I.e., how is an admin user supposed to think about this if we're avoiding the concept of a server that has both roles and databases?

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Dealing with Ownership and granting access
+#### Considerations
+* PostgreSQL has ownership for all DB objects.
+* Tables/schemas created by one role cannot be viewed/accessed by another role, unless the access is granted explicitly.
+* When tables/schemas are created via Mathesar UI, it may cause collaboration issues among Mathesar users if they utilize different roles.
+
+#### Approach for tables
+* When a table is created via the UI, the ownership will belong to the role of the Mathesar user that created that object.
+  * Admins will have access to all roles on the DB and hence will be able to use the role that created the table and obtain full ownership.
```
mathemancer, line 109: I would say they're acting as the owner. They're not really obtaining anything, and we need to make sure they understand that in the UI. They'll get the wrong understanding if they think acting on the table somehow takes ownership from the user who created the table.

mathemancer (changes requested):
The long and short of it is: No matter how I try to see the namespacing of a role under a database as a simplifying abstraction, it just seems to make things more complicated, difficult, and confusing.

mathemancer:
> ## Encapsulation
> Within Mathesar, configuration of a Postgres server and its roles should be performed _within the context of a database_. All such configuration should be siloed to that database. For example:
> 
> * Within a database, Mathesar should only present Postgres roles which have explicitly configured _for that specific database_. Mathesar should hide roles which have been configured for other databases, even if those other databases exist on the same server.
> * Changing the password stored in Mathesar for a Postgres role should only affect the Postgres role configured _within one database_. If the same role is used for multiple databases, then the password will need to be changed multiple times.
> * Changing the port number configured for one Mathesar user database should not affect the port number configured for other Mathesar user databases. Like passwords, updates to port numbers will need to be performed multiple times.

Why attach it to the database in this way? It seems to add a dependency without any value. I can see the value of listing the roles that were configured in a given context, but the rest of the abstraction seems to add complexity.

docs/product/specs/permissions-revamp.md
pavish: I updated an older version of the spec I had and forgot to remove this point. Thanks for pointing this out, I'll remove this.

Mathesar admins **should not** be able to create a new database on external servers.

We decided to remove the feature of being able to add a new database on external servers and sharing credentials. When we try to introduce it in the future, we can figure out a way of introducing the concept of database servers.

docs/product/specs/permissions-revamp.md
pavish: > This doesn't imply creating a new default role for each database, correct? I.e., we could have a DB server with only one role, used as default for each DB?

Yes, this is correct.

docs/product/specs/permissions-revamp.md
pavish: Here's the commit https://github.com/mathesar-foundation/mathesar-wiki/pull/110/commits/189b77285da6da17d2179f28c83d6233d2d557af that should make this clear.

docs/product/specs/permissions-revamp.md
pavish: This is no longer relevant after https://github.com/mathesar-foundation/mathesar-wiki/pull/110/commits/189b77285da6da17d2179f28c83d6233d2d557af

pavish:
@seancolsen and @mathemancer

1. Regarging Encapsulation, it is a decision I want to take during the UX design phase and not solidify as a product level spec.
    - In our discussions regarding the [concern](https://github.com/mathesar-foundation/mathesar-wiki/pull/108#discussion_r1544547713) raised by Sean, my stance was that we keep it flexible so that we'd have more freedom to experiement with the UX.
1. I do not want the ability to share credentials for the beta and I would like the role configuration panels to be present within the context of a 'Database page', as stated in Sean's comment.
1. I would however like to see if we can make it clear to the user in the UI that the roles are created in the underlying database server, even if we do not have a dedicated UX for managing database servers.

I'd say let's move this particular discussion to the UX design sessions.

seancolsen:
@pavish When we release Mathesar beta, perhaps it will function as I've described within the "encapsulation" section above. Perhaps not. There could be subtle differences from what I've laid out, or drastic differences. I'll call this the "encapsulation question". _"What is our approach to encapsulating server and role config within a database?"_ — that's the encapsulation question. I want to answer that question immediately. You seem to want to answer that question later in our design and development process. I don't care whether we classify the encapsulation question as "Product design" or "UX design"; as far as I'm concerned, that classification is a matter of semantics. But I do feel very strongly that kicking this can down the road will introduce significant risk implementation delays before beta. The sooner we figure this out, the faster we can build. Why can't we answer the encapsulation question out now? What's blocking us from making a decision on that?

pavish:
@seancolsen 

> The sooner we figure this out, the faster we can build. 

I agree.

> Why can't we answer the encapsulation question out now? What's blocking us from making a decision on that?

I thought that the question was answered already. Things would work like you've mentioned in your [comment](https://github.com/mathesar-foundation/mathesar-wiki/pull/110#issuecomment-2048749443). My recent [comment](https://github.com/mathesar-foundation/mathesar-wiki/pull/110#issuecomment-2051676373) was related to documenting it in the product spec instead of thinking of it as a UX concern. 

I'd like @ghislaineguerin to work on the UX, ask questions we might have not thought about yet, and present a preliminary barebones UX flow so that we can eliminate any remaining confusion.

Mainly, I'd like to prevent any questions later down the road, like Brent's [recent comment](https://github.com/mathesar-foundation/mathesar-wiki/pull/110#issuecomment-2051451205), and I think seeing the UX design would help than discussing it further on the doc. I'm good with updating the spec if you feel strongly about it.

We're not going to be kicking this down the road till implementation, the UX design phase is the immediate next step which Ghislaine has already started work on. I've added Ghislaine to the "Technical beta check in" call on Tuesday so that we can go through the UX flow.

seancolsen:
@pavish 

> I thought that the question was answered already.

If this question _is_ answered already, then I think it would be very useful to have it in the spec. But it's not clear to me that everyone agrees on one unified approach to encapsulation. If we have disagreement, then I don't see the question as being answered. Just to reiterate: I'm personally staking out a neutral position on the encapsulation question. I'm perfectly fine with many different approaches to encapsulation. What I _am_ pushing for is greater clarity on this question, because it will allow me to better tailor my critique towards the other aspects of our permissions plans and designs.

It sounds like you think we need to see a "preliminary barebones UX flow" before we can get the level of clarity that I want. Personally I don't need to see that; I'd be inclined to make decisions about encapsulation before doing any wire-framing or mockups, just for the sake of speed. But that's only an inclination, not a strong opinion. If other people need to see wire-frames or mockups in order to form opinions and come to agreements, then I'm okay with that — so long as we do it very soon. It sounds like the next steps you have in-mind are for us to look at wire-frames or mockups on Tuesday and then make sure we're in agreement on our approach to encapsulation. That's fine with me, but I'd like to request that the design material be sent out in advance of that meeting so that Brent and I have a chance to review it thoroughly before the call. And I would request that we prioritize seeking agreement on this encapsulation question at the start of the meeting, before diving into detail-level conversation about the myriad other UX questions and concerns that I predict we'll all inevitably have. I'd expect that call to take at least 1.5 hr.

pavish:
@seancolsen I've added encapsulation to the spec in this commit: https://github.com/mathesar-foundation/mathesar-wiki/pull/110/commits/e1356078ecbb29aa1164136fa187a97eedb030d0.

My initial hesitation towards adding your [comment](https://github.com/mathesar-foundation/mathesar-wiki/pull/110#issuecomment-2048749443) to the spec was also due to how it was phrased. I've modified it to clearly mention that this is how we'd like the user experience to look like, and not necessarily as a rule of thumb on how the models and code should be structured. I hope this clears things up regarding encapsulation.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
+# Permissions - Revamp
+
+This spec describes the revamped product considerations for managing Permissions within Mathesar. This is targetted towards the Beta release.
+
+## Goals
+1. Improve our permissions architecture and user flows to utilize PostgreSQL's permission system for DB objects.
```
kgodey, line 6: @pavish please note I updated this list's formatting since the combination of the bullets + lettered list just seemed wrong to me somehow.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
...
+	1. Our current permission handling is custom written and managed entirely within the Mathesar service layer.
+	1. Several users have expressed interest towards using PostgreSQL roles, a few even mentioned that it's a blocker for them to start using Mathesar.
+	1. Building our permission layer on top of PostgreSQL's permission system is future-proof and robust.
+1. We should try to maintain feature parity with the flows we currently have.
+1. Provide a rudimentary UX for ownerships and get users aquainted with the concept.
+
+## Features we intend to remove
```
kgodey, line 13: aren't we also removing the ability for users to have DDL permissions to a schema or DB through the UI?

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
...
+* We will remove the ability to share credentials across databases.
+  - We currently have this while creating/configuring a database.
+  - We will revisit this feature later down the road when we have requirements from users.
+* Mathesar admins will no longer be able to create a new database on external database servers (servers that do not contain the Django DB).
+
+## Terminologies
+* User refers to Mathesar user.
```
kgodey, line 20: Can we put the terms in quotation marks for readability?

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
...
+  - We should make it clear that these roles & their privileges would only apply for new objects created via the Mathesar UI.
+  - For existing objects, we'll provide an option to iteratively provide privileges to these roles.
+    - This will however depend on ownerships of each object and can fail.
+  - We will no longer have the manager role as it would be the same as having admin access.
+    - Future consideration: We could provide a concept of database admins as a Mathesar level access control option per database. This will help maintain feature parity with the 'DB level admins' feature we currently have.
+#### New database
+* Mathesar admins should be able to create a new database on the internal database server by only entering the database name.
+  - This will use the same role utilized by the Django database.
+    - The user should be able to create new roles on the internal db and use them if they wanted to.
+  - The role used should contain both CREATEDATABASE and CREATEROLE privileges.
```
kgodey, line 48: Another minor formatting thing, I would put the privilege names in backticks for readability.

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
...
+* We'll have separate onboarding instructions/flows in the UI for pre-existing and new databases.
+
+#### Pre-existing database
+* Mathesar admins should be able to configure a pre-existing database, using the following details:
+  - DB server Host
+  - DB server Port
+  - Database name
+  - PostgreSQL role -> username and password
+* Mathesar admins should be able to optionally generate pre-defined roles for Editor, and Viewer, to partially satisfy Goal (B).
+  - This requires the PostgreSQL role used in the previous step to contain CREATEROLE attribute on the database server.
```
kgodey, line 37: How are we planning to communicate this requirement?

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
...
+  - This applies for explorations and metadata such as table and column settings.
+  - Future consideration:
+    - We can allow an additional option 'Infer from DB privileges' that will automatically identify access to each Mathesar-object based on the privileges on the underlying DB objects.
+
+#### Role for admins
+* An admin is not associated with a single role.
+* Admins should be able to utilize any role that's configured for a database.
+* When an admin opens a DB object, lets say a table,
+  - If the owning role of that table is configured, then any operation performed by the admin will utilize that role.
+  - If the owing role isn't configured, we'll try to figure out the best role for the admin to use. In cases where it's not possible, we will use the `default` role.
```
kgodey, line 87: How will we figure this out?

docs/product/specs/permissions-revamp.md
```diff
@@ -0,0 +1,153 @@
...
+  - See 'Role for admins' section on how roles are used by Admins.
+* Admins cannot be removed from a database.
+* Standard Mathesar users can be added/removed from a database.
+* Standard Mathesar users will not be able to access a database without being provided explicit access by a Mathesar admin.
+
+#### Permission checks for objects
+* The permissions for DB objects will be determined by the priviliges of the role the user is associated with.
+* For Mathesar specific objects, we will have a setting per user per database that allows the user to 'read' or 'read-write'.
+  - This applies for explorations and metadata such as table and column settings.
+  - Future consideration:
```
kgodey, line 79: I think the spec would be easier to read if we put all future considerations into a separate section at the end rather than within other sections. It will also allow us to find our future plans easier.

kgodey (commented):
I left some comments (feel free to ignore the formatting related ones, they're just suggestions). I am confused about the state of the spec, though, given the ongoing discussion about encapsulation.

docs/product/specs/permissions-revamp.md
pavish: I had the lettered list to be able to reference them later in the docs, like:
> to satisfy Goal (B)

It's only mentioned in one place, I can update the text there.

pavish:
I'm going to go ahead and merge this PR in. The spec is updated with the latest agreed upon product spec (the periwinkle proposal). I do not think this requires any further review since everyone is in agreement.

PR #111, opened 2024-04-09
RPC client plans

seancolsen:
Here's what I have in-mind for how the client side of the JSON-RPC API would work.

**[Rendered](https://github.com/mathesar-foundation/mathesar-wiki/blob/rpc_client/docs/projects/2024/architecture-transition/rpc-client.md)**

### DIFF
--- a/docs/projects/2024/architecture-transition/rpc-client.md
+++ b/docs/projects/2024/architecture-transition/rpc-client.md
@@ -1,0 +1,153 @@
+# JSON-RPC client plans
+
+## Examples of how the API would be called
+
+### Non-batched, non-cancellable
+
+```ts
+/** @type {ApiRequest<Spacecraft>>} */
+const request = api.spacecrafts.get({ id: 42 });
+
+/** @type {Promise<Spacecraft>} */
+const promise = request.run();
+
+/** @type {Spacecraft} */
+const spacecraft = await promise; // Might throw!
+```
+
+### Non-batched, cancellable
+
+```ts
+/** @type {ApiRequest<Spacecraft>>} */
+const request = api.spacecrafts.get({ id: 42 });
+
+/** @type {CancellablePromise<Spacecraft | undefined>} */
+const promise = request.cancellableRun();
+
+/**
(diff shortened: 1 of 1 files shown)


seancolsen:
@pavish Can you review these plans and let me know if you have any concerns? They're loose plans at the moment. I would expect some things within "Lower level types" to change. That section is mostly there to help supplement the top section. The "Examples of how the API would be called" is what I want you to focus on. If you're okay with this, then I'd expect the structure of that code to mostly follow what I have here.

docs/projects/2024/architecture-transition/rpc-client.md
```diff
@@ -0,0 +1,153 @@
+# JSON-RPC client plans
+
+## Examples of how the API would be called
+
+### Non-batched, non-cancellable
```
pavish, line 5: Why do we need `non-cancellable` requests?

docs/projects/2024/architecture-transition/rpc-client.md
```diff
@@ -0,0 +1,153 @@
...
+
+```ts
+/** @type {ApiRequest<Spacecraft>>} */
+const request = api.spacecrafts.get({ id: 42 });
+
+/** @type {CancellablePromise<Spacecraft | undefined>} */
+const promise = request.cancellableRun();
+
+/**
+ * Resolves to `undefined` if canceled.
```
pavish, line 28: Cancelled requests should not resolve or reject at higher layers. They definitely should not resolve to `undefined`.

The idea behind a cancelled request is that any layer of top of it should not have to worry about it anymore. Resolving to `undefined` makes it a problem of the higher layers instead of dealing with it within CancellablePromise.

pavish (changes requested):
@seancolsen I have some comments regarding cancellable requests that we should address before proceeding.

seancolsen:
@pavish and I had a quick call about this. I had misunderstood the way our `CancellablePromise` works.

I've updated the spec now. It's much simpler! Yay!

Please re-review, @pavish

pavish (approved):
Looks good to me!

PR #112, opened 2024-04-15
Resolve remaining questions within RPC plans

seancolsen:
This PR makes changes to our RPC plans document to resolve some of the outstanding questions by proposing answers to them.

[Rendered doc](https://github.com/mathesar-foundation/mathesar-wiki/blob/rpc_plans/docs/projects/2024/architecture-transition/rpc.md)

Specifically:

- Timeline is clarified to indicate that we don't intend to transition _all_ REST endpoints to RPC by the beta release.

- `rpc-client.md` is updated to specify that we will manually maintain TypeScript types and we won't attempt to auto-generate them.

- We'll avoid attempting to transition the data_files endpoint to RPC any time soon.

- We'll use "fixed ids", keeping the scope of unique id values limited to a single HTTP request.

- We'll use the standard verbs: **list**, **get**, **add**, **patch**, **replace**, and **delete**.

- We'll allow arbitrarily deep namespacing.

- I removed questions about response structure and error structure, thinking that we'll basically resolve this as we go by establishing precedent. Perhaps we'll choose to document this later, but I'm not sure.

- I removed the "casing transformation" and "error types" questions because I think the answer is basically "we don't want to spend time on this", including tiding up and properly documenting our decision process. Basically this decision process ocurred within my head and I don't think it's worth the time to capture it.

- I filled in names for all of the RPC functions.

### DIFF
--- a/docs/projects/2024/architecture-transition/rpc.md
+++ b/docs/projects/2024/architecture-transition/rpc.md
@@ -6,7 +6,7 @@ It is maintained by Sean and Brent.
 
 ## Timeline
 
-We want to be fully transitioned to RPC APIs by the **beta release**. Ideally we would not have any REST-based APIs at that point.
+We want _most_ (but not all) of our REST endpoints to be fully transitioned to RPC APIs by the **beta release**. Some REST endpoints will likely remain, especially ones like `data_files` that will be difficult (or impossible?) to transition to JSON-RPC.
 
 ## JSON-RPC spec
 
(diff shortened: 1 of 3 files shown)


seancolsen:
@pavish @mathemancer please review to make sure you're on board with the small decisions I'm proposing here.

seancolsen:
One small comment/question for @mathemancer and @pavish...

My proposed method names use plural nouns like this:

```
schemas.add
schemas.delete
schemas.get
schemas.list
schemas.patch
```

I have a slight preference for using singular nouns like this:

```
schema.add
schema.delete
schema.get
schema.list
schema.patch
```

I'm fine with plural nouns if that's what you two prefer though. I used plural nouns in the PR because that's how Brent started off in https://github.com/mathesar-foundation/mathesar/pull/3524

pavish (approved):
@seancolsen Looks good to me!

I have a slight preference towards plural nouns in the function names. I do not have a strong opinion and I'm good with either.

docs/projects/2024/architecture-transition/rpc.md
```diff
@@ -128,76 +77,136 @@ The move to JSON-RPC opens up the possibility for the front end to consolidate m
...
+| `/api/db/v0/connections/{connectionId}/`                    | PATCH       | `connections.patch`                     |
+| `/api/db/v0/connections/`                                   | GET         | `connections.list`                      |
+| `/api/db/v0/data_files/{id}`                                | GET         | _(keep)_                                |
+| `/api/db/v0/data_files/{id}`                                | PATCH       | _(keep)_                                |
+| `/api/db/v0/data_files/`                                    | POST        | _(keep)_                                |
+| `/api/db/v0/links/`                                         | POST        | `data_modeling.add_link`                |
+| `/api/db/v0/queries/{queryId}/`                             | DELETE      | `explorations.delete`                   |
+| `/api/db/v0/queries/{queryId}/`                             | GET         | `explorations.get`                      |
+| `/api/db/v0/queries/{queryId}/`                             | PUT         | `explorations.replace`                  |
```
mathemancer, line 91: Is this actually a thing? I.e., is there a UI flow for wholesale replacing an exploration?

docs/projects/2024/architecture-transition/rpc.md
```diff
@@ -111,7 +43,20 @@ We are enforcing a standard of always using named parameters and never using pos
...
+| ✅ Use      | ❌ Don't use |
+| --          | --               |
+| **list**    | ~~get all~~, ~~select~~ |
+| **get**     | ~~get one~~, ~~fetch~~, ~~return~~ |
+| **add**     | ~~create~~, ~~insert~~, ~~make~~, ~~new~~ |
+| **patch**   | ~~partial update~~, ~~edit~~, ~~alter~~ |
+| **replace** | ~~full update~~, ~~put~~, ~~set~~ |
+| **delete**  | ~~remove~~, ~~drop~~ |
 
 ### Docstrings
```
mathemancer, line 62: We're no longer using Sphinx, but rather Google-style docstrings.

docs/projects/2024/architecture-transition/rpc.md
```diff
@@ -128,76 +77,136 @@ The move to JSON-RPC opens up the possibility for the front end to consolidate m
...
-| `/api/ui/v0/tables/{tableId}/shares/`                       | GET         |           |
-| `/api/ui/v0/tables/{tableId}/shares/`                       | POST        |           |
-| `/api/ui/v0/users/{userId}/`                                | DELETE      |           |
-| `/api/ui/v0/users/{userId}/`                                | GET         |           |
-| `/api/ui/v0/users/{userId}/`                                | PATCH       |           |
-| `/api/ui/v0/users/{userId}/password_reset/`                 | POST        |           |
-| `/api/ui/v0/users/`                                         | GET         |           |
-| `/api/ui/v0/users/`                                         | POST        |           |
-| `/api/ui/v0/users/password_change/`                         | POST        |           |
+| Endpoint                                                    | HTTP Method | Function                                |
```
mathemancer, line 80: Overall, I'm supportive of the naming scheme you have here. Thanks for going through the details on it!

mathemancer (changes requested):
Overall, the changes look great. One change I'd suggest adding is to update the documentation of our docstring style from Sphinx to the [Google style](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

docs/projects/2024/architecture-transition/rpc.md
seancolsen: Yes. When the user clicks the "Save" button on an exploration, we go [here](https://github.com/mathesar-foundation/mathesar/blob/develop/mathesar_ui/src/systems/data-explorer/exploration-inspector/ExplorationTab.svelte#L76) and then [here](https://github.com/mathesar-foundation/mathesar/blob/develop/mathesar_ui/src/stores/queries.ts#L224). That's a PUT request. This is code that @pavish wrote, I think with Dom's help on the backend. I imagine they both had a good reasons for using PUT instead of PATCH, but I don't understand enough about the inner-workings of the exploration data structures to understand why. In my audit of REST endpoints, this one did stand out to me as the only endpoint where we're using a PUT request.

@pavish if you want to shed any light on your reasoning here, that might be helpful.

docs/projects/2024/architecture-transition/rpc.md
seancolsen: Fixed in 632e0cefb78c35bd9cd7e8785a69bfc29394aee5

seancolsen:
@mathemancer this is ready for re-review.

mathemancer (approved):
All good from my end.

PR #114, opened 2024-05-07
Propose changes to record summaries for beta

seancolsen:
This PR describes my plans for the changes to record summaries for beta.

**[Rendered](https://github.com/mathesar-foundation/mathesar-wiki/blob/record_summary/docs/projects/2024/architecture-transition/record-summaries.md)**

### DIFF
--- a/docs/projects/2024/architecture-transition/record-summaries.md
+++ b/docs/projects/2024/architecture-transition/record-summaries.md
@@ -1,0 +1,79 @@
+# Record summaries
+
+As part of the 2024 architecture transition (for Mathesar's beta), we're making some changes to the way that record summaries work. This document describes those changes, along with the rationale.
+
+## Motivation
+
+### Short term (for beta)
+
+- We are seeking to improve performance for beta, and record summaries are currently a major performance drag. Brent sees some low-hanging fruit for improving performance by changing the way record summaries are generated.
+
+- The process for generating record summaries currently relies on significant work being performed in the service layer. As part of our larger architectural transition, we are seeking to move more logic from the service layer into the DB layer. Keeping record summaries in the service layer blocks us from making other architectural changes. The record summary logic ideally needs to be moved down into the DB layer to facilitate our other architectural goals.
+
+### Long term (post-beta)
+
+In their current form, record summaries have a number of limitations. Here are some examples of record summaries that we can't generate:
+
+- (A) A record summary for a person, using their nickname when available while falling back to their first name if needed. This requires calling a function like `COALESCE` within the record summary, which we can't do.
+
+- (B) A record summary which formats a date column into string that's more human-readable than ISO format. This requires calling a formatting function.
+
+- (C) Imagine an e-commerce database where products have categories organized in a hierarchically category tree modeled by a `category` table with a self-referential `parent` FK column. One category has title "Computers" (with `parent` being null), while a child category has the title "Laptops" (with `parent` referring to the "Computers" category).  We'd like to summarize the category table such that the laptops record displays as "Computers > Laptops". Each category should show the full path to itself from the root, joined by a string separator.
+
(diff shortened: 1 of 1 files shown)


seancolsen:
@kgodey @mathemancer @pavish I've assigned you to review.

kgodey (approved):
This all seems fine to me. I think the syntax changes are fine as long as the UX makes sense. I know the UX is still TBD, but when we do have it figured out, I want to make sure the UX does the following:
(1) provides the user a GUI way to select columns, including transitive columns.
(2) explains the syntax to the user in case they just want to enter text.

A nice to have would be helping the user understand how their syntax is wrong if there's some sort of syntactical error.

docs/projects/2024/architecture-transition/record-summaries.md
```diff
@@ -0,0 +1,79 @@
...
+[
+  [2],
+  ", by ",
+  [10, 2],
+  " ",
+  [10, 3]
+]
+```
+
+That means: first include a reference to the column in the base table having attnum 2; then include a string literal, then include a reference to a column that can be found by joining on the FK at column 10 and finding column with attnum 2 in the joined table... and so on.
```
mathemancer, line 72: I think we should reuse the `fkey_path` and/or `jp_path` structures we already have defined for holding this info. Out of curiosity , are you planning to be able to go multiple joins deep? In that case I _really_ think we should use the more involved structures. They're more flexible in the long-run since they can handle arbitrary joins (at least in the case of `jp_path`). It also lets us reuse some machinery we've already build for getting the needed info (see `1_msar_joinable_tables.sql` for example).

docs/projects/2024/architecture-transition/record-summaries.md
```diff
@@ -0,0 +1,79 @@
...
+
+## New record summaries
+
+_This section describes how we intend for record summaries to work in Mathesar's beta release._
+
+Without any user configuration, records will be summarized by their first _text-like_ column. This is a slight change. For example, if the PK is text, then we use that. If the PK is numeric and the first non-pk column is a numeric FK column, then we keep looking to see if we can find a text column.
+
+Users will still be able to customize the template, but the template DSL will have slightly new syntax. Column references within curly braces will need to be quoted in the same manner as in SQL. So `{first_name} {last_name}` will work if the columns are `first_name` and `last_name`. But if the columns are `First Name` and `Last Name`, then the template will need to be `{"First Name"} {"Last Name"}`. The same quoting and escaping rules apply as with SQL. Any references which fail to parse correctly are interpreted as string literals to be included in the template verbatim. Users should rarely be interacting with the template DSL manually because Mathesar has (and will continue to have) a button to insert a column into the template.
```
mathemancer, line 50: Where do you foresee doing the translation between the "human friendly" version here and "machine friendly" storage form below? If in the front end, then I think we should consider coming up with something a bit simpler w.r.t. quoting and escaping. I'd like to avoid translating human-friendly to numeric identifiers in the DB if possible, since we can reuse some pieces from the data explorer and avoid some round-trips to the DB that way.

mathemancer (commented):
I remember having some ~arguments~ collegial discussions at the beginning of the record summary work wherein my position was that the data needed to fill record summaries was already gettable though the `joinable_tables` and associated infrastructure in the data explorer. That's still true, and I still think we should reuse as much of that stuff as possible. I _think_, unless I'm missing some important details, that this should be well aligned with your goals of making these more general and query-like for extensibility in the long run. It also avoids having two big heavy machines dedicated to automatically joining tables together and using the results for some kind of display.

For this spec specifically, the only change (I think) needed is to use the current join path data structure for the join specifications, and to emphasize that the front end is responsible for putting names to IDs for showing the user (using info returned from the back end, of course). Then, if we're doing that, then SQL-style quoting and escaping isn't necessary.

pavish (approved):
This looks good to me. I do not have any concerns.

mathemancer:
@seancolsen is this still relevant? I was thinking of closing it, but we could merge if it's accurate to what you actually implemented.

seancolsen:
I think it's accurate enough for the wiki. I didn't read over it right now, but I think we've held true to the spirit of this doc well enough that it's worth keeping around.

PR #119, opened 2024-12-06
Fix many typos

TheGiraffe3:
Fixes many typos across the wiki found via automated means.
Unfortunately, this is only about half of what was found; I kind of ran out of steam after an hour of typo fixes.
The rest of them can be found here (though for some it is debatable whether they actually are typos): https://github.com/TheGiraffe3/mathesar-wiki/actions/runs/12202373484/job/34042912364


TheGiraffe3:
@seancolsen may be interested in reviewing this.

seancolsen:
@TheGiraffe3 No need to open PRs like this in the future. Most of the pages on the wiki are of very little value. It's a place for us to put specs and internal documents, but they become irrelevant rather quickly, so I don't think it's worth putting time into cleaning them up.

PR #120, opened 2024-12-09
Move Zack from "Emeritus" to the "Maintainers" section

zackkrida:
This PR moves me into the Maintainers section of the wiki. It also removes the now-empty Emeritus section.


kgodey:
@zackkrida I think your matrix username is wrong here.

zackkrida:
Fixed in a8d3758523616bad5d978fe4e4e4beed234bcfb4, great catch.

PR #127, opened 2025-04-25
Explorations display improvements specs

seancolsen:
This PR adds two new specs documents to the [Table Query Integration project](https://wiki.mathesar.org/projects/table-query-integration/):

1. [Exploration Display Improvements](https://github.com/mathesar-foundation/mathesar-wiki/blob/explorations_dml_specs/docs/projects/table-query-integration/specs/exploration-display-improvements.md) (to be completed first)
2. ~~[Exploration DML](https://github.com/mathesar-foundation/mathesar-wiki/blob/explorations_dml_specs/docs/projects/table-query-integration/specs/data-explorer-dml/index.md) (to be completed subsequently)~~

    ~~This is further broken out into multiple sub-specs, only the first of which is completed:~~

    1. ~~[Updating Base Table Cells](https://github.com/mathesar-foundation/mathesar-wiki/blob/explorations_dml_specs/docs/projects/table-query-integration/specs/data-explorer-dml/updating-base-table-cells.md)~~

    Edit: I took this out. It'll be in a subsequent PR.

I'm seeking enough agreement on the specs to merge this PR and get to work on the display improvements and base table cell updates. And I'll continue fleshing out the remaining specs after we can agree/merge the stuff here.


ghislaineguerin (approved):
Thanks @seancolsen 

I'm good with the first spec for Display Updates. Only thing I’d flag is we should rethink how we represent "read-only." Instead of making everything gray (could cause low contrast), maybe we should style the editable state instead? Something like a hover effect could work better? I don't have a strong opinion on this.

As for the DML update for base table cells, I don’t recall us discussing how we’d handle record retrieval when the id isn’t included in the query. I’ve got a few questions around that but I don't think they require changes to your spec.

As far as the UX goes, my understanding is that these are intermediate steps toward creating a cohesive experience between tables and explorations, but they do not represent the final flow or UX we want users to have. If that is the case, I am fine with this.

seancolsen:
Thanks @ghislaineguerin

- Read-only UX

    > style the editable state instead?

    I'm open to these kinds of changes. But I want to point out that we already have a UX defined for read-only cells vs editable cells. It's implemented for the table page, and it takes effect when permissions stipulate that cells can't be modified. My proposal here is just to take that _existing_ UX (and implementation) and to apply it to the data explorer. I want to be cautious of making changes to existing constructs unless we have a good reason to do so. Personally, I don't see anything wrong with the current UX for editable/read-only cells. I think the contrast is plenty high. If we want to change the read-only/editable UX, then I'd want those changes to be consistent across the table page and data explorer. If we can agree on that premise of consistency, then I think we can effectively decouple this small UX decision from the larger plans here — because those UX changes can happen independently (before or after) these exploration display improvements.

- DML update for base table cells

    > record retrieval when the id isn’t included in the query.

    Hmmm, I'm not sure what you mean by this. Since you have a few questions, perhaps it'd be best to hop on a call and chat through them?

- cohesive experience

    > these are intermediate steps

    Absolutely!
    
    Currently, there is a large set of "things" that you can do in the table page but can _not_ do in the data explorer. At a very high level, I'm trying to gradually reduce that set of "things" by making the data explorer more and more capable of the table page features. At some point the data explorer will be capable of doing _everything_ the table page can! And at that point we'll be able to _remove_ the table page entirely. That's the "cohesive" experience I'm pursuing — one user-facing construct instead of two. But we're a long way from that. So yes, this is an intermediate step to get there. And we'll have many more such increments along the way.

    > they do not represent the final flow or UX we want users to have

    I will say though... in some respects, yes, the UX I'm specifying here _does_ represent the final flow I'd imagine. For example the read-only vs editable UX. That's an example of UX which I don't see any need to change. So it might be a good idea to chat through some of these concepts on a call to make sure we're on the same page about the longer-term vision.

zackkrida:
@seancolsen I'm very happy with the work here. Great job. I particularly appreciate the goal of "building the minimum viable record identification process" and I think that the proposal for "Updating Base Table Cells" strikes a great balance. It restricts the users to editing a single row from the base table at a time (great!), is optimized for the simplest possible happy path (i.e, your edits save without fanfare, you don't have to add the ID column first, etc.), and the error paths provide clear documentation. 

For others reading: @seancolsen and I discussed the ["Exploration Display Improvements" spec](https://github.com/mathesar-foundation/mathesar-wiki/blob/explorations_dml_specs/docs/projects/table-query-integration/specs/exploration-display-improvements.md) and I had one idea I wanted to share and get feedback on, from @ghislaineguerin specifically:

What if we move the left side "Query" panel into the Inspector? Much like we show a first "table" tab in the inspector in the table view:

image

We would move everything from the "Query" panel into an "Exploration" first tab in the inspector. Here's a rudimentary mockup:

image

seancolsen:
@zackkrida I'd be open to moving the query-building UI into the right sidebar. I'd be curious to hear what @ghislaineguerin and @pavish think of this.

For others reading: I also want call out an idea that both @mathemancer and @zackkrida independently suggested... We already allow the user to define column display options in the table page. When building an exploration, we could potentially utilize _those_ display options as a starting point for exploration display options. I argued against this due to the additional complexity it would present. But since two people so far have suggested it, perhaps we should consider it. I agree it would be a useful feature. We could always add it later, so I see this question most a question of scope and priority.

ghislaineguerin:
@zackkrida Moving the query builder to the inspector panel makes a lot of sense. It would help declutter the UI and let users hide it while still interacting with the exploration.

We should also reduce reliance on the exploration definitions listed in this tab and look for ways to surface some of that directly in the grid UI instead of only in the inspector.

I'm not too concerned about transformations being slightly hidden (say if a lot of columns are listed). We could support column-level links to the steps. For example, if a filter applies to a column, we could show the related transformations there and link to the relevant transformation step by clicking? Overall, this feels like a solid improvement and worth exploring.

pavish:
@seancolsen 

I like the proposals in the spec. I only noticed one item that I'd like to flag that others haven't done already.

> Merge routes

- We need to take into account that an exploration (or a worksheet in the future) can be shared externally via a public url, where editing capabilities would not be present (We removed the public sharing functionality but we are going to be doing that for forms, which would make it easier to enable it again for tables & explorations. Atleast one user has asked for this specifically for explorations).
- We also need to take into account that we would require advanced permissions for explorations in the near-future.
- Merging the routes is okay with me, provided we handle a read-only mode for explorations in the "Explorations page", which is currently served by the "Viewing page".

>  We already allow the user to define column display options in the table page. When building an exploration, we could potentially utilize those display options as a starting point for exploration display options.  I argued against this due to the additional complexity it would present.

This was the plan when we implemented the data explorer. I don't think it should be too complicated to implement, but I'm fine with deferring it to later.

@zackkrida @ghislaineguerin 

> We would move everything from the "Query" panel into an "Exploration" first tab in the inspector.

The current layout in the product makes it clear to me that the left panel is the input, the middle is the result and right panel is modifications on top of the output. Moving it all to the inspector panel is slightly confusing from my perspective.

Particularly, the 'Select columns' section does not align well with my interpretation of the inspector, I'm fine with the 'Transform Results' moved to the inspector. This is because the columns shown in the 'Select columns' section aren't what is reflected in the results, the results are the processed columns. Our existing UX also doesn't provide a clear way of removing a column. I'd like us to put a bit more effort and spec out how the end-user experience would work, before moving this to the inspector.

seancolsen:
Thanks @pavish

- **Merge Routes**

    > Merging the routes is okay with me

    Thanks for being clear on this point.

    > shared externally via a public url

    Yeah, I understand the potential future need to share explorations publicly, and I agree that we should be keeping an eye on that.

    > advanced permissions for explorations in the near-future

    I'd like to understand what you mean by "advanced permissions". This seems to tie into the next point...

    > provided we handle a read-only mode for explorations

    **Can you elaborate** on this point? This seems to be your only blocking concern with my proposal. But I don't understand what you want.
    
    Here's how I see it: there are three kinds of data to consider when thinking about a "read-only mode":

    - User data — this should be covered by permissions
    - Changing the exploration name/description or deleting the exploration — this is currently possible with both of our two exploration routes. I'm inclined to follow that precedent by not worrying about any "read only mode" here.
    - Modifying the exploration _definition_ — currently this is only possible via one of the two routes. But aside from the additional step of navigating to the other route, we have no metadata permissioning that could prevent some collaborators from modifying the exploration definition.

    So I'm curious which of those three types of data you're talking about when you say you want to handle a read-only mode. And what exactly does "handle" mean?

- **Default display options**

    Yeah, I'm still leaning toward deferring this, just to keep scope tight.

- **Panels and layout**

    I don't have very many strong opinions on this topic, but it sounds like several other people have some ideas and light opinions. So I think the next step here is probably a design call with you, me, Ghislaine, and Zack. I'll work on scheduling this.

seancolsen:
Hmmm...

> I am hoping to implement this for forms since the unauthenticated route & endpoints & potential security concerns force us to. 

I'm not seeing the connection between forms and more advanced permissions on the exploration definition.

I can see how more advanced permissions on exploration definitions would be useful. I just don't see it as a high priority right now. Aside from publicly shareable explorations (which, to me, is a different feature), I'm not aware of any users that have requested permissions on exploration definitions. So I'm okay with all collaborators having read/write access to all explorations.

But let's pretend we both agreed that it would be a high priority to add more advanced permissions on exploration definitions very soon. Here's how I would be inclined to design it...

- I'd still want to have one route — the exploration page (or eventually the worksheet page)
- When loaded with a saved exploration, the exploration page would allow all users to modify the exploration definition _ephemerally_. Those modifications would persist in the browser. The user could could change the exploration name, description, query definition, and display options.
- _If_ the user had the proper permissions, _then_ they could save over the original exploration object — or delete the object entirely.
- But if they lacked those permissions to write or delete, then they'd still be able to work with the exploration in their own little ephemeral sandbox. And they might even have separate permissions to "save as" a different named exploration. Eventually I'd really like to have "private" worksheets ­- worksheets that get saved to user's account only.
- Here's a couple use-cases to keep in mind:
    - I load a saved exploration but don't have permissions to modify it. I want to make a column wider — I should be able to do that. But when I do, where does that column width _persist_? That's a display option. It would mutate the exploration definition. If I want to come back to it later and still see my wider column, I might want to save the exploration privately or with a different name if someone else has not given me permission to modify the original exploration I loaded.
    - Adding an additional sort condition or filter is similar 
- So with all this behavior, the only UI that would be affected by the permissions would be the save and delete UI. Thus, that'd be very easy to change while maintaining one route for the exploration.
- This design mirrors the behavior of opening a file with a desktop application (e.g. LibreOffice Calc) when I don't have write permissions on the file. I can still make changes — I just can't _save_ my changes to disk. Maybe I can "save as", but it depends on the filesystem permissions.

There is, of course, an entirely different way to approach designing more advanced permissions on exploration definitions.... we could say: _"if you don't have permission to modify it, then the UI should prevent you from changing it"_. This would be more like opening a Google Sheet. I can see how there would be use cases for this behavior too.

With a sufficiently sophisticated permissioning system, we could even support _both_ of the above patterns, allowing users to configure permissions and behavior on a very granular level. There could be different permissions on the query vs the display. We could really go wild with thinking about different approaches here!

But regardless of permissions, exploration definitions already behave _ephemerally_. That is: they behave more like LibreOffice Calc than Google Sheets. My changes to the definition persist ephemerally first. Then I need to take the extra step of saving those changes before they stick. I really like this behavior because it allows users to fiddle with the definition without affecting the workflow of other collaborators. That's why I'd be inclined to take the ephemeral approach first and foremost if we were to begin designing more advanced exploration permissions. And later on (if a user need arises) we could consider building a more "locked down" exploration where the definition cannot even be modified ephemerally. That sort of UI work would have a lot of overlap with the "publicly sharable" exploration feature, and I think we'd be able to re-use much of the same code for it.

There's potentially a lot to think about here! So I want to stay focused on the steps necessary to combine the data explorer with the table page. And to keep scope tight I want to stick to the metadata permissions model we already have: all metadata is writable by all collaborators.

pavish:
> I can see how more advanced permissions on exploration definitions would be useful. I just don't see it as a high priority right now.

> Aside from publicly shareable explorations (which, to me, is a different feature), I'm not aware of any users that have requested permissions on exploration definitions. So I'm okay with all collaborators having read/write access to all explorations.

Here's how both these are connected, and advanced permissions becomes essential, (it's a security concern):
- Say, we implement publicly shareable explorations, which is tied to a postgres role when users view it from a public url.
- Take this scenario: A Mathesar admin creates an exploration on the base_table "Movies" and makes it public, named "movie_exploration", and attach their own role to the public view. It only contains columns from the "Movies" table.
- We have a Mathesar user "Ben" who is a collaborator using a role which doesn't have access to SELECT certain tables in the schema. Eg., they don't have access to "Actors".
- If they're able to edit explorations, they could modify the definition to choose the tables they don't have SELECT privileges on and save the exploration. Eg., They modify "movie_exploration" and join the table "Actors".
- They won't be able to view the results in the exploration editor, but they can view all tables & columns definitions because it's available to everyone.
- Now, they visit the public url and they are able to view the records within "Actors".

While there are ways to block this particular scenario, I'm sure there are going to cases which we don't foresee. That's why it's important to decide who can edit an exploration.

> And to keep scope tight I want to stick to the metadata permissions model we already have: all metadata is writable by all collaborators.
> Is this approach okay with you?

Yes, this is okay with me. I wanted to bring this up so that we have permissions in mind while modifying the UX and code. We don't want to face hurdles due to architectural/UX decisions we make now, while implementing it in the future.

seancolsen:
Thanks for walking me through that example @pavish. It's a good one! It sounds like you are comfortable moving forward with consolidating the two routes. I will be thinking about this example (and this class of problems) as we continue to iterate on the data explorer.

seancolsen:
@mathemancer @zackkrida and I had a [meeting](https://tldv.io/app/meetings/682349874031d700139c1e13) today where we discussed this PR.

Brent still has some qualms with the work as I specified it.

In pursuit of moving forward, I've pushed the following commits to pare this PR down into only the work that I'm confident we all agree on:

- In b40ed2db89f8b5ae23f0fc5b43591faac81e4991 I removed the record summary stuff from the display improvements.
- In 4a748231f65933f3d7cc235f44210f53315c73dd I removed the DML specs altogether.

Now this PR is focused only on display improvements. It's heavy on the front end changes and requires almost no backend work.

I'm going to merge it now, and I'll open other PRs we we can focus discussion on the remaining topics.

@mathemancer please feel free to review retroactively to make sure I've only merged the stuff that you do indeed agree with!

PR #128, opened 2025-04-28
Remove "Docker integrated image" references from QA

zackkrida:
Removes QA steps relating to the "Docker integrated image". I _think_ this refers to a previous installation method for Mathesar, but I may be mistaken.


Anish9901:
We include postgres as part of the docker image that we publish, this allows potential users to try out mathesar with a single `docker run` command without the need to set up a postgres server to connect with mathesar. The included postgres server is not turned on if mathesar is provided with the appropriate environment variables for an external posgres server, as is the case with our `docker compose` setup.

Here is how we do the switch if you are curious:
https://github.com/mathesar-foundation/mathesar/blob/e29a63c7654d3fe66c986023fcaa59786db73495/run.sh#L6-L17

zackkrida:
@Anish9901 Okay, so "integrated image" just refers to using the image's postgres instance for the Django DB rather than an external instance. Got it! I'll open a new PR to improve the language. As written the documentation sounds like a separate image needs to be tested. Thanks for clarifying.

PR #129, opened 2025-04-29
Improve clarity of necessary QA database configurations

zackkrida:
This PR aims to clarify language around which DB configurations need to be tested during Mathesar's QA.

### DIFF
--- a/docs/release/qa.md
+++ b/docs/release/qa.md
@@ -46,28 +46,42 @@ For both:
     ```sh
     docker image pull mathesar/mathesar-caddy:<version_number>
     docker image tag mathesar/mathesar-caddy:<version_number> mathesar/mathesar-caddy:latest
-    
+
     docker image pull mathesar/mathesar:<version_number>
     docker image tag mathesar/mathesar:<version_number> mathesar/mathesar:latest
     ```
+
+### Understanding Database connections
+
+#### Database types
+
+- **Mathesar Database** - The Django database used by Mathesar
+- **Data database** - The user database(s) being explored in Mathesar's UI
+
+#### Connection types
+
+- **Local** - On the host machine
+- **External** - Outside of the host machine
+- **Integrated** - Inside of Mathesar's docker image
 
 ### Tasks
 
 - [ ] Test installing the release from scratch by following our documentation
 - [ ] Test all different installation methods
-    - [ ] Docker compose
-        - [ ] All defaults, local
-        - [ ] Exposed on domain, DB managed by Mathesar
-        - [ ] Exposed on domain, DB preexisting for users, managed DB for Django
-        - [ ] Exposed on domain, DB preexisting for all data
-    - [ ] Docker integrated image
-        - [ ] All defaults, local
-        - [ ] Exposed on domain, DB managed by Mathesar
-        - [ ] Exposed on domain, DB preexisting for users, managed DB for Django
-        - [ ] Exposed on domain, DB preexisting for all data
-    - [ ] Build from scratch
-        - [ ] All defaults, local
-        - [ ] Exposed on domain, DB preexisting for all data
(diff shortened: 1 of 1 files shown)


docs/release/qa.md
```diff
@@ -46,28 +46,42 @@ For both:
...
-        - [ ] Exposed on domain, DB preexisting for users, managed DB for Django
-        - [ ] Exposed on domain, DB preexisting for all data
-    - [ ] Build from scratch
-        - [ ] All defaults, local
-        - [ ] Exposed on domain, DB preexisting for all data
+    - [ ] [Docker compose](https://docs.mathesar.org/latest/administration/install-via-docker-compose/)
+        - [ ] Local using all default settings
+        - [ ] Exposed on domain with different database configurations:
+          - [ ] Integrated Mathesar, Integrated Data
+          - [ ] Integrated Mathesar, External Data
```
Anish9901, line 75: Hmm, I don't think these points should be listed under the docker compose section. Its quite tedious to test these without making some changes to docker compose file, I suggest listing it as a separate point like before. This method is listed on the first page of [docs](https://docs.mathesar.org/latest/#try-mathesar) so, we need to make sure that it works properly.

docs/release/qa.md
```diff
@@ -46,28 +46,42 @@ For both:
...
+          - [ ] External Mathesar, External Data
+          - [ ] External Mathesar, Integrated Data
```
Anish9901, line 77: I don't think we should test this, if someone already has an external django db configured the integrated postgres server would be inaccessible as it would be turned off.

docs/release/qa.md
zackkrida: OOOH, now I understand. This is the piece I was missing. I didn't realize these steps in QA were meant to test running Mathesar _without_ docker compose, using the `docker run -it --name mathesar -p 8000:8000 mathesar/mathesar:latest` command. 

Now I have all of the information I need, I'll revise the PR once more. Thank you!

zackkrida:
@Anish9901 I've revised the PR can you re-review? Thanks!

docs/release/qa.md
```diff
@@ -46,28 +46,43 @@ For both:
...
-        - [ ] Exposed on domain, DB managed by Mathesar
-        - [ ] Exposed on domain, DB preexisting for users, managed DB for Django
-        - [ ] Exposed on domain, DB preexisting for all data
-    - [ ] Build from scratch
-        - [ ] All defaults, local
-        - [ ] Exposed on domain, DB preexisting for all data
+    - [ ] [Docker compose](https://docs.mathesar.org/latest/administration/install-via-docker-compose/)
+        - [ ] Local using all default settings
+        - [ ] Exposed on domain with different database configurations:
+          - [ ] Integrated Mathesar, Integrated Data
```
Anish9901, line 74:
```suggestion
          - [ ] Local docker pg Mathesar, Local docker pg Data
          - [ ] Local docker pg Mathesar, External Data
```
I'm not sure what's the best way to describe what needs to be tested here is, but basically, for the docker compose method, we want to test mathesar against the `mathesar_db` container that we provide in [our docker-compose config](https://github.com/mathesar-foundation/mathesar/blob/e29a63c7654d3fe66c986023fcaa59786db73495/docker-compose.yml#L185-L206).

Anish9901 (approved):
I'll approve the changes, should be good to merge once my suggestion is addressed.

docs/release/qa.md
zackkrida: I know what you mean. There's so many different options here that I forgot that we don't want to use the postgres "baked in" to the mathesar container, we want to use the postgres defined in the docker compose. 

I'll further revise this, thank you for the patience as I learn every layer here one at a time 😅

kgodey: Unsolicited opinion: Would recommend just merging Anish's change and updating terminology later again if needed, this looks like a major improvement that is only blocked by this terminology thing.

PR #130, opened 2025-04-29
Forms - spec

pavish:
This PR adds a spec for the "Forms" feature.

- [Rendered spec](https://github.com/mathesar-foundation/mathesar-wiki/blob/forms_spec/docs/product/specs/forms.md)


pavish:
@zackkrida I'd like a review from the product level, and to ensure if the scope needs to include/exclude items.

@mathemancer I'd like a review from the backend & DB perspective and see if we're in agreement. I'd appreciate product level concerns as well.

@ghislaineguerin I'd like a general review of the spec and see if the UX I have in the spec makes sense, and schedule calls to come up with the final UX. 

I'd like this PR merged if the requirements & constraints are satisfactory, and everyone is in agreement. We could have calls for any UX and implementation related disagreements/suggestions separate from the PR.

ghislaineguerin (commented):
Thanks @pavish 

A few things we should consider for the design:

- Batch selection of linked records, for example, where you show the list of genres, we could select multiple all at once. If need to create, could we say add new and then open a floating form instead? I feel we might nest too many items and clutter the UI.
- What happens with a form if the schema changes?
- Minimize initial work by having a way to auto-create a form for a specific table with some defaults. 
- How does the search record summaries work for date and time types? Do you need to insert the exact date format as shown in the summary?
- I think the form builder should not need to show the visual rendering of the actual form, it makes editing more cumbersome. We should have a way to configure then preview. We can discuss this.

seancolsen:
I did a very quick skim of this, and I'll offer some quick thoughts:

1. First of all, I'm really impressed with this, @pavish. You've done a fantastic job writing a spec that is thorough enough to answer a lot of in-the-weeds questions while also being cogent enough for me to easily digest. Very nice work!

1. In "Searching through record summaries", you said:  

    > Open to consideration: Use a table view with lookup column as we do right now in the table page

    I _think_ this mirrors an thought I had after seeing the "Filling in a foriegn key field" mockups. Basically: why don't we use the record selector that we already have? I imagine we'd need to remove some columns so as to avoid exposing too much information to end-users. But what I'm after is trying to re-use as much of the existing code as possible. This is really only for the sake of reducing implementation time/complexity. This could be a useful topic to touch on when we get a chance to chat about this on a call.

1. I see a good amount of detail specified for the behavior to the end user. But I'm not seeing as much detail for the behavior to the Mathesar user. Perhaps I've not read carefully enough. But I'd like to better understand the experience of _building_ a form.

1. My understanding is that the record summaries for related records would be constructed using the record summary template defined in Mathesar's current metadata. Is that correct? I think it might eventually be important to have dedicated record summary templates _per-form_. I'd be okay starting without this level of customization. But I think it'd be good to keep that possibility in-mind for the future.

1. You seem to be suggesting a new structure for the record summary template, i.e. with the "alias" property. I'd like to chat more with you about this to poke at it and toss around some alternative ideas.

I'll probably read the whole thing again when I have some more time to sit with it. But I wanted to get a quick "first pass" in before our meeting today.

pavish:
@seancolsen 

> My understanding is that the record summaries for related records would be constructed using the record summary template defined in Mathesar's current metadata. Is that correct? I think it might eventually be important to have dedicated record summary templates per-form. 

I'm answering this one first, the other answers depend on the context.

The record summary templates are per-form. Each form element that has a fk would have a `record_summary_template`.

> why don't we use the record selector that we already have? 
> But what I'm after is trying to re-use as much of the existing code as possible.

1. The current record selector relies on fetching table data. Our endpoints are authenticated and making them public for forms would be considerable work taking into account ensuring that end-users are only be able to fetch the records, tabular structure, constraints etc., for the specific tables mentioned in the form structure. Using a separate endpoint to just display record summaries is a lot less work to implement.
1. I'd like to reduce dependencies on table metadata for record summaries. Let's say the end-user wants to select the `director_id`, they open the record summary table, and they see a column for `person_id` which then shows a record summary based on the template configured in the `Directors -> person_id` column. After selecting, the displayed record summary would be the template configured by the `Movies -> director_id` column. This may confuse end-users of the form who do not understand any of the db structure. This also adds dependencies on the metadate of all the tables involved, which might later change and affect the form. I'd like to avoid that.
1. Record summaries can be configured across multiple linked tables, I'd like the end user do not have to understand or worry about any of it. They see a list of values and search through it to select one. Though we show columns in the dropdown filtering area, these columns correspond to all the leaf nodes in the record summary templates, not just the immediate columns in the linked table. I'd like to avoid the end-users from having to navigate multiple tables in nested modals to filter for a specific record they're looking for.
1. We could discuss this more over our call.

> But I'm not seeing as much detail for the behavior to the Mathesar user.

Yes, I haven't added this on purpose. I wanted the spec to be focused on the product requirements, technical considerations, and feasibility, and get us all in agreement over the goals.

I see the form builder part as UX work, separate from the project spec.

> I'd like to better understand the experience of building a form.

Once we have a UX design, we could have a review session for it. It doesn't have to be team wide.

> You seem to be suggesting a new structure for the record summary template, i.e. with the "alias" property. I'd like to chat more with you about this to poke at it and toss around some alternative ideas.

Yes, lets talk on this over our call.

seancolsen:
@pavish 

> The record summary templates are per-form

Ahh — lovely!

> Using a separate endpoint to just display record summaries is a lot less work to implement.

Interesting. I think was reaching a slightly different prediction about the quickest path. I would recommend that we touch on this during our call, just to make sure you're not overlooking some of the complexities that I'm seeing within the implementation of the forms-specific record selector.

pavish:
@mathemancer @seancolsen I've [updated the spec](https://github.com/mathesar-foundation/mathesar-wiki/blob/forms_spec/docs/product/specs/forms.md#2025-05-05-updates-after-review-calls-with-team) based on our discussions.

seancolsen (approved):
Good to go from my end!

zackkrida (approved):
The latest changes look good, LGTM!

docs/product/specs/forms.md
```diff
@@ -0,0 +1,466 @@
...
+	- Password,
+	- Captchas, etc.,
+1. Email notifications after a form is submitted.
+1. Creating an 'exploration' from a 'form' to view de-normalized data filled via forms.
+1. Perform DDL operations while building form.
+	- Being able to create a new table directly in the form view.
+	- Being able to create new columns directly in the form view.
+1. Advanced form-level validations.
+1. Multi-language support for end-users filling the form.
+1. Customizable Logos.
```
zackkrida, line 167: I just want to make some small suggestions here:

```suggestion
1. Customizable logos, banners, and accent color.
1. Conditional logic to display form fields based on sibling field values
```

docs/product/specs/forms.md
```diff
@@ -0,0 +1,466 @@
...
+	- I want to configure labels, descriptions, placeholders, and validation rules for each field to guide end users.
+	- I want to reorder fields via drag-and-drop and insert dividers or instructional text to structure my form visually.
+	- I want the main record and any related/nested records to all be created in one shot, when the end user submits the form.
+
+- **Form Submitter (End User)**  
+	- I want a clear, labelled form that only shows fields relevant to my task.
+	- I want descriptive messages to guide me fill the form.
+	- I want clear indications when the form is successfully submitted.
+	- I want useful help messages when form submission fails due to an error.
+	- I want to know when I'm filling the form incorrectly while I'm filling it, not only while submitting.
```
mathemancer, line 92: This section isn't a list of user stories at all, it's a list of features prefixed with "I want". Attempting to wedge a list of feature requirements into a user story framework is more distracting than helpful, so I suggest dropping the pretext and reformatting/rewriting this section to be more readable, e.g., drop the "I want" prefixes, and consider sub-bulletpoints.

mathemancer (approved):
Regarding the spec itself, I think we're good to go. I did make an optional stylistic suggestion that would make the requirements section more readable.

docs/product/specs/forms.md
pavish: @mathemancer I'm not sure we align on your definition on what an user story is. From what I've worked on in the past, user stories are usually feature descriptions written with "I want". Perhaps you're conflating user stories to use-case scenarios, which is presented after this section.

Here are a few documentations on user stories:
- https://www.mountaingoatsoftware.com/agile/user-stories
- https://www.productplan.com/glossary/user-story/

docs/product/specs/forms.md
pavish: Thanks! I've committed the suggestion.

PR #132, opened 2025-05-13
Propose specs for DML on exploration base table

seancolsen:
**[Rendered](https://github.com/mathesar-foundation/mathesar-wiki/blob/exploration_base_table_dml/docs/projects/table-query-integration/specs/data-explorer-dml/updating-base-table-cells.md)**

This was originally in #127 but I split it out to a separate PR in order to merge the stuff we could agree on.


seancolsen:
During a [call](https://tldv.io/app/meetings/682349874031d700139c1e13) today Brent had some ideas for changing this DML logic around to rely on the entirety of the query definition when performing DML. I'm keen on going down this road so long as this is possible while allowing DML on queries formed from raw SQL.

The next step here is for Brent to present a more fleshed out vision for how the APIs would work.

Here's a slightly modified example of what we worked on during the call:

1. Run this query:

    ```sql
    SELECT
      cast_roles.character_name,
      people.name,
      people.birth_date
    FROM cast_roles
    JOIN people ON people.id = cast_roles.person
    WHERE cast_roles.movie = (
      SELECT id FROM movies WHERE title = 'The Sixth Sense'
    )
    ```

1. See this:

    |character_name|name|birth_date|
    |--------------|----|----------|
    |Visitor #2|Neill Hartley||
    |Malcolm Crowe|Bruce Willis|1955-03-19|
    |Cole Sear|Haley Joel Osment|1988-04-10|
    |Lynn Sear|Toni Collette|1972-11-01|
    |Anna Crowe|Olivia Williams|1968-07-26|
    |Tommy Tammisimo|Trevor Morgan|1986-11-26|
    |Vincent Grey|Donnie Wahlberg|1969-08-17|
    |Darren|Peter Anthony Tambakis||
    |Bobby|Jeffrey Zubernis|1987-07-24|
    |Stanley Cunningham|Bruce Norris|1960-05-16|
    |Sean|Glenn Fitzgerald|1971-12-21|
    |Mr. Collins|Greg Wood||

1. Set a birth date for Greg Wood.

As currently spec'ed in this PR, the UPDATE we'd run would be like this:

```sql
UPDATE people
SET birth_date = '1960-01-01'
WHERE
  name = 'Greg Wood' AND
  birth_date IS NULL;
```

But that's no good. Because it would update two rows (hence throwing an error to the user).

mathemancer:
I spent some time tinkering with this, and I'm a bit excited about the potential. Unfortunately, my excitement led me to write a bit of a novella. Sorry about that.

## Goal

The `UPDATE` statement should be related to, and filtered on, what the user actually sees, and what their query is. Any errors (e.g., "too many updates") should be because their query produces multiple rows where a filter defined by elementwise equality with a row that they see in the result set matches.

### Algorithm

1. Translate any joins to cross-join `FROM` plus filters
2. Create new `WHERE` clause using:
   - filters from first step, and
   - filters from second step, and
   - filters defined by values visible to the user when they edit on the row they're editing
3. Use `UPDATE ... FROM ... WHERE` construction with:
   - `UPDATE` table as target table
   - `FROM` as comma-separated list of all joined tables found in first step (excepting update target)
   - `WHERE` clause constructed in 3rd step.

## Update example

Using the example query from @seancolsen

```sql
SELECT
  cast_roles.character_name,
  people.name,
  people.birth_date
FROM cast_roles
JOIN people ON people.id = cast_roles.person
WHERE cast_roles.movie = (
  SELECT id FROM movies WHERE title = 'The Sixth Sense'
)
```

### First step

```sql
FROM
  cast_roles, people
WHERE
  people.id = cast_roles.person
```

Applying this in the original query results in:

```sql
SELECT
  cast_roles.character_name,
  people.name,
  people.birth_date
FROM cast_roles, people
WHERE
  cast_roles.movie = (SELECT id FROM movies WHERE title = 'The Sixth Sense')
  AND people.id = cast_roles.person;
```

Note that this produces identical query results (modulo ordering, which is indeterminate)

### Second step

Adding the additional filtering based on the visible row selected results in:

```sql
WHERE
  cast_roles.movie = (SELECT id FROM movies WHERE title = 'The Sixth Sense')
  AND people.id = cast_roles.person
  AND cast_roles.character_name = 'Mr. Collins'
  AND people.name = 'Greg Wood'
  AND people.birth_date IS NULL;
```

Applying _this_ to the original query results in:

```sql
SELECT
  cast_roles.character_name,
  people.name,
  people.birth_date
FROM cast_roles, people
WHERE
  cast_roles.movie = (SELECT id FROM movies WHERE title = 'The Sixth Sense')
  AND people.id = cast_roles.person
  AND cast_roles.character_name = 'Mr. Collins'
  AND people.name = 'Greg Wood'
  AND people.birth_date IS NULL;
```

The result set is then:

|character_name|name|birth_date|
|--------------|----|----------|
|Mr. Collins|Greg Wood||

If there happened to be multiple rows matching, that's because there are multiple rows in the query result which are identical.

### Third step

The resulting `UPDATE` statement is then:

```sql
UPDATE people
  set birth_date = '1960-01-01'
FROM cast_roles
WHERE
  cast_roles.movie = (SELECT id FROM movies WHERE title = 'The Sixth Sense')
  AND people.id = cast_roles.person
  AND cast_roles.character_name = 'Mr. Collins'
  AND people.name = 'Greg Wood'
  AND people.birth_date IS NULL;
```

This updates precisely the row desired.

## API

I'd like to make this the back end's problem entirely. What I'd want an API request to look like, in English, is

- submit the exact same query, plus
- an additional filter defining the rows of the query result to update, plus
- the updates to perform.

The filter should use our preexisting filter spec, identifying columns via array index (this is a little different from what we have, but it makes things unified and simple for the front end). The updates to perform should similarly use the column array indices.

So, the update request would look something like:

```python
records.update(
    database_id: int,
    column_idx: int,
    new_value: Any,
    query_type: Literal['exploration', 'sql', 'basic', ...],
    query_spec: Any,
    filter: FilterSpec,
    update: {<int>: <Any>},
)
```

### Notes

- I'd be okay with combining, e.g., `column_idx` and `new_value` into a JSON object.
- Front end doesn't need to know the provenance of the column being updated, necessarily. It would be nice to avoid allowing updates when it's inappropriate, e.g., many-to-one aggregated columns and the like.
- Updating multiple columns at once becomes a bit tricky (what if they're in different tables?), but it's tricky anyway. Plus, we don't currently have "update multiple columns simultaneously" in our UI AFAIK.
- Back end would need to figure out the table to update, based on the query. We already need to be able to determine column provenance in the back end, so this doesn't really seem to add any work there.


## Regarding SQL inspection

- This requires a tiny bit of SQL inspection for raw SQL queries, but it's highly restricted. You only have to inspect the `FROM` clause of the query, and only in a specific way: to translate join styles. As an initial iteration, we could require users to create their queries using the filtered-cross-join style, but I don't really think it's too hard to do the translation. Every join type is a precisely-defined subset of the cross-join.

## What about CTEs and so on?

Still working out some details, but it does seem to work.

- For any `SELECT` query involving the actual table being updated (e.g., a query where it's joined with some CTEs) the algorithm does not change.
- For a `SELECT` where the `FROM` clause doesn't involve the table being updated, you need to add any `CTE` being selected from which has columns whose provenance is the target table to the `FROM` and filter for equality between the `CTE` column and the column it derives from. This actually _over_-specifies the filter, and could be bad for speed in edge cases, but is easier than trying to be more sophisticated. Plus, the PG query planner is great at simplifying over-specified filters in many cases.
- For explorations, this is pretty easy to set up, since we already keep track internally of the provenance of any result column.

## Performance

I think the biggest weakness of this methodology is that it requires re-running the query, including any joins, to do the update, which could be expensive. Given the scale we assume we're operating at at this point, I'm okay with that. But, it's something to consider.

## Wrap-up

I really think this style of editing joined data will satisfy most (maybe all) well-defined update possibilities, and isn't overly complicated to implement.

@seancolsen What do you think?

seancolsen:
Thanks for spelling all this out, @mathemancer.

I was hoping to avoid deriving SQL from SQL, just because I've been worried that appropriately handling all the edge cases would get complex. But I'm still very warm to this path forward! If we could nail this, it would certainly provide a better user experience than what I had in mind (because the user wouldn't need to have any many columns in their queries in order to do DML).

I'd like to spend some time playing with this approach. Off the cuff, here are some things on my mind that I want to spend more time considering and poking at.

- I want to spend like 30 minutes looking into the ease of SQL mods using pglast, just by tinkering within my little [query-lens](https://github.com/seancolsen/query-lens) project. I'm confident it should be feasible to reconstruct SQL fragments from AST nodes, but I just want to do a small amount of validation to get a rough sense of how much friction might be present.

- For some queries, I'm not sure that a join's `ON` clause can be logically moved up to the `WHERE` clause. For example, this exclusion join:

    > Show me the items that were not rented out during 2023:

    ```sql
    SELECT barcode
    FROM items
    LEFT JOIN rentals ON
      rentals.item = items.id AND
      EXTRACT(YEAR FROM time_out) = 2023
    WHERE
      rentals.id IS NULL;
    ```

    I'd be curious to see how you'd handle this example.

    _(Note: obviously barcode is unique in this example, but I want to understand how the algorithm would work respect to the joins. If necessary, I can produce a more fleshed out example to demonstrate this same concept within a real-world business-oriented problem. But I'm hoping that you can work off of this low-fidelity example for the sake of discussion velocity.)_

- I want to think a bit about how left and full joins would work.

- My instinct is that we'd probably want to give up if we see `UNION`. I've not thought about this too deeply yet though.

- I can see how CTEs would be okay if they're used for filtering/sorting or bringing in columns unrelated to the DML operation. But I'm not immediately seeing how we'd update a table that _originated within a CTE_. For example:

    ```sql
    WITH m AS (SELECT title, release_date FROM movies)
    SELECT title FROM m WHERE release_date = '1955-03-07';
    ```

    I'd be curious to see how you'd handle this example.

- My intuition tells me that we might also need to remove `LIMIT`/`OFFSET` clauses from the query in order to account for any non-deterministic sort order. But I'd like to think about this more. Perhaps we try to determine whether the sort order will be deterministic. That might get tricky.


I have some critique of the API parameters too, but that's secondary to the underlying algorithm, so I'll focus on the DB-layer stuff first.

Basically I want to continue poking at this and chewing on it a bit. I expect to have more time for this work on Monday.

mathemancer:
Regarding `LEFT JOIN`: You are correct that you can't get the same join by simply moving `ON` logic down to `WHERE`.

However, while it's possible I'm missing something, but I'm 85% sure that for `UPDATE` statements, _only records that satisfy the inner join_ are relevant when editing linked records. Otherwise, you'd need to `INSERT`.

For records on the left side of a `LEFT` join, it gets a bit messier (or maybe more thinking is warranted). The intuition is the same: the right side of a left join only comes into play for filtering whenever the inner join condition is met. So, you need to construct a "the join condition isn't met, and all left-side columns match filter, and add it to the `WHERE` using an `OR`. I need to work through some details, but I hope that gives an idea of what I'm thinking.

seancolsen:
Issue now here: https://github.com/mathesar-foundation/mathesar/issues/4500

PR #134, opened 2025-05-27
Update release process to account for QA'ing direct installation methods

pavish:
- This PR updates the release process to create a proper versioned tag when cutting the release branch.
- This ensures that the commands in the documentation can be followed as-is.
- This also makes it essential to clean up releases created for QA before publishing.


zackkrida:
This looks good, @pavish! I'm going to merge as-is.

PR #140, opened 2025-06-26
Deployments requiremeents and text

kgodey:


zackkrida (commented):
@mathemancer I see where you're coming from in the thinking that this shouldn't be a project. However, I think it _could be_ if we add a few more specifics, such as criteria for when we could consider this work "complete".

@kgodey Brent's comments are extensive so I'll let them be the primary focus of your re-review. I am comfortable moving forward with this as a project after some refinement around more specific goals, or moving it into discrete experimental issues for each provider we want to consider adding.

PR #141, opened 2025-06-26
SSO requirements and proposal

kgodey:


docs/product/requirements/2025/sso-oidc.md
```diff
@@ -37,91 +37,14 @@ SSO may also help us grow Mathesar adoption, because:
...
 ### Is it feasible to solve?
 
-Yes! We use Django for our users, and there are several libraries we can just drop into our codebase.
+Yes! We use Django for our users.
 
 * [juanifioren/django-oidc-provider](https://github.com/juanifioren/django-oidc-provider)  
 * [jazzband/django-oauth-toolkit](https://github.com/jazzband/django-oauth-toolkit)  
 * [mozilla/mozilla-django-oidc](https://github.com/mozilla/mozilla-django-oidc)  
 
 Further feasibility research is unnecessary, it should be quick to implement. I briefly considered just implementing it instead of writing this.
```
zackkrida, line 46: What is the intention behind removing "and there are several libraries we can just drop into our codebase."? Is it to be less prescriptive about a specific solution? I find "Yes! We use Django for our users." followed by a list of libraries confusing, as there's no context for what the list of libraries means.

docs/product/requirements/2025/sso-oidc.md
```diff
@@ -198,40 +121,61 @@ Grace (customer success team)
...
 * As Grace, I want to access live customer data in Mathesar by signing in with my usual company account so that I can self-serve metrics without waiting on the data team.
 
-## Out of Scope
+## Success Criteria
 
-This is explicitly an MVP, so the following things are out of scope:
+How we'll know we've succeeded with our goals for adding SSO:
 
-Authentication Methods:
+* We should transition [internal.mathesar.org](http://internal.mathesar.org) to this login style and not hate it.
```
zackkrida, line 128:
```suggestion
* We transition [internal.mathesar.org](http://internal.mathesar.org) to use Google Workspace OIDC SSO and not hate it.  
```

For clarity

docs/product/requirements/2025/sso-oidc.md
```diff
@@ -198,40 +121,61 @@ Grace (customer success team)
...
+* See clear error if a user is unauthorized based on domain or mapping.
 
-* No UI for viewing login attempts, group mappings, or audit trails  
-* No real-time SSO diagnostics or health checks  
-* No alerting for failed logins or misconfigurations
+End user experience:
 
-User Experience:
+* Users see a “Sign in with Google” / “Sign in with Okta” option on the login page.  
+* Sign in and immediately access the right data without manual setup.
```
zackkrida, line 153: Aside from normal access post-sign-in, are there any other implications to this bullet point? I'm not sure what it's for.

docs/product/requirements/2025/sso-oidc.md
```diff
@@ -198,40 +121,61 @@ Grace (customer success team)
...
 
-* No UI for viewing login attempts, group mappings, or audit trails  
-* No real-time SSO diagnostics or health checks  
-* No alerting for failed logins or misconfigurations
+End user experience:
 
-User Experience:
+* Users see a “Sign in with Google” / “Sign in with Okta” option on the login page.  
+* Sign in and immediately access the right data without manual setup.  
+* Use their existing work account to access data self-serve.
```
zackkrida, line 154: This one feels redundant given the above points.

docs/product/proposals/2025/sso-oidc.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Community Engagement
+
+We should:
+
+- seek input from community members who have asked for SSO, to ensure that we're building the right set of features and addressing their use cases.
+- update mathesar.org and our `README.md` to include SSO in tandem with the next release.
+	- On mathesar.org, this should include the homepage, product page, and a few use cases.
+- consider posting on self-hosting communities to let people know that we have SSO.
+- consider adding ourselves to https://ssotax.org/friends-of-sso.html
```
zackkrida, line 84:
```suggestion
- seek input from community members who have asked for SSO, to ensure that we're building the right set of features and addressing their use cases.
- update mathesar.org and our `README.md` to include SSO in tandem with the next release.
	- On mathesar.org, this should include the homepage, product page, and a few use cases.
- post on self-hosting communities to let people know that we have free SSO.
- add ourselves to https://ssotax.org/friends-of-sso.html
```

docs/product/proposals/2025/sso-oidc.md
```diff
@@ -0,0 +1,84 @@
...
+
+User Experience:
+
+* No role-selection prompt on login  
+* No way for end-users to request access or change roles  
+* No customization of login UI
+
+## High-Level Implementation Plan
+
+**Work needed**: Backend
```
zackkrida, line 64: Documentation work will also be critical here, since so much of the "interface" for setting up SSO will be via configuration file. We should be explicit about that.

docs/product/proposals/2025/sso-oidc.md
```diff
@@ -0,0 +1,84 @@
...
+| **Role** | **Person / Item** | **Status** |
+|-|-|-|
+| **Requirements** | [SSO (OIDC) requirements](/product/requirements/2025/sso-oidc) | |
+| **Author** | Kriti Godey |  🟢 Done |
+| **Reviewer** | Brent Moran | 🔵 In review |
+| **Reviewer** | Zack Krida | 🔵 In review |
+
+## Solution
+
+Yes! We use Django for our users, and there are several libraries we can just drop into our codebase.
```
zackkrida, line 12: I think we need to explain the specific solution we're using beyond just listing these libraries. I thought we discussed, for example, only using Google Workspace as the first OIDC SSO provider. This document as-written isn't prescriptive about which provider to use.

zackkrida (changes requested):
The requirements doc looks good to merge. In my opinion the project proposal is close, but is missing specifics in a few key areas and provides insufficient detail for someone to implement a solution.

docs/product/requirements/2025/sso-oidc.md
kgodey: The aim of this bullet is to document that from the end user's perspective, they just sign into Mathesar and have access to the things they need without any additional steps like e.g. having to request access, or having to do some sort of onboarding. It's "obvious" in the sense that we know how Mathesar works and that there's no reason to do any of that, but it still seemed good to include so that the full experience of the user was represented.

docs/product/proposals/2025/sso-oidc.md
kgodey: No, we're not limiting to Google Workspace, it was just the example in the use case. I'm sorry if I somehow implied otherwise during the discussion.

There is no reason to limit to Google Workspace, OIDC is a protocol, supporting Google Workspace is the same effort as supporting any other OIDC provider. For example, the [Mozilla Django OIDC](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html#quick-start) library's setup involves setting up:

```
OIDC_RP_CLIENT_ID = os.environ['OIDC_RP_CLIENT_ID']
OIDC_RP_CLIENT_SECRET = os.environ['OIDC_RP_CLIENT_SECRET']
OIDC_OP_AUTHORIZATION_ENDPOINT = "<URL of the OIDC OP authorization endpoint>"
OIDC_OP_TOKEN_ENDPOINT = "<URL of the OIDC OP token endpoint>"
OIDC_OP_USER_ENDPOINT = "<URL of the OIDC OP userinfo endpoint>"
# optional
OIDC_RP_IDP_SIGN_KEY = "<OP signing key in PEM or DER format>"
OIDC_OP_JWKS_ENDPOINT = "<URL of the OIDC OP jwks endpoint>"
```

Someone setting this up with Google Workspace would set up an internal [OAuth app](https://developers.google.com/identity/openid-connect/openid-connect) and then configure the above stuff using the info they got there. If they used a different provider e.g. Okta, they'd get the OIDC info from there. I'll put this all in the file too.

docs/product/requirements/2025/sso-oidc.md
kgodey: Ah I see that the line is also in the project proposal doc so I must've had some harebrained thought about differentiating between the two.

kgodey:
@zackkrida All changes addressed, I think.

@mathemancer Tagging you too since the change seems significant enough that it's worth a quick re-skim.

docs/product/proposals/2025/sso-oidc.md
zackkrida: Oh, that's great then! Does that have implications for frontend work, though? To ask another way, what does the "Log in with ODIC" button say or look like, and how does the user know what it means if it's not something "hardcoded" to a particular provider?

kgodey: I assume we can just say "Log in using SSO" or something generic, or maybe let users configure what the login button text should say as a setting?

kgodey: I looked at the Mozilla OIDC library and their example just uses a button named "Login", the actual login is handled by the IdP website anyway. So even something simple like that should work.

PR #145, opened 2025-07-01
Internal forms requirements doc

zackkrida:
This PR adds a new [requirements](https://wiki.mathesar.org/product/requirements-template/) document for "internal forms" in Mathesar.

### DIFF
--- a/docs/product/requirements/2025/forms-internal.md
+++ b/docs/product/requirements/2025/forms-internal.md
@@ -1,0 +1,56 @@
+# Internal Forms
+
+| **Role**     | **Person**  | **Status**     |
+| ------------ | ----------- | -------------- |
+| **Author**   | Zack Krida  | 🟢 Done        |
+| **Reviewer** | Kriti Godey | 🔵 In review   |
+| **Reviewer** | Brent Moran | 🔵 In review   |
+
+## The Problem
+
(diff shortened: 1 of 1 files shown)


zackkrida:
I'm not completely satisfied with this document; hoping review can help refine it significantly. It's definitely _closer_ than previous iterations.

kgodey (changes requested):
As we discussed in our call earlier today, @zackkrida, I don’t think we should merge this requirements document. 

**Reasoning:** This does not appear to be solving a problem that any of our current users have indicated is a priority. 

“The Problem” section appears to be reverse-engineering a problem based on the feature we want, it’s basically defined as “users often need to submit information without having full visibility into the underlying dataset” with the benefits being “cleanliness, consistency, and control”, but this is all too high-level and does not offer enough constraints to define the problem.

This issue is also visible in the “Success Criteria” part of this project, which involves finding real-world use cases.

The library makerspace use case _works_, but does not seem like a sufficiently compelling use case to justify building this feature. A 3D printer workflow likely needs to tie into external tooling as well. Timesheets could be a better use case (since the feature depends more on limiting access to the data set).

The requirements also add in default dynamic values, which should probably just be their own feature, since they are not really related to forms.

We could spend the time to solve these issues, but I do not think it’s a priority to work on, given that it does not currently come from user need.

zackkrida:
@kgodey I agree we should close this project. I'll create standalone issues for some of the smaller features that were part of this project next week.

PR #148, opened 2025-07-07
File attachments proposal

zackkrida:
This PR introduces a proposal for the File Attachments project. 

**[Rendered](https://github.com/mathesar-foundation/mathesar-wiki/blob/file-attachments-proposal/docs/product/proposals/2025/file-attachments.md)**

Open questions I have:

- Is deferring remote backend work the right call?
- Are there other import methods that could be simpler that I'm not thinking of?
  - Example: let user move files manually into the local files directory and then manually structure and add the correct cell metadata

### DIFF
--- a/docs/product/proposals/2025/file-attachments.md
+++ b/docs/product/proposals/2025/file-attachments.md
@@ -114,12 +114,6 @@ Data in the Postgres user database will work as follows:
 
 When a user downloads a file, the name of the saved file will be taken from the file name as stored in Postgres.
 
-### File name editing
-
-Users will be able to _edit_ the file name for each file. These edits will only affect the data in Postgres. They will not touch the files stored in the file attachments directory.
-
-Null and empty-sting file names will be _allowed_.
-
 ### Loading behavior when adding a file
 
 When the user uploads a file, the loading experience will be very basic.
@@ -276,6 +270,10 @@ Mathesar administrators could conceivably wish to restrict file uploads to a cer
 Initially, file previews will be constrained to images _only_. All other file types will require the user to open them in a new tab, leveraging built-in browser file previewing, or to download the file to view it.
 
 Other projects in the ecosystem tend to show previews for many more file types, like PDFs. This functionally technically depends on generating thumbnails, which we also do not plan to support initially.
+
+### No file name editing
+
+Airtable allows users to rename files after they have been uploaded. We won't support this feature initially.
 
 ## High-Level Implementation Plan
 


seancolsen:
My suggested changes are in #153

zackkrida:
@seancolsen I've merged in your changes, they look great. This is ready for @kgodey and @mathemancer to review.

kgodey (changes requested):
@zackkrida I've reviewed this PR. I appreciate the technical detail and the implementation plan is clear. 

### Evaluation issues
However, this reads more like a technical spec than a project proposal. It seems like an implementation was decided on and then described, rather than deriving from the requirements (or maybe I just don't see the link to the requirements?)

Specifically, it is difficult for me to:
1. evaluate this from a product perspective
    a. there is no description of specific user research or problems to solve, although I know that this feature is being prioritized based on user feedback. 
    b. the document does not include reasoning for most / many of the decisions, so it is difficult to know how to evaluate them. As a random example, why multiple files per cell and not one file per cell?
    c. the document is not written from a user point of view (as one example, the documentation section lists a bunch of technical things to address, but does not explain what the user would see or care about)
    d. the reasoning that we _do_ have in the document is focused on "because other products do X / don't do X", rather than deriving it from "our users need to do X". This is an extension of (a) and (c), but seemed worth calling out separately.
    e. the tradeoffs section reads more like technical limitations, but this was meant to be _product_ level tradeoffs. See the [SSO tradeoffs section](https://wiki.mathesar.org/product/proposals/2025/sso-oidc/#tradeoffs), we're listing things that we chose not to prioritize based on the requirements, they're a lot higher-level. I am not sure if I've articulated the distinction well, please let me know if I haven't.
2. determine whether this is the most minimal solution to our requirements
3. see what other solutions or approaches were considered

Additionally, I think the level of technical detail is a little overwhelming and makes the document seem more structured or contextualized than it is. (I'm not sure if I articulated that well either, let me know if confusing.)

### Next steps
I could sign off and say "looks good" because it seems coherent and internally consistent and we would probably have a file attachment feature that works. But I'm concerned that:
- it wouldn't be as minimal as it could be
- it may be solving problems we don't need to solve yet
- it may take longer to implement than we need to take

I think it is worth taking some time to discuss, since this is one of our first new features using this process, and the first one we didn't have a rough idea of what we wanted to do before we started.

I don't necessarily want to block this until next week, though. Can we discuss sooner?

### Potential solutions
I think these issues are expected; we're still ironing out our processes and this is still one of the "test case" features.

I think the potential solutions here may be a combination of:
- Needing to update our requirements process to focus more on existing user feedback and inputs
- Needing to distinguish the aims of the project proposal better
- Perhaps creating a separate technical spec stage for some projects?
- Updating this PR.

### Aside
I pushed some minor formatting changes to this PR.

kgodey:
Thanks @zackkrida! 

> A discussion sounds good, could we include @seancolsen? It's likely he'll author some of these proposals in the future and I think his input could be useful.

As discussed on Matrix, I agree that we should have separate next steps for:
- the file attachment scope (resolve ASAP, @seancolsen should absolutely be involved)
- product process implications + updating this document (can lag behind by a week-ish, I don't think this is a good use of Sean's time)

It's not as clean of a process, but we're ultimately making these product process changes to make things easier, not to slow down work, and moving forward with work should take priority. I think it makes sense that we're discovering these issues with the product process through creating new proposals, but if we continue to find that the process is taking a lot of time over the next couple of months, it's probably a sign of too much process.

> Next steps as I see them are:
> 
>     1. Chat a bit about your concerns in more detail.
>     2. Update the requirements doc for this project to include specific user feedback we've received.
>     3. Revise this proposal by:
>        a. Extracting some of the information into a technical spec.
>        b. Add in user-facing rationale for the decisions made.

I think next steps would now be:

1. Chat about file attachment scope; approve work.
2. Discuss the product process changes / implications / goals at next week's product planning meeting.
    a. Decide what changes need to be made to the doc / whether to split into multiple docs.
    b. Make those changes

I don't want us to spend a lot of time on revising the doc if it's not unblocking work; I want to prioritize applying what we learn here and making the process better for the next project.

> I think this might point to the requirements doc being insufficient. I consciously referenced the [SSO proposal](https://wiki.mathesar.org/product/proposals/2025/sso-oidc/) (because it's our only approved proposal) and I didn't see this type of information present there. My inclination is that this information belongs in the requirements doc, and then this proposal doc is evaluated on whether it solves the problems in the requirements doc or not.

I agree with you, I think the requirements document should include what user problem we're solving and what the minimum requirements for solving it are. We probably need better guidelines or heuristics about how to document those and also how to derive use cases from those.

> Agree with this, but I'm curious on what kind of reasoning _you're_ looking for. If a solution meets the user needs in the requirements doc, to my mind we'd evaluate based on heuristics like:
>     * Our intuition as OSS maintainers and users of software
>     * What's idiomatic to the ecosystem and existing offerings
>     * Minimum-viability based on technical feasibility, effort, and timeline concerns

This may just be a secondary effect of the requirements issue in this particular case.

In general, reasoning that indicates decision-making heuristics is most helpful to me i.e. what other approaches were considered and why did this one win? What things are consequences of a given decision, and what things were separate choices, and what else was considered for _those_ choices? What's easy to change about the proposal, and what's difficult without rethinking a whole bunch of other stuff? 

This helps me understand and identify assumptions in the problem space and constraints more easily, this is basically what I try to build in my head when I review something.

This is also the answer to this question you asked:

> > 2. determine whether this is the most minimal solution to our requirements
> 
> What aspects made this difficult? The aggregation of the things you've listed, or some other factors? I'm asking because I think questions like "Do we need this feature this way" are really important and should be a crucial part of reviewing a proposal like this. I'd want to see specific inline comments for each individual piece of functionality that could be simplified, maybe even with suggestions proposed. Maybe after this is revised subsequent reviews could put more focus on that.

Without understanding the reasoning / having to develop it all myself, it's hard to know what's change-able and what's not. For example, to pick a random detail, I don't know why we need a MIME type sniffer or what problem it's solving or what the alternatives are, so it's hard to say "Do we need this for the MVP?" Multiply that times however many headers this spec has.

That doesn't mean we need reasoning for each individual thing, just a description of what follows from what, or what assumptions things are based on. Or maybe we just need a separate technical spec stage.

>     * Users who have asked for this feature have asked for it to behave like other products.
>     * Other products, particularly Airtable alternatives, have taken the time to determine what's minimally-viable from Airtable's implementations in making their own implementations. We can benefit from that shared effort.
>     * The user signals we _do_ have are few (but strong, we've had 1 user express this as a "dealbreaker" feature) so there isn't as much to draw from.

This seems like a sign of needing more user research. Ideally, we would understand the user's precise needs and use case as part of developing requirements. I don't think this needs to be a lengthy or arduous process; we can just reach out to one or two users who have requested this and ask some targeted questions (email or call). 

To your specific points:
- It makes sense that users would ask for behavior like other products, generally people focus on what's familiar to them, we need to be disambiguating that.
- I don't think what's minimally viable from other products necessarily transfers to us. We also don't know for sure that they were prioritizing minimum viability. Our use cases are different. I think this is fine to use as _a_ signal but not as a primary decision-making tool.
- Makes sense, I'm not disputing the priority of this feature.

> I think this is the weakest element of this proposal and definitely needs revision. The user facing justification for many of these features were discussed and considered but not captured here.

I think we also need to get into the habit of _framing_ the proposal from the user's PoV, FWIW.

> Here's a quick example from SSO I recently encountered that underscores the point. Both the requirements doc and proposal for SSO refer to using a _config file_ for `SSO` rather than user interface elements within Mathesar. We never explain _why_, though! Some potential reasons I can think of:
>     * A config file is considered simpler to implement than UI changes
>     * A config file might be more portable than UI changes
>     * Technical users might prefer configuration over working in a UI
>
> In practice, there were some qualities of using a config file I didn't necessarily expect:
>     * We had to figure out how to encode the config in an ENV variable for installs where the filesystem is ephemeral
>     * We had to _document_ how to write the YAML config, which is much more complicated than explaining how to fill out a form in Mathesar, for example.
>
> It's possible that if we discussed the "why" for that approach in the document, we would've gone with a different solution more optimized for whatever the goals were.

Good catch! My reasoning here was that it was less work to implement, but also because it was a Django setting, it would be hard to implement within a UI. But I'm not sure if those assumptions make sense, I agree it makes sense to discuss those kinds of things in the document. 

I know it's hard for authors to catch their own assumptions, but perhaps we could update the template or framework to make this easier.

zackkrida:
@kgodey agreed on all points, and your revised next steps look excellent. I'm going to draft this PR in the meantime.

seancolsen:
@zackkrida and @kgodey, per yesterday's [call](https://tldv.io/app/meetings/6887ba579aee6e00138f2386) I've pushed some changes to update this proposal:

- 53d294ff58b0da673edfa271c04a68182278517a changes it to specify one file per cell in the DB layer and in the UI.
- ac574fe290021c8b3968f7d61790af137e3021ea cuts file renaming from scope

I think we had agreed that I'd make some more changes too, like listing open questions and seeking other places to cut scope. But the more I thought about it, the more aspects of this proposal I saw which are dependent on the file storage approach. If we change that approach (e.g. to leverage functionality from Django), then there's potentially a _lot_ more to change about this proposal. So for the sake of efficiency, I'd like to sort out that file storage approach via discussion before I get deeper into adjustments to the proposal.

mathemancer:
I'm going to add a comment here, rather than a full review. After doing some tinkering at @kgodey's behest, and discussing with her, I've got some ideas for this feature (some are more properly @kgodey's ideas):

The central idea is: We shouldn't have a database type at all for this. What we need is a "treat this column like a value could point to (or be) a file" metadata option on the column. In principle, there isn't really anything to restrict on the db layer at this point, so using a DB type to enforce validity isn't useful. We could (and, I propose, should) set up machinery at the service layer that, whenever dealing with values in a column marked as "could be files" that:

- generates a temporary "file download" link from a URI (e.g., one starting with `s3://...`) if applicable for each value
- generates a temporary "file download" link for files stored as bytes in the DB
- generates a thumbnail from a URI if applicable and possible, and a download link for that thumbnail
- bundles that stuff and other derived or generated metadata in the response
- etc.

The overall flow when building the table page would be something like:

- front end gets table column info, notices one is configured to work with files
- when listing records, asks for whatever metadata is desired for that column (e.g., download links for thumbnails if applicable, download links for files, mime types, etc.)
- Metadata is in the response (no files yet)
- The front end grabs thumbnails from download links
- If looking at the file itself, the front end uses the download link provided
    - links would be temporary (maybe 5 minutes of life)
    - front end could request ttl up to some limit in the records request

The flow when uploading a file:

- Front end POSTs the file (hand wave here) to Mathesar
- The file is saved to the configured backend, pointer or other persisted info is saved to the database cell.
  - Challenge: we need to only allow uploading files to columns where the DB type is aligned with the configured file persistence backend. E.g., if the backend is in the DB, then the DB type can't be a URI. If the backend is S3, the DB type can be anything "text-like" and big enough, but not `bytea`.
- The file name, suffix, etc. aren't changed.

If the user just types into the cell in the "file" functionality column:

- Just insert the value like usual (as long as it fits the underlying DB type).
- When listing, all the same machinery applies

Challenges:

- less caching and persisting of metadata and thumbnails could mean performance issues to solve.
- Deleting files when deleting a row is not guaranteed, and may not be desirable in many situations

What do you all think?

I'll try to prototype something like what I'm thinking would work, to see how we feel about it without spending too much time (think experimental mode).

seancolsen:
Interesting, @mathemancer

I've read over this a few times, but I'm still not seeing it as clearly as I'd like. I have a lot of questions, but I'm not sure that it would be a good use of time to try sorting all of them out async. (I predict answers would generate more questions.)

Seeing an example would help me. I'd want to see exactly what would be stored in the user DB, what would be stored in the internal DB, and would be stored in S3 (or other file storage). And with all that, I'd also want to see any _caching_ storage we'd implement at any of those layers.

I'm not sure what the process is supposed to be to move this towards implementation. And it's no longer clear to me what my role is in this project. So I'm cautious about diving too deep into discussion until someone explicitly asks me to do something.

kgodey:
@seancolsen Sorry for the confusion; the process is still a work in progress, this is our first new feature that we're speccing from scratch in the new product planning workflow, and we're iterating on both this proposal and the process for projects like this at the same time, so there's been some ambiguity as we've been figuring that out.

Here's the state of things as I understand it.

1. You, Zack, and I had a couple of calls on Friday and Monday after I reviewed the proposal and had concerns.
2. The outcomes from those calls were that we had several backend items to investigate (storage mechanism, DB structure, etc.) and also some frontend flow to figure out.
3. The plan (as I understood it) was for
    1. You to document the changes we were sure of (done!)
    2. Brent to investigate backend items
    3. You and @zackkrida to work through the frontend flow
    4. Me to approve backend and frontend work.
5. We had also decided to prioritize getting the implementation moving rather than getting the proposal fully updated, and apply the things we learned from this process to the next proposal rather than trying to hold this one up for the new process.
6. At the product planning meeting on Tuesday, we updated Brent on this plan so that he could complete the backend investigation by Wednesday, so that I could review from a product perspective before passing off to you / Zack / Brent for implementation.
7. Brent and I met on Wednesday to discuss what he had so far; I had several product-level concerns with the approach so the next step there was for Brent to do another pass. We didn't think it made sense to involve you at that point because I still had issues with the approach.
8. Brent did a second pass on the plan yesterday and we discussed it last night, and I signed off on it.
9. Separately, Zack updated me on the frontend flow that you discussed and I signed off on that too, after asking about the data explorer and getting a new plan on that.
10. Now it's ready for implementation speccing or prototyping and whatever process/discussion that entails. Hopefully it shouldn't take too long given that there's a clear idea for both the backend and frontend flows now.
11. I don't expect to be involved any longer in this process unless either the frontend plan or the backend plan needs to change such that it needs product level review again.

It's up to you, @mathemancer and @zackkrida how you want to structure next steps; I asked @mathemancer to post here to get the conversation started. I would suggest that if you need to see examples, it may be a better use of time for @mathemancer to prototype the backend (as opposed to describe it before prototyping, either here or on a call), but I'll leave that up to you.

I hope this helps.

zackkrida:
@mathemancer I really like where this has landed! It'll be a very flexible model for supporting lots of different backends in a unified way. I have some questions about the specific scope we want to constrain ourselves to for _this_ release:

- Do we plan to document or "support" specific storage backends? Or will we frame this as "any S3-compatiable API" will work? Or do we plan to support different backends up-front, like in-DB or local? I know they're all _possible_ with the structure defined, but given we have three weeks to work on this it seems we should be as constrained as possible. 
- The same question but for "working with existing file columns". If someone has a column of S3 paths, they configure Mathesar with an S3 backend, and then check the "This column contains files" button, what will happen? It sounds like everything could "just work", but I'm not certain.

What are next steps? What will @seancolsen need to start frontend work? Are you going to start prototyping this backend, @mathemancer? Please let me know how I can help coordinate.

PR #153, opened 2025-07-21
Sean's suggested changes to the File Attachments proposal

seancolsen:
As [discussed](https://tldv.io/app/meetings/68793255de25ad0013314203) on a call, I made some edits to the file attachment proposal.

### DIFF
--- a/docs/product/proposals/file-attachments.md
+++ b/docs/product/proposals/file-attachments.md
@@ -21,7 +21,7 @@ Additionally, we'll use the following terms internally:
 
 ### Multiple files per cell
 
-Each cell in a Files column can contain zero, or multiple files.
+Each cell in a Files column may contain zero, one, or multiple files. Cells can not be restricted to contain a certain number of files.
 
 ### File immutability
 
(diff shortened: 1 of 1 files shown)


seancolsen:
@zackkrida please review. Thanks!

zackkrida:
@seancolsen this is really great work, it captures what we discussed and adds some great implementation specifics as well. I won't be able to fully review and merge this until early Wednesday with some release prep I'm doing. That said, I don't see any functionality here that I object to or that we shouldn't do. 

The main thing I'll be reviewing for is the scope of these "product proposal" documents and how much detail we want here or relegated to some kind of "technical spec" document to compliment this one.

docs/product/proposals/file-attachments.md
```diff
@@ -4,82 +4,444 @@
...
+- **"Files"** — the name of the UI type for the column
+- **"File"** — a single file attachment within a cell
+
+Additionally, we'll use the following terms internally:
+
+- **"File cell"** — one cell within a Files column (Note this is not "Files Cell", just for the sake of simplicity)
+
+### Multiple files per cell
+
+Each cell in a Files column can contain zero, or multiple files.
```
zackkrida, line 24:
```suggestion
Each cell in a Files column may contain zero, one, or multiple files. Cells can not be restricted to contain a certain number of files.
```

